PR #188, opened 2024-09-24
[Draft] Use nemo run CLI

skothenhill-nv:
## Overview

This PR contains a prototype for a NeMo-Run based CLI for BioNeMo. Importantly, this involves defining the configuration layer between NeMo-Run and BioNeMo. These are constructed in `factories.py` with the combination of:

- Config classes (these are the literal things we are making)
- Factories (NeMo-Run registered functions that produce Configs)
- Functions (regular python functions that combine configs into a real workflow).

Additionally we have added an entrypoint under `sub-packages/bionemo-geneformer/src/bionemo/geneformer/run/factories.py` and a SLURM example under `scripts/singlecell/geneformer/slurm-run.py`.

When adding new factories, configs, or entrypoints, each must be tested carefully. Errors from NeMo run are often difficult to debug. Factories are registered to types, so ensuring correct typing is critical.

## Pitfalls
- Not all fields are serializable, to get around this, we create configs that expose only the parameters required. ExposedModelConfig is the best example of this.
- Generics are not handled correctly by NeMo-Run, factories which return concrete types on a generic must use the generic return signature, or NeMo-Run will fail.
- NeMo-Run CLI captures stdout and stderr, be mindful of this when attempting to debug.
- NeMo-Run CLI handles launching processes for torch distributed, meaning we should NOT use the SLURM context for PTL trainers (plus al few other things need to be setup correctly).
- CLI help does not provide help strings for registered factories or expected fields.

## Examples
Confirmation of configs before execution
image

Help Message, displays the various factories for each
image

Command line usage
image


spoulos-nvidia:
closing as out of date

PR #209, opened 2024-09-27
Initial commit for DNA datasets

cristianregep:


sub-packages/bionemo-dnadl/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[project]
+name = "bionemo-dnadl"
+license.file = "LICENSE"
+readme = "README.md"
+description = "DNADL, a Dataset class for DNA data."
+authors = [{ name = "BioNeMo Team", email = "bionemofeedback@nvidia.com" }]
+dynamic = ["version"]
+dependencies = [
+    "pyfaidx>=0.8.1.2",
```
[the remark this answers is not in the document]
gwarmstrong: We have had major threading issues in the past with pyfaidx, I would highly recommend finding a thread-safe alternative. @skothenhill-nv may have some suggestions here.

sub-packages/bionemo-dnadl/tests/bionemo/dnadl/io/test_dnadl_dataset.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Write the dummy FASTA content to the temporary file
+    with open(temp_fasta_path, "w") as fasta_file:
+        fasta_file.write(fasta_content)
+    return temp_fasta_path
+
+
+@pytest.fixture(scope="module")
+def temp_bed_file(tmp_path_factory):
+    """Fixture that creates a temporary FASTA file with dummy content and yields the path to the file."""
```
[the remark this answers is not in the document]
cristianregep: done

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class Genome:
+    """Class that creates a genome object from a fasta file. It can be used to extract sequences from the genome."""
+
+    def __init__(self, fasta_file: str | os.PathLike):
+        """Instantiate the class."""
+        fasta_file = Path(fasta_file)
+        assert fasta_file.exists(), "path to fasta file must exist"
+        self.seqs = Fasta(str(fasta_file))
```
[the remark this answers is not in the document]
cristianregep: I've done a workaround, the fasta file gets created and destroyed upon access. I understand there's also something else to replace pyfaidx being prepped, but this should work in the meantime.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,206 @@
...
+def _thread_safe_genome_access(
+    fasta_file_path: Path, chromosome: str, start_index: int | None = None, end_index: int | None = None
+) -> Sequence:
+    """Bypass the multiprocessing issues with pyfaidx by destroying the fasta object after use."""
+    fasta = Fasta(fasta_file_path)
+    if start_index is not None and end_index is not None:
+        sequence = fasta[chromosome][start_index:end_index]
+    else:
+        sequence = fasta[chromosome]
+    del fasta
```
skothenhill-nv, line 55: While this may resolve the majority of issues from simultaneous reads, it also introduces a race condition on `.fai` creation and adds a big overhead for reads. I've left a more detailed comment in the main discussion with suggestions.

skothenhill-nv:
We should find an alternative for PyFaidx. I dont think there is a straightforward solution to integrating it in a way that works well with pytorch DataLoaders.

Reading in the pyfaidx inside the workers I _think_ resolves the issue with pyfaidx (although i'm not 100% sure). But it has two problems:

1) race condition is introduced if the `.fai` file does not already exist, each worker will try to create this at once. 
2) Creation of the `pyfaidx.Fasta` object is quite slow (order of seconds, tens of seconds for gigabase scale fastas). 

Fortunately, in the last year the pyfaidx authors actually commented on the issue~ it seems fundamental to the c implementation of faidx- as htslib is similarly not process safe.

https://github.com/mdshw5/pyfaidx/issues/211

We should look for an alternative. I have to suggestions:

Check what we did in `bionemo1`, we have a branch on github with an example. In this case, we used NeMo's CSVMemMapDataset with '>' as the key delimeter. To make this work, we had to drop the newlines from sequences in the underlying fasta file. Not a huge fan of this solution in general, but it has a very light memory footprint.

https://github.com/NVIDIA/bionemo-framework/blob/bionemo1/bionemo/data/fasta_dataset.py

Alternatively, we could implement a wrapper on a different fasta index implementation. I've been looking at the implementation here for sometime, unclear if its process safe, but its relatively straightforward to wrap rust objects as PyObjects (PyO3 does this, takes very little rust knowledge to implement).

https://github.com/zaeleus/noodles

Let me know if we can be of help with this. A _safe_ fasta reader has been on our radar for a while.

skothenhill-nv:
> Alternatively, we could implement a wrapper on a different fasta index implementation. I've been looking at the implementation here for sometime, unclear if its process safe, but its relatively straightforward to wrap rust objects as PyObjects (PyO3 does this, takes very little rust knowledge to implement).
> 
> https://github.com/zaeleus/noodles
> 

I went ahead and got started on this, the noodles-faidx reader is thread safe. Need to do a little more profiling but I should be able to get you an update soon @cristianregep

skothenhill-nv:
@cristianregep feel free to work on it with us! effort is happening [here ](https://github.com/NVIDIA/bionemo-framework/pull/437/files) (and in the target branch)

skothenhill-nv:
@cristianregep hey Cristian, we just merged a python interface for safe fasta index backed by mmaps. It should also help with memory usage. 

Checkout `sub-packages/bionemo-noodles`, the tests have some examples of using it with a dataset class. The interface is quite similar to outside, but if there is missing functionality feel free to add it to this PR.

If you have any issues just let me know!

cristianregep:
> @cristianregep hey Cristian, we just merged a python interface for safe fasta index backed by mmaps. It should also help with memory usage.
> 
> Checkout `sub-packages/bionemo-noodles`, the tests have some examples of using it with a dataset class. The interface is quite similar to outside, but if there is missing functionality feel free to add it to this PR.
> 
> If you have any issues just let me know!

Thanks, sorry to have been MIA, it's been intense delivery on the Relation side. This looks great

cristianregep:
@skothenhill-nv @ntadimeti I've updated my branch to use NvFaidx. Thing this is good to go

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
cristianregep: I've now updated to use NvFaidx

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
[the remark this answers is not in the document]
cristianregep: I've now updated to use NvFaidx

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+        self.tokenizer = tokenizer
+
+    def __getitem__(self, ind: int) -> torch.tensor:
+        """Get an item from the dataset."""
+        sequence = self.get_dna_sequence(ind)
+        return self.tokenizer(sequence)
+
+    def get_dna_sequence(self, ind: int) -> str:
+        """Get the DNA sequence."""
+        ...
```
skothenhill-nv, line 97: Could you expand the docstrings here? no specific feedback other than adding arguments, but here's my perspective as a reader: 

- the constructor implies this is doing a transformation with the tokenizer, like a `MappedDataset`
- it looks like `ind` is meant to provide a deterministic mapping from dataset 'index' to some DNA sequence.
- I can think of how I might want to use this, but maybe providing a few concise examples in text would help future implementors (or just point them at your implementations below).

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+        self.seqs = NvFaidx(str(fasta_file))
+
+    def adjust_interval(self, interval: GenomeInterval, context_length: int) -> GenomeInterval:
+        """Used to extend an interval by a fixed amount on both sides."""
+        start, end = interval.start, interval.end
+        interval_length = interval.end - interval.start
+        chromosome_length = self.seqs[interval.chromosome].length
+        if interval_length >= context_length:
+            return interval
+
```
skothenhill-nv, line 62: This docstring really needs help. I arrived here from the method below:

its clear context length expands the interval, and is truncated if it reaches the max. What are the bounds of sequence length that this can return? it seems like it could return any length region, which is okay, but would be best to be clear about that in the docstring.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+    ):
+        """Instantiate the class.
+
+        Args:
+            genome: Genome object
+            bed_file_df: A bed file in the form of a polars DataFrame with chrom, start, and end columns at least
+            context_length: Size of the window
+            chr_bed_to_fasta_map: Mapping between chromosome names in the bed file and the fasta file
+            **kwargs: Additional arguments
+        """
```
skothenhill-nv, line 119: I'm confused here. Context-length is used to set the window size, but the sequence returned seems to be a function of both the interval size- which could vary within the bedfile, the context length, (and potentially the end of the contig). Is this intentional? Or maybe more specifically, are there restrictions on the contents of `bed_file_df` that youre expecting?

If I were to use this dataset out of the box, depending on the model, I think I would need to transform the outputs to have a fixed size, with padding or truncation.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+    """Dataset that creates sequences based on a VCF. It needs a genome and a bed file to specify the windows."""
+
+    def __init__(
+        self,
+        vcf_file: str | Path,
+        genome_interval_dataset: GenomeIntervalDataset,
+        sample_ids: Optional[List[SampleId]] = None,
+        **kwargs: Any,
+    ):
+        """Initialize the class."""
```
skothenhill-nv, line 160: can we get docstrings for the arguments? Why is an interval dataset required? What does this do? Just guessing- its applying variants in the VCF to any position contained in the regions? Let's provide some examples in the class docstring.

skothenhill-nv: We might also want to make this class name less generic. It's a VCF dataset, but its doing something more specific- applying variants to fixed intervals.

There are models out there that are interested in windows centered around a variant (like `[NNNNNNNANNNNNN]`, which would have a difficult time benefitting from this class.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+        """Get an item from the dataset."""
+        interval_index = ind // len(self.sample_ids)
+        sample_index = ind % len(self.sample_ids)
+
+        ref_sequence = self.genome_interval_dataset.get_dna_sequence(interval_index)
+        interval = self.genome_interval_dataset.get_interval(interval_index)
+        sample_variants = read_variants_in_interval(interval, self.vcf_file, self.sample_ids)[
+            self.sample_ids[sample_index]
+        ]
+        return create_personal_sequence(ref_sequence, interval, sample_variants)
```
skothenhill-nv, line 189: somewhere we should write down that this only supports SNVs and NOT indels.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/genome_editing.py
```diff
@@ -0,0 +1,51 @@
...
+def _mutate(
+    sequence: str,
+    interval: GenomeInterval,
+    alternate: str,
+    reference: str,
+    position: int,
+) -> str:
+    """Introduce a mutation (substitution) in one or more bases of the sequence."""
+    if len(alternate) != len(reference):
+        # TO:DO Insertion or deletion to be supported later
```
skothenhill-nv, line 41: nice TODO, but fyi , indels require a multiple-sequence alignment if you want to compare them to a reference sequence. quite tricky to do this correctly and efficiently.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/genome_editing.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+import pandas as pd
+
+from bionemo.dnadl.tools.genome_interval import GenomeInterval
+
+
+def create_personal_sequence(ref_sequence: str, interval: GenomeInterval, variants: pd.DataFrame) -> str:
+    """Create a personal sequence by introducing mutations in the reference sequence."""
+    # TO:DO Implement in a more efficient way as to not bottleneck dataloaders
```
skothenhill-nv, line 24: or make the model bigger? win win haha.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/vcf.py
```diff
@@ -0,0 +1,92 @@
...
+    """Get the maximum allele index from a tuple of allele indices.
+
+    `sample.allele_indices` is a tuple of two indices that associate the sample with alleles
+    listed in the list 'variant.alleles', where the zeroth entry is the REF allele. A sample
+    could have one or both entries being 0, indicating that one or both strands are REF at
+    the specified position. By taking max(sample.allele_indices) you ensure to always extract
+    an alternative allele if present or REF otherwise.
+
+    Sometimes, None is contained in sample.allele_indices. This is treated as if it refers
+    to the REF allele (ie 0)
```
skothenhill-nv, line 38: nice docstring, super helpful!

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/vcf.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    adjusted_allele_indices = [0 if idx is None else idx for idx in allele_indices]
+    max_allele_ind = max(adjusted_allele_indices)
+    return max_allele_ind
+
+
+def _match_contig_name(chrm: str, available_contigs: set[int | str]) -> str:
+    """Let's try either the standardized version that i expect to be "chr22" or only the number string "22"."""
```
skothenhill-nv, line 46: My opinion: we should just fail in this case. It's too easy to end up with a reference that has the wrong amount of `N`s or something else stupid.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/vcf.py
```diff
@@ -0,0 +1,92 @@
...
+    """Read variants in an interval from a vcf file."""
+    with pysam.VariantFile(str(vcf_path)) as fid:
+        contig = _match_contig_name(interval.chromosome, set(fid.header.contigs))
+        variants_iter = fid.fetch(contig=contig, start=interval.start - 1, end=interval.end, reopen=True)
+        sample_to_alleles_for_interval = {sample_id: [] for sample_id in sample_ids}
+        for variant in variants_iter:
+            if variant.pos <= interval.start:
+                # TODO: sometimes the variant overlaps the boundary of a window. The fetch function is zero base
+                # so the interval works fine the problem is that sometimes a variant can indeed
+                # overlap the region even if its start is outside.
```
skothenhill-nv, line 83: :( perhaps we should make this a param? 

```
allow_partial_overlap
```

I'd set it to False by default. Also you should move this comment to the docstring.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
```diff
@@ -0,0 +1,189 @@
...
+        with pysam.VariantFile(str(self.vcf_file)) as vcf_file:
+            self.sample_ids = list(vcf_file.header.samples)
+
+    def __len__(self) -> int:
+        """Return the length of the dataset."""
+        return len(self.genome_interval_dataset) * len(self.sample_ids)
+
+    def get_dna_sequence(self, ind: int) -> str:
+        """Get an item from the dataset."""
+        interval_index = ind // len(self.sample_ids)
```
skothenhill-nv, line 181: this is a great place to surface and document all the edge-cases you handle when mutating the reference sequence.

sub-packages/bionemo-dnadl/tests/bionemo/dnadl/tools/test_genome_editing.py
```diff
@@ -0,0 +1,28 @@
...
+import pandas as pd
+
+from bionemo.dnadl.tools.genome_editing import create_personal_sequence
+from bionemo.dnadl.tools.genome_interval import GenomeInterval
+
+
+def test_mutate():
+    sequence = "AAAAAAAAAAAAAAAAAAAA"
+    interval = GenomeInterval("chr1", 1, 10)
+    variants_df = pd.DataFrame([{"POS": 5, "REF": "A", "ALT": "G"}, {"POS": 2, "REF": "AA", "ALT": "TC"}])
```
skothenhill-nv, line 26: can we add tests for the `_mutate` function directly?

I think we should also cover:
- variants that overlap region boundaries
- variants that are not allowed (indels!)

skothenhill-nv (changes requested):
Left comments almost entirely around docstrings and providing more visibility to the assumptions and limitations of these dataloaders. I love the work, and let me know if you need or want help with any of these things. Cheers.

sub-packages/bionemo-dnadl/tests/bionemo/dnadl/tools/test_genome_editing.py
cristianregep: with regards to _mutate that was the case previously but was asked to change it https://github.com/NVIDIA/bionemo-framework/pull/209#discussion_r1788097852 . I can add the corner cases at any level of abstraction, but would be useful to have a general guideline on this.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/vcf.py
cristianregep: Happy to do that but FYI internally we did this because hg19 and hg38 VCF files used different conventions https://gatk.broadinstitute.org/hc/en-us/articles/360035890951-Human-genome-reference-builds-GRCh38-or-hg38-b37-hg19 . We wanted it to work with both. Theoretically you could say that the user needs to  take care of this, but it becomes less user friendly if you get into situations where windows on top of TSSs come from Gencode which uses a specific standard.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/genome_editing.py
cristianregep: I mean all topline models will take a lot more for the forward than this, but then some of the smaller models actually have on par performance so some may start to use them https://openreview.net/pdf?id=pemixr1Isx . this was also previously discussed at some in the PR process. I can combine _create_personal_sequence and _mutate into 1 and make it faster, but that would be incompatible with added indels. I do have an implementation that just keeps track of reference sequence indeces after insertions/deletions. But sounds like you have something else in mind here https://github.com/NVIDIA/bionemo-framework/pull/209#discussion_r1884639355  that I would like to chat to you about.

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/genome_editing.py
cristianregep: see this https://github.com/NVIDIA/bionemo-framework/pull/209#discussion_r1892808679

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
cristianregep: Fair point, one can use site-only VCF for that. What do you prefer for the one with intervals GenotypedVCFDataset, or FixedIntervalsVCFDataset? I can also add a VCF dataset that looks only at the sites, each sample being centered on the variant

sub-packages/bionemo-dnadl/src/bionemo/dnadl/io/dataset.py
cristianregep: This is actually from the previous BioNeMo, but i think I know where it comes from because we also have this issue in-house. Back in the day when language modelling wasn't a thing in DNA there weren't tokenizers and tokens (e.g. Enformer). you one hot encoded the sequence. for a batch you needed to have windows of the same size, and you didn't have PAD tokens. It was more biologically relevant to include more of the actual DNA sequence than add Ns (if the context_length was longer than the intervals specified in your file), so you had to deal with this at the Dataset level, rather than at the point of collation.. The name is not vey informative, I can change to force_window_size but let's align on a strategy here.

sub-packages/bionemo-dnadl/tests/bionemo/dnadl/tools/test_genome_editing.py
skothenhill-nv: ah, i missed that. feel free to follow peter's advice, or to add the test

sub-packages/bionemo-dnadl/src/bionemo/dnadl/tools/vcf.py
skothenhill-nv: fine with me, per our call- add a warning if the seqids dont match between the vcf and reference.

PR #274, opened 2024-10-08
Justfile improvements

malcolmgreaves:
This PR adds several small improvements to the `justfile`. The `build-{dev,release}` commands will now skip the image build step if there is already an image with the same tag present locally. 

The Justfile's `setup` command now logs the user into the NGC image repository. Login uses the `NGC_API_KEY` value from the user's `.env` file.


tshimko-nv (approved):

pstjohn (approved):
LGTM!

I personally prefer mounting my local env files into the container rather than using the `.env` file -- maybe we could support that workflow sometime in down the road?

malcolmgreaves:
For sure @pstjohn -- one thing that is convenient about justfiles is they can load `.env` files as variables into the justfile itself. This made it easier to adapt for our existing stuff. But I hear you that you want an option for volume mounting the file.

PR #304, opened 2024-10-11
ESM2 Infer partial batches using predict method

farhadrgh:
This depends on changes in https://github.com/NVIDIA/NeMo/pull/10837 to expose `drop_last` in `MegatronDataSampler` and allow inference of partial batches.

Note:

The https://github.com/NVIDIA/NeMo/pull/10934 has overridden the changes to expose drop_last. We should now wrap dataloaders with nemo.lightning.data.WrappedDataLoader that can store the mode attribute when creating the dataloader in datamodules. Then drop_last=False if dataloader is in test or predict mode.


farhadrgh:
The https://github.com/NVIDIA/NeMo/pull/10934 has overridden the  changes to expose `drop_last`. We should now wrap dataloaders with `nemo.lightning.data.WrappedDataLoader` that can store the `mode` attribute when creating the dataloader in datamodules. Then `drop_last=False` if dataloader is in `test` or `predict` mode.

CC @jstjohn

pstjohn (approved):
LGTM, but where does this use the new `drop_last` argument?

farhadrgh:
> LGTM, but where does this use the new `drop_last` argument?

`drop_last` is no longer exposed. Megatron Sampler set it to false when dataloader mode is test/predict: https://github.com/NVIDIA/NeMo/pull/10934/files#diff-c423cc7981d5621754bb3e5e509fceaa9adb8537b269b7def042995cfa21d529R81

PR #362, opened 2024-10-25
Match resumption loss curve

sichu2023:
## Summary
Add additional resumption mechanism to datamodule to ensure stop-and-go and continuous pretraining has the same training loss curve.

## Details
Still has a small discrepancy right at resumption but does not affect subsequent training/validation curve.
[WANDB log](https://wandb.ai/clara-discovery/bionemo2-esm2-debug/reports/val_ppl-24-10-25-20-01-48---Vmlldzo5ODg4ODA1)

## Testing
To be implemented in subsequent PR on stop-and-go test refactoring.


sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
```diff
@@ -18,19 +18,18 @@
...
 from nemo.utils import logging
 from pytorch_lightning.utilities.types import EVAL_DATALOADERS, TRAIN_DATALOADERS
 
 from bionemo.esm2.data import dataset, tokenizer
 from bionemo.llm.data import collate
+from bionemo.llm.data.datamodule import MegatronDatamodule
 from bionemo.llm.utils.datamodule_utils import infer_num_samples
 
 
-class ESMDataModule(pl.LightningDataModule):
```
farhadrgh, line 33: Can you comment on why is this change needed?

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
```diff
@@ -172,11 +171,13 @@ def setup(self, stage: str = "") -> None:
...
-    def _create_dataloader(self, dataset, **kwargs) -> torch.utils.data.DataLoader:
+    def _create_dataloader(self, dataset, **kwargs) -> WrappedDataLoader:
         assert self._tokenizer.pad_token_id is not None, "Tokenizer must have a pad token id."
+        self.init_global_step = self.trainer.global_step
+        self.data_sampler.init_global_step = self.init_global_step
 
-        return torch.utils.data.DataLoader(
-            dataset,
+        return WrappedDataLoader(
+            dataset=dataset,
```
jstjohn, line 180: You forgot to pass in `mode`

jstjohn: OOOOOH I see it sneaks in through `kwargs`. I have a pretty strong readability preference to expose this and pass it through since it's so important.

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
farhadrgh: nvm I thought its an import from megatron

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
jstjohn: So just add like a `mode:Literal[whatever]` arg and then pass `mode` into the `WrappedDataLoader` along with a comment saying "mode is used to determine if consumed_samples in the sampler should be initialized to 0 (non-train loaders) or be set to the previous consumed samples in the case of checkpoint resumption.` or something. I think over-communicating this is important for users who copy/paste code to see for when they modify.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/datamodule.py
```diff
@@ -174,14 +172,15 @@ def train_dataloader(self) -> TRAIN_DATALOADERS:  # noqa: D102
...
 
     def test_dataloader(self) -> EVAL_DATALOADERS:  # noqa: D102
         return self._create_dataloader(self._test_ds)
 
-    def _create_dataloader(self, dataset, **kwargs) -> DataLoader:
-        return DataLoader(
-            dataset,
+    def _create_dataloader(self, dataset, **kwargs) -> WrappedDataLoader:
+        self.update_init_global_step()
+        return WrappedDataLoader(
```
jstjohn, line 182: Same comment here to expose `mode`

scripts/protein/esm2/esm2_pretrain.py
```diff
@@ -380,6 +381,19 @@ def main(
...
+    required=False,
+    help="Number of steps between logging. Default is 50.",
+)
+parser.add_argument(
+    "--save-every-n-steps",
+    type=int,
+    required=False,
+    default=10000,
+    help="Number of steps between saving checkpoints. Default is 10000.",
+)
```
jstjohn, line 396: Generally I think the val_check_interval is a good default for `save-ever-n-steps` since then when you go through the process of validation, you also save the checkpoint if it's worth saving. Otherwise why are we pausing to validate if we don't have a record of the model? I'd personally prefer getting rid of this and having val_interval == save_interval. Log interval is great though!

scripts/protein/esm2/esm2_pretrain.py
```diff
@@ -55,6 +55,7 @@ def main(
     warmup_steps: int,
     limit_val_batches: int,
     val_check_interval: int,
+    log_every_n_steps: Optional[int],
```
jstjohn, line 58: I see this new arg, which I love the idea of, but I don't see where it's used. I would guess you should pass it into the NeMoLogger thing?

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
jstjohn: Yeah, this is super important. If you do not do this and you restart your training job it will go back to step 0 rather than picking up from where it left off! You also need to call that function when you return the dataloader. So brittle. We need to heavily document this!

jstjohn: Yeah I called this `MegatronDataModule` because of how it's important in the context of the megatron strategy specifically. This DataModule class is needed to expose the functionality and state that is critical for enabling the megatron strategy to pick up where it left off in terms of sample index during training.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
```diff
@@ -0,0 +1,69 @@
...
+# limitations under the License.
+
+
+from typing import Any, Dict
+
+import pytorch_lightning as pl
+from nemo.utils import logging
+
+
+class MegatronDatamodule(pl.LightningDataModule):
```
jstjohn, line 23: Let's rename this `MegatronDataModule`. Also @ashors1 might want to use this in NeMo?

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
```diff
@@ -0,0 +1,69 @@
...
+    """A mixin that adds a `state_dict` and `load_state_dict` method for datamodule training resumption in NeMo."""
+
+    def __post_init__(self):
+        """Set init_global_step to 0 for datamodule resumption."""
+        self.init_global_step = 0
```
jstjohn, line 28: @sichu2023 this is a cool idea, but I don't think `__post_init__` gets called outside of in `@datamodule` classes? Double check that.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
jstjohn: To be clear if it gets moved to NeMo we can add a future PR to remove it on our end and use the NeMo version. Not saying to do this right now as part of this PR.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/datamodule.py
```diff
@@ -191,7 +193,6 @@ def _create_dataloader(self, dataset, **kwargs) -> DataLoader:
                 min_length=None,
                 max_length=self.max_len,
             ),
-            **kwargs,
```
jstjohn, line 194: You can still pass `**kwargs` through incase there's anything else a user wants to pass in? Not saying we should remove this.

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
```diff
@@ -186,16 +196,15 @@ def _create_dataloader(self, dataset, **kwargs) -> torch.utils.data.DataLoader:
                 min_length=self._min_seq_length,
                 max_length=self._max_seq_length,
             ),
-            **kwargs,
```
jstjohn, line 189: Same comment here, you can still pass `**kwargs` through for future usability.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
pstjohn, line 28: yeah just return to
```suggestion
    def __init__(self):
        """Set init_global_step to 0 for datamodule resumption."""
        super().__init__()
        self.init_global_step = 0
```

scripts/protein/esm2/esm2_pretrain.py
```diff
@@ -598,7 +606,7 @@ def main(
         save_last_checkpoint=args.save_last_checkpoint,
         metric_to_monitor_for_checkpoints=args.metric_to_monitor_for_checkpoints,
         save_top_k=args.save_top_k,
-        save_every_n_steps=args.val_check_interval,
+        save_every_n_steps=args.save_every_n_steps,
```
jstjohn, line 609: shouldn't this change? I thought you just deleted the save_every_n_steps arg?

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
sichu2023: Renamed.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def update_init_global_step(self):
+        """Please always call this when you get a new dataloader... if you forget, your resumption will not work."""
```
pstjohn, line 31: is there no way we can automate this call?

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
sichu2023: Ya. This is only for `dataclass`.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
jstjohn: I tried moving it to a few different places, it seems like it needs to happen after `setup`. I'm guessing the correct fix is to add a new megatron callback hook at the right point of initialization and have this automatically called on the `trainer.datamodule` and require the datamodule type to support this, or if it doesn't at _least_ log a warning. 

This is SUPER HACKY as is now, and should be handled upstream by NeMo. @marcromeyn @ashors1.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
sichu2023: Just did the same fix but with `*args` and `**kwargs`.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
jstjohn: So to be clear I think the megatron strategy should prefer to work with a couple key objects and warn if they are not encountered:

1. The DataloaderWrapper thing that communicates the `mode`. There should be a warning if the mode is not found with an explanation of how resumption is either going to go wrong on the trainer or val steps.
2. The DataModule piece needing state, and this global step communication from the trainer. The strategy should warn if a vanilla datamodule is encountered, and use some special datamodule type that exposes these key features for resumption.

sichu2023: I am still unfamiliar with its mechanism. Probably @jstjohn can answer more.

sichu2023: Should be something I have missed.

jstjohn: Sure, so if you forget to update `self.init_global_step` _and_ `self.data_sampler.init_global_step` then the `consumed_samples` calculations are way off. They get inflated and a bunch of samples are skipped on resumption.

jstjohn: It's ok to call multiple times apparently.

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
```diff
@@ -18,19 +18,21 @@
 import os
 from typing import Literal
 
-import pytorch_lightning as pl
-import torch
-import torch.utils.data
+from nemo.lightning.data import WrappedDataLoader
```
sichu2023, line 21: I wonder if we should do something like this instead.
`import nemo.lightning.data.WrappedDataLoader as MegatronDataLoader`

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
jstjohn: maybe we can try moving this into `def train_dataloader` or something.

sub-packages/bionemo-esm2/src/bionemo/esm2/data/datamodule.py
sichu2023: @jstjohn

sichu2023: Just to be consistent with `MegatronDataModule`.

sub-packages/bionemo-llm/src/bionemo/llm/data/datamodule.py
jstjohn: unless it's required to be called in `def val_dataloader`. I'm not sure exactly where this needs to be called to be honest... Just that it needs to happen after `setup` at some point.

scripts/protein/esm2/esm2_pretrain.py
```diff
@@ -183,6 +183,7 @@ def main(
         strategy=strategy,
         limit_val_batches=limit_val_batches,  # This controls upsampling and downsampling
         val_check_interval=val_check_interval,
+        log_every_n_steps=log_every_n_steps,
```
jstjohn, line 186: You need to add this in one other spot to get logging to be less frequent, in your `nl.MegatronStrategy` above add ```
        log_interval=log_every_n_steps,
```

jstjohn (approved):
Approved with one more comment about reducing the logging interval. Thanks!

Also note I pushed changes that fix geneformer to `sichu/fix-pplcallback-sanity`

sichu2023:
./build-ci

PR #371, opened 2024-10-28
better error handling in SCDL collections.

polinabinder1:
With the async worker queue that is used to read in multiple h5ad files, errors in a specific process are not properly raised and propagated. Additionally, if there is an empty directory, a correct error message is not raised. 
## Testing
I added two test cases. One creates an empty directory and checks that the appropriate error is raised. In the other test case, and empty h5ad file is created which should raise an error when it is read in.


jstjohn (approved):

malcolmgreaves (approved):
Excellent work 👏

PR #373, opened 2024-10-28
Changes to get toward better training perf in geneformer

jstjohn:
## Summary
Pull in the latest defaults from NeMo after their optimization work. Also do nsys profiling. Speed up the training loop and generate wandb runs showing reasonable proximity.

## Details
* Added cluster gradient synchronization settings + options
* Added garbage collection synchronization option
* Found a 10% performance boost on single device by not doing the transpose of logits before computing the loss.


scripts/singlecell/geneformer/train.py
```diff
@@ -209,6 +226,7 @@ def main(
         num_workers=num_dataset_workers,
     )
     geneformer_config = config_class(
+        # TODO let users set different num layers/model shapes here to support bigger/smaller architectures
```
pstjohn, line 229: duplicate TODO?

pstjohn (approved):

sub-packages/bionemo-geneformer/src/bionemo/geneformer/api.py
```diff
@@ -34,6 +35,18 @@
 GeneformerModel = MegatronBioBertModel
 
 
+class BERTMLMLossWithReductionNoForward(BERTMLMLossWithReduction):
```
sichu2023, line 38: In the future, we need to add a mechanism to pass options into a configurable `LossReductionClass`. Right now we have to override the class/use the default.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py
```diff
@@ -684,7 +684,8 @@ def _get_loss_from_model(model_config: GeneformerConfig, seed: int) -> float:
                 attention_mask=batch["attention_mask"].cuda(),
             )
             loss_mask = batch["loss_mask"].cuda()
-            logits = result["token_logits"]
+            # token_logits is s,b and for simplicity here let's transpose to b,s. In general this reduces performance.
+            logits = result["token_logits"].transpose(0, 1).contiguous()
             target = batch["labels"].cuda()
 
             loss += F.cross_entropy(logits[loss_mask].float(), target[loss_mask], reduction="sum")
```
sichu2023, line 691: How about reusing `unreduced_token_loss_fn` if we trust that function?

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -289,34 +289,34 @@ def process_item(  # noqa: D417
...
-    )
-
-    if prepend_cls_token:
-        masked_tokens, labels, loss_mask = masking.add_cls_and_eos_tokens(
-            sequence=masked_tokens,
-            labels=labels,
-            loss_mask=loss_mask,
-            cls_token=tokenizer.token_to_id(tokenizer.cls_token),
-            eos_token=None,
+    with torch.no_grad(), torch.device("cpu"):
```
sichu2023, line 292: Didn't know `torch.device` can be used as context manager!

sichu2023 (approved):

scripts/singlecell/geneformer/train.py
```diff
@@ -173,11 +219,8 @@ def main(
         val_check_interval=val_check_interval,  # TODO(@jstjohn) Checkpoint saving is currently broken, fix and change this.
         log_every_n_steps=log_every_n_steps,
         num_nodes=num_nodes,
-        callbacks=[
-            PerplexityLoggingCallback(log_train=False, log_val=True),  # TODO(@sichu) fix this
-            RichModelSummary(max_depth=4),
-            LearningRateMonitor(),
-        ],
+        callbacks=callbacks,
+        use_distributed_sampler=False,
```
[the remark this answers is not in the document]
jstjohn: I pulled it from NeMo. Hopefully nothing, but it basically tells the lightning trainer not to try to wrap your dataset sampler in a "DistributedSampler" object. We have our megatron sampler which is already distributed.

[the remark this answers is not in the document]
jstjohn: It's there for peace of mind I guess?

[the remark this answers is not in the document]
sichu2023: I guess it isn't the first time we are seeing potentially duplicated options.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py
jstjohn: I would then have to sum reduce myself? I'd rather leave it like this for now.

scripts/singlecell/geneformer/train.py
```diff
@@ -138,14 +149,28 @@ def main(
         tensor_model_parallel_size=tensor_model_parallel_size,
         pipeline_model_parallel_size=pipeline_model_parallel_size,
     )
+    if aligned_megatron_ddp:
+        ddp: str | DistributedDataParallelConfig = DistributedDataParallelConfig(
```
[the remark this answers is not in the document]
jstjohn: I tried to comment it in the else block. What "megatron" does behind the scenes is get changed into:
```# this will launch DistributedDataParallelConfig(check_for_nan_in_grad=True).```

So same thing but with different settings. I pulled this version with these settings from many of the nemo2 recipes. It doesn't seem to make a big difference, but maybe it will if we get past whatever other cluster issues are slowing us down.

skothenhill-nv (approved):

PR #374, opened 2024-10-29
Export FUSED_ATTN in release container and fix geneformer stop and go

pstjohn:
Looks like for some reason, the output tensors are not consistent between the interrupted and continuous training runs without `NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0` set in the release container. These are set in the development container though, so we should keep these consistent.

Possibly the model isn't as deterministic without these set?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -204,4 +204,5 @@ COPY ./docs ./docs
 RUN chmod 777 -R /workspace/bionemo2/
 
 # Transformer engine attention defaults
+# We have to declare this again because the devcontainer splits from the release image's base.
 ENV NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0
--- a/Dockerfile
+++ b/Dockerfile
@@ -202,3 +202,6 @@ COPY ./ci/scripts ./ci/scripts
 COPY ./docs ./docs
 
 RUN chmod 777 -R /workspace/bionemo2/
+
+# Transformer engine attention defaults
+ENV NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0
--- a/sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_stop_and_go.py
+++ b/sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_stop_and_go.py
@@ -28,7 +28,6 @@ import math
 import pathlib
 from typing import Literal
 
-import pytest
 import pytorch_lightning as pl
 import torch
 from megatron.core.optimizer.optimizer_config import OptimizerConfig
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -202,3 +202,6 @@ COPY ./ci/scripts ./ci/scripts
 COPY ./docs ./docs
 
 RUN chmod 777 -R /workspace/bionemo2/
+
+# Transformer engine attention defaults
```
jstjohn, line 206: Can you add a comment saying why this needs to be declared twice? Thanks!

malcolmgreaves:
@pstjohn I believe you can just keep the `ENV` statement in the `base` target image and they will be propagated to all other target images that extend `FROM` it. Are you not seeing this behavior? 🤔 


I was curious about how this works so I put together a small demonstration:

==> Dockerfile:
```
FROM python:3.10-slim as base

ENV foobar=1

FROM base as dev1

WORKDIR /workspace/dev1
ENV foobar=5
RUN touch hello_world_from_dev_1
RUN echo "${foobar}" > hello_world_from_dev_1

FROM base as dev2

WORKDIR /workspace/dev2
RUN touch hello_world_from_dev_2
RUN echo "${foobar}" > hello_world_from_dev_2
```



The base image has the `foobar` env var as 1:
```
$ docker build -f Dockerfile -t base --target base .
...

$ docker run --rm -it --entrypoint /bin/bash base
root@1a1dfa31f351:/# ls
bin  boot  dev	etc  home  lib	media  mnt  opt  proc  root  run  sbin	srv  sys  tmp  usr  var
root@1a1dfa31f351:/# echo $foobar
1
```

Building the `dev1` image has the changed `foobar` env var:
```
$ docker build -f Dockerfile -t hello_dev1 --target dev1 .
...

$ docker run --rm -it --entrypoint /bin/bash hello_dev1
root@5aee42a14461:/workspace/dev1# cat hello_world_from_dev_1
5
root@5aee42a14461:/workspace/dev1# echo $foobar
5
```

But `dev2` has the original `foobar` value it inherits from `base`:
```
$ docker build -f Dockerfile -t hello_dev2 --target dev2 .
...

$ docker run --rm -it --entrypoint /bin/bash hello_dev2
root@02061bc61729:/workspace/dev2# cat hello_world_from_dev_2
1
root@02061bc61729:/workspace/dev2# echo $foobar
1
```

pstjohn:
> Approving to unblock, but @pstjohn I think it would be better to put `ENV NVTE_FUSED_ATTN=1 NVTE_FLASH_ATTN=0` into the base image. If it poses a problem for building dependencies in the build image, then I believe it'd be safe to put it at the end of the base image target section.
> 
> More details here: [#374 (comment)](https://github.com/NVIDIA/bionemo-framework/pull/374#issuecomment-2444810752)

The only noodle here is https://github.com/NVIDIA/bionemo-framework/blob/7f5c2c04477d76dd113574cebbb0185cb3a2858b/Dockerfile#L113 where the dev container doesn't actually extend the same image as the release. We'd still need multiple ENV declarations, but we could move this further up in the container image. My only thought was that by putting it last, we don't invalidate caches just to set a couple env variables

jstjohn (approved):

PR #375, opened 2024-10-29
Move Geneformer dataset to MultiEpochDatasetResampler

pstjohn:
Moves the remaining changes from #312 into a fresh PR

### DIFF
--- a/sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
+++ b/sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
@@ -23,6 +23,7 @@ import torch
 from nemo.utils import logging
 from torch.utils.data import Dataset
 
+from bionemo.core.data.multi_epoch_dataset import EpochIndex
 from bionemo.core.utils import random_utils
 from bionemo.geneformer.data.singlecell.utils import sample_or_truncate
 from bionemo.geneformer.tokenizer.gene_tokenizer import GeneTokenizer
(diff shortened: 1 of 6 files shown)


sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -193,11 +194,11 @@ def lookup_cell_by_idx(self, idx) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
         )
         return gene_data, col_idxs, feature_ids
 
-    def __getitem__(self, idx: int) -> types.BertSample:  # noqa: D105
-        rng = np.random.default_rng([self._seed, idx])
+    def __getitem__(self, index: EpochIndex) -> types.BertSample:  # noqa: D105
+        rng = np.random.default_rng([self._seed, index.epoch, index.idx])
```
sichu2023, line 198: Is `IdentityMultiEpochDatasetWrapper` relevant?

sichu2023 (approved):
LGTM

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
jstjohn: I think not, since we want to pass the rng state into process_item

jstjohn (approved):

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
pstjohn: Thanks for flagging this wierd docstring -- fixed! On `IdentityMultiEpochDatasetWrapper` though, no -- we need that epoch index to increment the random state. You might be able to refactor it to use the `apply_transform` method of `MultiEpochDatasetWrapper`, but it's pretty easy to just set the rng here based on the epoch index

PR #377, opened 2024-10-29
Add docs on MegatronDataModule and WrappedDataLoader

sichu2023:


docs/docs/user-guide/background/megatron_datasets.md
```diff
@@ -56,6 +56,42 @@ for sample in MultiEpochDatasetResampler(dataset, num_epochs=3, shuffle=True):
...
+## Training Resumption
+To ensure identical behavior with and without job interruption, BioNeMo provides [MegatronDataModule][bionemo.llm.data.datamodule.MegatronDataModule] to save and load state dict for training resumption, and provides [WrappedDataLoader][nemo.lightning.data.WrappedDataLoader] to add a `mode` attribute to [DataLoader][torch.utils.data.DataLoader].
+
+```python
+class MyDataModule(MegatronDataModule):
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+        ...
+
+    def train_dataloader(self):
```
jstjohn, line 68: TODO add `self.update_init_global_step()` just before the WrappedDataloader return.

docs/docs/user-guide/background/megatron_datasets.md
```diff
@@ -56,6 +56,42 @@ for sample in MultiEpochDatasetResampler(dataset, num_epochs=3, shuffle=True):
...
+        return WrappedDataLoader(
+            ...,  # any other arguments for DataLoader
+            mode="train",
+        )
+
+    def val_dataloader(self):
```
jstjohn, line 74: Same here on `self.update_init_global_step()`

docs/docs/user-guide/background/megatron_datasets.md
```diff
@@ -56,6 +56,42 @@ for sample in MultiEpochDatasetResampler(dataset, num_epochs=3, shuffle=True):
...
+    def test_dataloader(self):
+        return WrappedDataLoader(
+            ...,  # any other arguments for DataLoader
+            mode="test",
+        )
+```
+
+!!! note "MegatronDataModule"
+
+    Users will see non-overlapping training curve if their datamodule is not inheritting from `MegatronDataModule`, unless similar logics are handled by the users. `MegatronDataModule` ensures that training resumes with the correct sample index instead of restarting from 0 everytime.
```
jstjohn, line 89: Note that you also need to call `self.update_init_global_step()` and call out that this needs to happen when the dataloaders are requested. Unfortunately we do not yet precisely know where it needs to go but this is the current pattern that works in our code and nemo code.

docs/docs/user-guide/background/megatron_datasets.md
```diff
@@ -56,6 +56,42 @@ for sample in MultiEpochDatasetResampler(dataset, num_epochs=3, shuffle=True):
...
+
+!!! note "WrappedDataLoader"
+
+    The WrappedDataLoader class is a wrapper around the PyTorch DataLoader class that adds the `mode` attribute to the dataloader. The dataloader will resume from the last sample index only when mode is 'train'. `val_dataloader` and `test_dataloader` are unaffected.
```
jstjohn, line 93: Also note that `train` is the default mode. So if you forget to return a wrapped dataloader than your val/test classes will change every time you restart since their consumed_samples will be set to a non-zero value.

jstjohn (approved):
Approved with a few comments. Thanks!

docs/docs/user-guide/background/megatron_datasets.md
sichu2023: Addressed.

docs/docs/user-guide/background/megatron_datasets.md
sichu2023: Addressed.

docs/docs/user-guide/background/megatron_datasets.md
```diff
@@ -1,4 +1,4 @@
-# Writing Megatron-LM compatible datasets
+# Writing Megatron-LM compatible datamodules
```
tshimko-nv, line 1: Can you convert this and other headings to title-case? (in this case, just capitalize Compatible Datamodules as well.

docs/docs/user-guide/background/megatron_datasets.md
```diff
@@ -56,6 +56,47 @@ for sample in MultiEpochDatasetResampler(dataset, num_epochs=3, shuffle=True):
...
+
+    Users will see non-overlapping training curve if their datamodule is not inheritting from `MegatronDataModule`, unless similar logics are handled by the users. In `MegatronDataModule`, `self.update_init_global_step()` must be called right before the dataloaders are returned to ensure that training resumes with the correct sample index instead of restarting from 0 everytime.
+
+!!! note "WrappedDataLoader"
+
+    The `WrappedDataLoader` class is a wrapper around the PyTorch DataLoader class that adds the `mode` attribute to the dataloader. The dataloader will resume from the last sample index only when mode is 'train'. `val_dataloader` and `test_dataloader` are unaffected.
+
+    WARNING: 'train' is the default value of `mode` in `WrappedDataLoader`. If not set, users might find their validation/test dataloader changes behavior by resuming from a non-zero sample index.
+
 ## Testing datasets for megatron compatibility
```
tshimko-nv, line 100: Title case

tshimko-nv (approved):
Looks good, please change all headings to title case. Otherwise approved with minor comments.

pstjohn (approved):
Thanks for adding these!

sichu2023:
./build-ci

PR #379, opened 2024-10-29
Savitha/integrate scdl into geneformer megatron compatible

savitha-eng:
## Summary
In this PR we refactor the Geneformer `SingleCellDataset` class  to integrate the `SingleCellMemmapDataset`(SCDL). The goal of this is to streamline and increase readability of the dataset class. 
## Details
We make the following changes:
- Input Format: 
     - The `SingleCellDataset` now assumes that the input path to the data is a directory formatted in the `SingleCellMemmap` format. 
     -  The SingleCellModule now assumes that the train, val, and test input paths are to directories that are formatted in the `SingleCellMemmap` format 
 - Get Item:
     - `_get_item()` now leverages the get_row function from SCDL (so we eliminate the need to store and parse information in metadata.json)
- Error Handling for Genes not in the Tokenizer Vocabulary:
   - We add an optional parameter to SingleCellDataset and SingleCellDataModule called `bypass_tokenizer_vocab` which is by default `False`. So by default, we throw an error if a gene ID is not in the tokenizer vocabulary. If a user wants to bypass this, they can change `bypass_tokenizer_vocab` to `True`.
-  Error Handling for Genes with Zero Expression Values:
   - We throw an invalid input error in the cases that certain cells have no gene expression values (i.e. `sc_dataset.scdl.get_item()` returns `[]` for the gene data value) 
   
## Usage
The main change from a user perspective is to ensure that they convert their single cell h5ad files (or directories of h5ad files) to SingleCellMemmap format. 
1) For a single h5ad file, i.e. `data.h5ad`, they can simply run the following, where `output_path` is the file path the SingleCellMemmap directory should be written to:
`    SingleCellMemMapDataset(output_path, data.h5ad) `
2) For a directory of h5ad files, they can run the following, where `output_path` is the file path the `SingleCellMemmap` directory should be written to and `test_input` is the path to the h5ad directory. 
                        ```
                        with tempfile.TemporaryDirectory() as temp_dir:
                                coll = SingleCellCollection(temp_dir)
                                coll.load_h5ad_multi(test_input, max_workers=4, use_processes=False)
                                coll.flatten(output_path, destroy_on_copy=True)
                        ```
 For more information on this they can refer to the `convert_h5ad_to_scdl` script and the SCDL ReadMe. 

## Testing
We test that the updated SingleCellDataset produces the same output as the old dataset on synthetic samples and samples from the cellxsmall dataset. We also test for Megatron compatibility (as this dataset uses the MultiEpochDatasampler / Epoch Index) and for correct error handling of the above cases. 
```shell
pytest -vsub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_dataset.py
```
Note that we have also updated the following test files to use the MemMap dataset format + set bypass_tokenizer_vocab=True in them, because the cellxsmall dataset does have a few genes not in the HuggingFace tokenizer vocab and so the tests will error otherwise:
`sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py`
`scripts/singlecell/geneformer/test_train.py`
`sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_stop_and_go.py`

### DIFF
--- a/scripts/singlecell/geneformer/test_train.py
+++ b/scripts/singlecell/geneformer/test_train.py
@@ -29,10 +29,17 @@ from bionemo.testing import megatron_parallel_state_utils
 from bionemo.testing.data.load import load
 
 
-data_path: Path = load("single_cell/testdata-memmap-format") / "cellxgene_2023-12-15_small_mmap"
+@pytest.fixture
+def data_path() -> Path:
+    """Gets the path to the directory with with cellx small dataset in Single Cell Memmap format.
+
+    Returns:
+        A Path object that is the directory with the specified test data.
+    """
+    return load("single_cell/testdata-memmap-format") / "cellxgene_2023-12-15_small_mmap"
 
 
-def test_bionemo2_rootdir():
+def test_bionemo2_rootdir(data_path):
     data_error_str = (
         "Please download test data with:\n"
         "`python scripts/download_artifacts.py --models all --model_dir ./models --data all --data_dir ./ --verbose --source pbss`"
@@ -42,7 +49,7 @@ def test_bionemo2_rootdir():
 
 
 @pytest.mark.skip("duplicate unittest")
-def test_main_runs(tmpdir):
+def test_main_runs(tmpdir, data_path):
     result_dir = Path(tmpdir.mkdir("results"))
 
     with megatron_parallel_state_utils.distributed_model_parallel_state():
@@ -86,7 +93,7 @@ def test_main_runs(tmpdir):
     ).is_file(), "Could not find experiment log."
 
 
-def test_pretrain_cli(tmpdir):
+def test_pretrain_cli(tmpdir, data_path):
     result_dir = Path(tmpdir.mkdir("results"))
     open_port = find_free_network_port()
     # NOTE: if you need to change the following command, please update the README.md example.
--- a/scripts/singlecell/geneformer/test_train.py
+++ b/scripts/singlecell/geneformer/test_train.py
@@ -29,7 +29,7 @@ from bionemo.testing import megatron_parallel_state_utils
 from bionemo.testing.data.load import load
 
 
-data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+data_path: Path = load("single_cell/testdata-memmap-format") / "cellxgene_2023-12-15_small_mmap"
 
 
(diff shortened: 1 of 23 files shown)


scripts/singlecell/geneformer/test_train.py
```diff
@@ -29,7 +29,7 @@
 from bionemo.testing.data.load import load
 
 
-data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+data_path: Path = load("single_cell/testdata-memmap-format") / "cellxgene_2023-12-15_small_mmap"
```
pstjohn, line 32: Could you actually put this under a pytest fixture?

```suggestion
@pytest.fixture
def data_path() -> Path:
    return load("single_cell/testdata-memmap-format") / "cellxgene_2023-12-15_small_mmap"
```

You'd then also need to add `data_path` as an argument to functions that use it

sub-packages/bionemo-testing/src/bionemo/testing/data/resources/single_cell.yaml
```diff
@@ -21,3 +21,9 @@
   sha256: ab038b184de52e53ff7bcea5e01d97d55944c507db88c0495bdf9e5e9e0303a4 # pragma: allowlist secret
   owner: John St John <jstjohn@nvidia.com>
   description: Golden values for geneformer QA model.
+
+- tag: testdata-memmap-format
+  pbss: "s3://bionemo-ci/test-data/singlecell/ordered_singlecell_mmap_test_data.tar.gz"
+  sha256: d1d3da1c831877100e55b326e4588ffe305638a06c77415468344bd63ea3657c # pragma: allowlist secret
+  owner:  Savitha Srinivasan <savithas@nvidia.com>
+  description: Test data for single cell models in SCDL Memmap format.
```
pstjohn, line 29: Is there a plan to upload this to NGC? Not sure how this works now that we're public, any tests that use this would fail for non-NV folks

pstjohn (approved):

sub-packages/bionemo-testing/src/bionemo/testing/data/resources/single_cell.yaml
savitha-eng: Yes! I'm working with Dorota on this.

malcolmgreaves (approved):

PR #382, opened 2024-10-30
use torchmetrics ppl logging

pstjohn:
Some experiments using torchmetrics for perplexity logging.

Example wandb chart for single-GPU training: https://wandb.ai/clara-discovery/bionemo-hf-pretraining/runs/fgn32r4p?nw=nwuserpstjohn

### DIFF
--- a/sub-packages/bionemo-llm/src/bionemo/llm/data/collate.py
+++ b/sub-packages/bionemo-llm/src/bionemo/llm/data/collate.py
@@ -101,7 +101,7 @@ def bert_padding_collate_fn(
         "text": padding_value,
         "types": 0,
         "attention_mask": False,
-        "labels": -1,
+        "labels": -100,
         "loss_mask": False,
         "is_random": 0,
     }
--- a/sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
+++ b/sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
@@ -17,6 +17,7 @@ from typing import Any, Callable, Dict, Generic, Iterable, Iterator, List, Optio
 
 import pytorch_lightning as pl
 import torch.distributed
+import torchmetrics.text
 from megatron.core import parallel_state
 from megatron.core.optimizer.optimizer_config import OptimizerConfig
 from nemo.lightning import io as nlio
@@ -258,6 +259,8 @@ class BionemoLightningModule(
         self._data_step = data_step
         self._forward_step = forward_step
         self.model_transform = model_transform
+        self.train_ppl = torchmetrics.text.Perplexity(ignore_index=-100)
+        self.valid_ppl = torchmetrics.text.Perplexity(ignore_index=-100)
 
     def configure_model(self) -> None:
         """Updates internal state: instantiates the model from the object's config, assigns to `model` attribute.
@@ -305,11 +308,19 @@ class BionemoLightningModule(
 
     def training_step(self, batch, batch_idx: Optional[int] = None) -> Tensor:
         """In mcore the loss-function is part of the forward-pass when labels are provided."""
-        return self.forward_step(batch)
+        outputs = self.forward_step(batch)
+        logits = outputs["token_logits"].transpose(0, 1)  #  [s, b] -> [b, s]
+        self.train_ppl(logits, batch["labels"])
+        self.log("train_ppl_step", self.train_ppl)
+        return outputs
 
     def validation_step(self, batch, batch_idx: Optional[int] = None) -> Tensor:
         """In mcore the loss-function is part of the forward-pass when labels are provided."""
-        return self.forward_step(batch)
+        outputs = self.forward_step(batch)
+        logits = outputs["token_logits"].transpose(0, 1)  #  [s, b] -> [b, s]
+        self.valid_ppl(logits, batch["labels"])
+        self.log("valid_ppl_step", self.valid_ppl)
+        return outputs
 
     def predict_step(self, batch, batch_idx: Optional[int] = None) -> Tensor:
         """Alias for forward_step."""
@@ -324,6 +335,14 @@ class BionemoLightningModule(
 
     def test_loss_reduction(self) -> MegatronLossType:  # noqa: D102
         return self.loss_reduction_class(validation_step=True)
+
+    def on_train_epoch_end(self):
+        """Log perplexity at the end of the epoch."""
+        self.log("train_ppl_step", self.train_ppl)
+
+    def on_valid_epoch_end(self):
+        """Log perplexity at the end of the epoch."""
+        self.log("valid_ppl_step", self.valid_ppl)
 
 
 def default_megatron_optimizer() -> MegatronOptimizerModule:


sub-packages/bionemo-llm/src/bionemo/llm/data/collate.py
```diff
@@ -78,7 +78,7 @@ def bert_padding_collate_fn(
         "text": padding_value,
         "types": 0,
         "attention_mask": False,
-        "labels": -1,
+        "labels": -100,
```
pstjohn, line 81: whoops 😄 . We should fix this, currently labels are padded with both -1 and -100

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
```diff
@@ -275,11 +278,17 @@ def forward_step(self, batch) -> Tensor:
...
+        self.log("train_ppl_step", self.train_ppl)
+        return outputs
 
     def validation_step(self, batch, batch_idx: Optional[int] = None) -> Tensor:
         """In mcore the loss-function is part of the forward-pass when labels are provided."""
-        return self.forward_step(batch)
+        outputs = self.forward_step(batch)
+        self.valid_ppl(outputs["token_logits"], batch["labels"])
+        self.log("valid_ppl_step", self.valid_ppl)
+        return outputs
```
pstjohn, line 291: Where does `validation_step` get called and how does this interact with megatron parallelism? Just wondering whether torchmetrics would correctly detect the parallelism settings and collect data from all ranks with this configuration

pstjohn:
Running with `tp=2`, I get the following error:

```
  File "/usr/local/lib/python3.10/dist-packages/bionemo/llm/lightning.py", line 319, in validation_step
    self.valid_ppl(outputs["token_logits"], batch["labels"])
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torchmetrics/metric.py", line 312, in forward
    self._forward_cache = self._forward_reduce_state_update(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torchmetrics/metric.py", line 381, in _forward_reduce_state_update
    self.update(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torchmetrics/metric.py", line 483, in wrapped_func
    update(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torchmetrics/text/perplexity.py", line 82, in update
    total_log_probs, count = _perplexity_update(preds, target, self.ignore_index)
  File "/usr/local/lib/python3.10/dist-packages/torchmetrics/functional/text/perplexity.py", line 83, in _perplexity_update
    _check_shape_and_type_consistency(preds, target)
  File "/usr/local/lib/python3.10/dist-packages/torchmetrics/functional/text/perplexity.py", line 55, in _check_shape_and_type_consistency
    raise ValueError(
ValueError: Input tensors `preds` and `target` are expected to have equaling first two dimensions, [batch_size, seq_len], but got torch.Size([1024, 16]) and torch.Size([16, 1024]).
```

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
jstjohn: Maybe this would be a good place for a coment about `token_logits` being `[s,b,*]` while other things are `[b,s,*]`

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
```diff
@@ -258,6 +259,8 @@ def __init__(
         self._data_step = data_step
         self._forward_step = forward_step
         self.model_transform = model_transform
+        self.train_ppl = torchmetrics.text.Perplexity(ignore_index=-100)
+        self.valid_ppl = torchmetrics.text.Perplexity(ignore_index=-100)
```
sichu2023, line 263: Would be great to be configurable once the PR has matured.

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
```diff
@@ -305,11 +308,19 @@ def forward_step(self, batch) -> Tensor:
 
     def training_step(self, batch, batch_idx: Optional[int] = None) -> Tensor:
         """In mcore the loss-function is part of the forward-pass when labels are provided."""
-        return self.forward_step(batch)
+        outputs = self.forward_step(batch)
+        logits = outputs["token_logits"].transpose(0, 1)  #  [s, b] -> [b, s]
+        self.train_ppl(logits, batch["labels"])
+        self.log("train_ppl_step", self.train_ppl)
```
sichu2023, line 314: And similarly, we need a different mechanism for CP, which isn't implemented even in the current PerplexityLoggingCallback.
https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-llm/src/bionemo/llm/lightning.py#L429

sichu2023, line 314: For TP, things are easier since there is already a TP aware method to get the loss.
https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py#L262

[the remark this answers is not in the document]
sichu2023: I _believe_ NeMo returns `0` by default when it isn't at the last PP stage but don't rely on my vague memory here. I don't know how it translates to other outputs such as `token_logits`.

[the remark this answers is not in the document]
sichu2023: If you need a device group for each parallel group, check out `parallel_states.get_pipeline_model_parallel_group`. You might have multiple PP group if PP and TP happen at the same time. Each TP under a PP group, or the other way around.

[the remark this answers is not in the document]
sichu2023: I am not sure how VPP plays a role here.

sichu2023, line 314: We don't need to `sync_dist` here.

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
```diff
@@ -325,6 +336,14 @@ def validation_loss_reduction(self) -> MegatronLossType:  # noqa: D102
     def test_loss_reduction(self) -> MegatronLossType:  # noqa: D102
         return self.loss_reduction_class(validation_step=True)
 
+    def on_train_epoch_end(self):
+        """Log perplexity at the end of the epoch."""
+        self.log("train_ppl_step", self.train_ppl)
+
+    def on_valid_epoch_end(self):
+        """Log perplexity at the end of the epoch."""
+        self.log("valid_ppl_step", self.valid_ppl)
```
sichu2023, line 345: We only need to log once in step.

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
```diff
@@ -305,11 +308,19 @@ def forward_step(self, batch) -> Tensor:
...
+        self.log("train_ppl_step", self.train_ppl)
+        return outputs
 
     def validation_step(self, batch, batch_idx: Optional[int] = None) -> Tensor:
         """In mcore the loss-function is part of the forward-pass when labels are provided."""
-        return self.forward_step(batch)
+        outputs = self.forward_step(batch)
+        logits = outputs["token_logits"].transpose(0, 1)  #  [s, b] -> [b, s]
+        self.valid_ppl(logits, batch["labels"])
+        self.log("valid_ppl_step", self.valid_ppl)
```
sichu2023, line 322: But we will need `sync_dist=True` and `on_epoch=True` here.

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
sichu2023, line 263: https://lightning.ai/docs/torchmetrics/stable/pages/overview.html#metrics-in-distributed-data-parallel-ddp-mode

sichu2023: https://lightning.ai/docs/torchmetrics/stable/pages/overview.html#metric-kwargs

pstjohn: For sure, I think we sould make an `ESM2LightningModule` with a lot of these args pre-defined where we can then attach metrics in an obvious way

pstjohn: Any changes to the defaults we should make there? I used torchmetrics with the HF pretraining and it seemed to work well, so I don't think the oversampling is too big of an issue

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
pstjohn: why is that? It seems to give similar perplexity values to the callback:
https://wandb.ai/clara-discovery/bionemo-hf-pretraining/reports/val_ppl-valid_ppl_step-24-11-20-18-12-31---VmlldzoxMDI2NjQ0MA

pstjohn: > The sync_dist, sync_dist_group and reduce_fx flags from self.log(...) don’t affect the metric logging in any manner. The metric class contains its own distributed synchronization logic.
https://lightning.ai/docs/torchmetrics/stable/pages/lightning.html#logging-torchmetrics

So I think we _do_ want `on_epoch=True` here (I'm not sure why my example run was so close without it...), but I think sync_dist is handled internally?

sichu2023: You need `sync_dist_group` because you don't want to sync with 0/other weird value from non-last-pp-stage devices.

sichu2023: Maybe they are the same on single device because you had `self.log("valid_ppl_step", self.valid_ppl)` in `on_valid_epoch_end`?

sub-packages/bionemo-llm/src/bionemo/llm/lightning.py
sichu2023: `sync_on_compute` is on by default so we are good on that end. I wonder if we need to specify `process_group` to skip non-last-pp-stage devices.

sichu2023: I also don't have a clear idea on how to treat tp devices.

sichu2023:
Related effort: https://github.com/NVIDIA/bionemo-framework/pull/525

PR #386, opened 2024-10-30
Manually pin pypi versions

pstjohn:
If we’re sticking to the pytorch base images, we may need to do a lot of this dependency pinning manually — tools like pip-compile or uv sync  won’t play nicely with the mix of inherited + newly installed packages.

Creates a large requirements-docker.txt file that pins all of the packages we add during our container build. This would hopefully let us be a bit more explicit about our 3rdparty dependencies in our images.

I'll try to gather feedback on container design in this document: https://docs.google.com/document/d/1Q81dnKPvHsnKn9c5Qc5EEKXDPJVZUA29c19Aj0CumkQ/edit?usp=sharing

Still TODO would be some way of automatically generating this file from "scratch" (maybe from some `requirements.in` file)


Dockerfile
```diff
@@ -3,9 +3,7 @@ ARG BASE_IMAGE=nvcr.io/nvidia/pytorch:24.02-py3
 FROM ${BASE_IMAGE} AS bionemo2-base
 
 # Install NeMo dependencies.
-WORKDIR /build
-
-ARG MAX_JOBS=4
+ARG MAX_JOBS=-1
```
pstjohn, line 6: Do we need to cap the number of compilation workers? Is this to be a good citizen a CI system somewhere?

Dockerfile
```diff
@@ -14,7 +12,8 @@ RUN git clone https://github.com/NVIDIA/apex.git && \
   cd apex && \
   git checkout ${APEX_COMMIT} && \
   pip install . -v --no-build-isolation --disable-pip-version-check --no-cache-dir \
-  --config-settings "--build-option=--cpp_ext --cuda_ext --fast_layer_norm --distributed_adam --deprecated_fused_adam --group_norm"
+  --config-settings "--build-option=--cpp_ext --cuda_ext --fast_layer_norm --distributed_adam --deprecated_fused_adam --group_norm" && \
+  cd .. && rm -rf apex
```
pstjohn, line 16: this is a big enough change that it's probably worth finally cleaning this up -- we need to delete these files inside this layer or they'll always be in our layer history and contribute to our overall size

Dockerfile
```diff
@@ -81,29 +65,14 @@ ENV UV_LINK_MODE=copy \
...
-uv pip install --no-build-isolation \
-  ./3rdparty/* \
-  ./sub-packages/bionemo-* \
-  -r /requirements-cve.txt \
-  -r /requirements-test.txt
-rm -rf ./3rdparty
-rm -rf /tmp/*
+  set -eo pipefail
+  uv pip install --no-deps --no-build-isolation -r /requirements-docker.txt
+  uv pip install --no-deps nemo_run@git+https://github.com/NVIDIA/NeMo-Run.git@${NEMO_RUN_TAG}
```
pstjohn, line 75: nemo_run has an undeclared `hatchling` build-time dep, so we need to install it separately

requirements-docker.txt
pstjohn, line 1: created this by diffing `pip freeze` from the base image and our current bionemo2 image, and cleaning up a few lines (apex, TE, etc) that we still install separately.

ci/docker/update_docker_pypi_deps.sh
pstjohn, line 1: this file lets us check the sanity of the `requirements-docker.txt` file (if we run it, it shouldn't change), and also lets you more easily add new dependencies -- you can add lines from `requirements-docker.txt`, run this script, and it should add new dependencies of that package.

Dockerfile
dorotat-nv: It depends on available memory. If you pick too high MAX_JOBS the docker with our dependencies will not build without cache, ie at least it was happening in bionemo1 and TransformerEngine

dorotat-nv: What does it mean --build-arg MAX_JOBS=N invalidates the cache?

Dockerfile
```diff
@@ -193,6 +156,23 @@ USER $USERNAME
 # existing release image build by copying over remaining files from the repo into the container.
 FROM bionemo2-base AS release
 
+WORKDIR /workspace/bionemo2
+
+# Install 3rd-party deps and bionemo submodules.
+COPY ./3rdparty /workspace/bionemo2/3rdparty
```
dorotat-nv, line 162: can we set workspace/bionemo2 as some variable for reference in the docker? it used to be BIONEMO_HOME. Or is it automatically set as HOME?

Dockerfile
pstjohn: I suspect (at least this was happening locally) that if you pass in a MAX_JOBS argument via the command line, you'll only cache layers that are downstream of that particular configuration. So if I build locally with 
```docker buildx build . -t bionemo --build-arg MAX_JOBS=-1```
That won't cache those built layers if I later build and omit that MAX_JOBS argument.

Dockerfile
pstjohn: So right, actually this is a good point, we should clarify this in the design doc. I would prefer to not copy in `3rdparty` and `subpackages` into the container (likewise with the `docs/` and `scripts/` we copy later).

We could just create a `bionemo` with some `$HOME` directory that users would use? But I think that requires us to 
(1) align on whether it's OK to omit tests and scripts from the image itself, with the expectation that users would mount files into the image they want to run
(2) change CI so that tests are mounted into the container

spoulos-nvidia:
closing as out of date

PR #396, opened 2024-11-04
Refactored `load` & `download_bionemo_data` into bionemo-core

malcolmgreaves:
`bionemo-testing` re-exports the same values defined in `__all__`: they are imported from `bionemo-core` as the implementations have moved.

Now, all sub-packages can use `load` at runtime, not just during tests.

Additionally moves over the YAML resource files from bionemo-testing into bionemo-core & adjusted the `get_all_resources` function.


jstjohn (approved):

sub-packages/bionemo-testing/src/bionemo/testing/data/load.py
pstjohn, line 1: IMO we don't need to maintain import backwards-compatibility. If we want to move this (which I'm OK with), I think we should just update all the existing import paths to point to `bionemo.core` instead of `testing`

jstjohn: Yeah I agree. I don't think this is in the checkpoint paths etc so it should be ok. Also if an outside user depends on testing when we told them not to (the testing package), that can be their problem for now IMO. They can do the refactor that we do.

farhadrgh (approved):
LGTM

PR #404, opened 2024-11-06
[QA BUG] apply batch_collator to predict results

farhadrgh:
This is to address QA Bug https://nvbugspro.nvidia.com/bug/4946953
By changing the output of predict method from a `list` to `dict` by using `batch_collator`


polinabinder1 (approved):

pstjohn (approved):
how did CI pass for this 😅 ??

PR #406, opened 2024-11-06
Move ESM2 scripts to sub-packages

farhadrgh:
Move and refactor ESM2 scripts into a package


skothenhill-nv (approved):
lgtm

jstjohn (approved):

sub-packages/bionemo-esm2/pyproject.toml
```diff
@@ -17,13 +17,15 @@ dependencies = [
...
 
-# Make sure that the tokenizer files are included along with the python files during installation.
-[tool.setuptools.package-data]
-"bionemo.esm2" = ["data/tokenizer/*.json", "data/tokenizer/*.txt"]
-
 [project.scripts]
 bionemo-esm2-train= "bionemo.esm2.run.main:main"
 bionemo-esm2-recipe= "bionemo.esm2.run.recipes:main"
+infer_esm2 = "bionemo.esm2.scripts.infer_esm2:infer_esm2_entrypoint"
+train_esm2 = "bionemo.esm2.scripts.train_esm2:train_esm2_entrypoint"
```
malcolmgreaves, line 24: @skothenhill-nv @farhadrgh what is different from `train_esm2` and `bionemo-esm2-train`? If these are going to be merged in the future, then perhaps we don't make a new command for it? If that function is run by `__main__`, then `python -m bionemo.esm2.scripts.train_esm2` will do the same thing as `train_esm2`.

farhadrgh: hm, I was following the same pattern as Geneformer here:
https://github.com/NVIDIA/bionemo-framework/blob/3c571dd42199cb4ac18130c4adae6e9c13e83bfe/sub-packages/bionemo-geneformer/pyproject.toml#L22-L27

I think the main difference is that `bionemo-esm2-train` requires CLI/recipe and JSON config but train_esm2 is a simple argparser

jstjohn: @malcolmgreaves `train_esm2` is the legacy `argparse` based system, `bionemo-esm2-train` is the version we're moving toward that is config based. Right now both exist while we work toward solidifying the new CLI.

jstjohn: Both methods went through QA testing.

PR #409, opened 2024-11-07
adds nbval test for ngc download

pstjohn:
Uses `nest_asyncio` to allow this to run in a jupyter notebook, and calls `client.configure()` to ensure that the output type is set to `json`.


jstjohn (approved):

jstjohn (approved):

jstjohn:
As a side-note, why didn't CI fail with this bug in the notebook cell? seems like it just powers through even if the cell fails? Wondering what the utility of the notebook test is if it ignores exceptions in the cells?

malcolmgreaves (approved):

malcolmgreaves:
> As a side-note, why didn't CI fail with this bug in the notebook cell? seems like it just powers through even if the cell fails? Wondering what the utility of the notebook test is if it ignores exceptions in the cells?

Because I made sure my changes were backwards compatible! 🤗 To avoid breaking things, I made `bionemo.testing.load` re-export the public things defined in `bionemo.core.load` [1]. This way, a `from bionemo.testing.load import load` still works as expected and we have what we want: the ability to use `load` w/o _needing_ to use `bionemo-testing`.

[1] https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-testing/src/bionemo/testing/data/load.py#L17-L31

PR #408, opened 2024-11-07
Training cluster performance boost for bert models part 2

jstjohn:
See slack thread https://nvidia.slack.com/archives/C074Z808N05/p1730301123648729

The main issue I was hitting was the setting `SBATCH --overcommit` in my personal slurm script. As a side effect of this exploration I did a few things though which this PR has:
* Identified that a bug in some combination of NeMo2 and Megatron-lm results in some fused kernels getting regularly re-compiled
* Verified that even with this re-compilation bug, we still have the best performance when using these kernels
* Added an option for geneformer to turn on the torch debugger that errors out if a kernel is getting recompiled, this is how I found which kernels were at fault and where/why the recompilations were happening.

## Performance summary of different settings
| Name                          | Replicate | Num GPUs | Time per 10 steps | Average Timing |
|-------------------------------|-----------|----------|-------------------|----------------|
| no_recompile                  |         0 |        1 |             2.487 |        2.50875 |
| no_recompile                  |         0 |        2 |             2.527 |                |
| no_recompile                  |         1 |        1 |             2.503 |                |
| no_recompile                  |         1 |        2 |             2.518 |                |
| fused_bias_act                |         0 |        1 |             2.489 |    2.496666667 |
| fused_bias_act                |         0 |        2 |             2.514 |                |
| fused_bias_act                |         1 |        1 |             2.487 |                |
| fused_bias_act                |         1 |        2 |                   |                |
| fused_bias_act_do             |         0 |        1 |             2.459 |        2.47425 |
| fused_bias_act_do             |         0 |        2 |             2.478 |                |
| fused_bias_act_do             |         1 |        1 |             2.471 |                |
| fused_bias_act_do             |         1 |        2 |             2.489 |                |
| fused_loss                    |         0 |        1 |             2.312 |          2.326 |
| fused_loss                    |         0 |        2 |             2.335 |                |
| fused_loss                    |         1 |        1 |             2.323 |                |
| fused_loss                    |         1 |        2 |             2.334 |                |
| fused_bias_do                 |         0 |        1 |             2.467 |         2.4845 |
| fused_bias_do                 |         0 |        2 |             2.499 |                |
| fused_bias_do                 |         1 |        1 |             2.472 |                |
| fused_bias_do                 |         1 |        2 |               2.5 |                |
| fused_bias_loss               |         0 |        1 |             2.282 |        2.28775 |
| fused_bias_loss               |         0 |        2 |             2.297 |                |
[...]

PR #421, opened 2024-11-08
Fix geneformer training instability bug

jstjohn:
See wandb runs here: https://wandb.ai/clara-discovery/geneformer_bionemo2_timing2 

See the results below, we can precisely control whether or not there is a grad norm instability by setting or unsetting the two NVTE env variables. Adding the NVTE env variables to our container is a recent change as well. Based on these results we are unsetting these variables for now. There is not a significant hit to performance by making this change.

## Old run where this was not an issue:
Screenshot 2024-11-12 at 9 42 45 AM

## Representative new run where we see a spike in grad norm
Screenshot 2024-11-12 at 9 43 25 AM

## We can make this spike go away by unsetting `NVTE_FUSED_ATTN` and `NVTE_FLASH_ATTN`
Screenshot 2024-11-12 at 9 43 44 AM

## We can introduce this spike on the old image that didn't have these env variables by setting them
Screenshot 2024-11-12 at 9 44 16 AM

## Example longer/larger batch run that fails with these env variables set
Screenshot 2024-11-12 at 9 45 07 AM

## We can stabilize this run by unsetting these env variables
Screenshot 2024-11-12 at 9 45 30 AM

It seems to be relatively recent so this PR is going to test some recent changes to see if any of them is causing this.

- [x] Check if the arange change is causing this?
- [x] Check if the grad buffer change (should not be enabled) is causing this 
- [x] bias fusions
- [x] garbage collection callback

Find out when this worked:
- [x] PR 409 right before second perf change and dset change
- [x] PR 410 after first perf change, CLI refactor, and wandb fix
- [x] PR 404 right before new CLI
- [x] PR 362 (2 weeks ago) but restarting job before the gradients start to increase
- [x] PR 362 (2 weeks ago)
- [x] **worked** https://wandb.ai/clara-discovery/geneformer_bionemo2/runs/0sSIf3tl?nw=nwusernvjstjohn **worked** uses `bionemo2-pr312--136b1889fc390d9dad04f077b32b8fbecf50e25d`
- [x] bionemo2-pr312--136b1889fc390d9dad04f077b32b8fbecf50e25d but with `NVTE_FUSED_ATTN=1` and `NVTE_FLASH_ATTN=0` set in my script **did not work **
- [x] bionemo2-pr312--136b1889fc390d9dad04f077b32b8fbecf50e25d but with `NVTE_FUSED_ATTN=1` and `NVTE_FLASH_ATTN=0` `unset` in my script **WORKED!!**
- [x] bionemo2-pr419--f2599382e4afaf061c9948628f3f72bb8e233fd6 (most recent PR merged) but manually unsetting `NVTE_FUSED_ATTN=1` and `NVTE_FLASH_ATTN=0` 

Notes on differences between TOT and `pr312--136b1889fc390d9dad04f077b32b8fbecf50e25d`

- `env` doesn't have `NVTE_FUSED*` env settings. Unclear if slurm script adds them properly or not.
    - `NVTE_FUSED_ATTN` and `NVTE_FLASH_ATTN` are set in `bionemo2-pr373--db2fe9cc240b12bfaf045654fc5350a7b985c9de` for example.
    - in slurm `--export=ALL` is default and passes all env variables. Perhaps this happens then, so the run where I have those env variables added might fail if those are causing the issue.
- Successful run was bs=32 vs 64. I'm running a test now that has the NVTE* settings in the docker script but not in the image.
- This was a closed branch, maybe some key changes didn't make it to main. 
- No `pip freeze` differences pop out that distinguish the branch that passes from the set that fail.
- NOTE: See the experiments above around `NVTE_FUSED_ATTN=1` and `NVTE_FLASH_ATTN=0` . I am pretty sure these settings are what cause the training instability in geneformer. Unsetting them works in the old PR and setting them causes that old PR to not work with this explosion of gradients.
- Currently I'm rerunning tests on a TOT branch but calling `unset` in my script on those variables so that they are removed from the container env prior to executing the script. If this fixes the TOT training curve I will feel very confident that this is what's going on, and we can focus on purging references to these variables from our docs, other than maybe highlighting how they result in training instability.


jstjohn:
Partial revert of #408

pstjohn (approved):

malcolmgreaves (approved):

polinabinder1 (approved):

PR #423, opened 2024-11-11
add fname parameter for pooch downloads

pstjohn:
Pooch automatically calculates a filename for downloads that is a function of the hash and the URL, but we want to ensure that downloads from NGC and PBSS have identical local filenames so that their cache is shared, and the URLs for both resources don't typically match.

Here we set the fname parameter manually so that we don't download duplicate objects from ngc and pbss

### DIFF
--- a/sub-packages/bionemo-core/src/bionemo/core/data/load.py
+++ b/sub-packages/bionemo-core/src/bionemo/core/data/load.py
@@ -165,6 +165,7 @@ def load(
 
     download = pooch.retrieve(
         url=str(url),
+        fname=f"{resource.sha256}-{filename}",
         known_hash=resource.sha256,
         path=cache_dir,
         downloader=download_fn,


sub-packages/bionemo-core/src/bionemo/core/data/load.py
```diff
@@ -165,6 +165,7 @@ def load(
 
     download = pooch.retrieve(
         url=str(url),
+        fname=f"{resource.sha256}-{filename}",
```
malcolmgreaves, line 168: And `filename` is unique per resource, regardless if its on PBSS or NGC -- IIRC it's the key for each resource, right?

malcolmgreaves (approved):

sub-packages/bionemo-core/src/bionemo/core/data/load.py
pstjohn: Filename doesn't have to be unique, but I think the sha256 is enough of a guarantee of uniqueness? The filename is really just there to give some additional context. The "filename" here comes from the pbss link, and has to be the same for both pbss and ngc sources

pstjohn: Ah, maybe misunderstood your comment -- the filename has consistent between ngc and pbss, yes

skothenhill-nv (approved):

PR #429, opened 2024-11-13
WAR: sets checkpoint filename to be more unique

skothenhill-nv:
Workaround for issue with async checkpointing where duplicate checkpoints occur with the same metric loss (causing async save failures).

Adds consumed_samples and steps to create a theoretically unique filename.


sub-packages/bionemo-geneformer/src/bionemo/geneformer/scripts/train_geneformer.py
```diff
@@ -192,8 +192,8 @@ def main(
         gradient_as_bucket_view=True,
         # FIXME there are intermittent errors with async checkpoint saving.
         #  see https://wandb.ai/clara-discovery/geneformer_bionemo2_goodslurm/runs/uAFi7DzI/logs
```
jstjohn, line 194: Remove this comment as well.

jstjohn: Maybe add a warning about how async save will fail if checkpoint names are not unique enough.

scripts/protein/esm2/esm2_pretrain.py
```diff
@@ -247,6 +249,7 @@ def main(
         save_top_k=save_top_k,
         every_n_train_steps=val_check_interval,
         always_save_context=True,  # Enables the .nemo file-like checkpointing where all IOMixins are under SerDe
+        filename='{epoch}-{val_loss:.2f}-{step}-{consumed_samples}'
```
jstjohn, line 252: Add a comment here saying that `step` and `consumed_samples` are included to make sure names are unique.

sub-packages/bionemo-llm/src/bionemo/llm/train.py
```diff
@@ -71,6 +71,7 @@ def nemo_logger_factory(experiment_config: ExperimentConfig, wandb_config: Optio
         save_top_k=experiment_config.save_top_k,
         every_n_train_steps=experiment_config.save_every_n_steps,
         always_save_context=True,
+        filename='{epoch}-{val_loss:.2f}-{step}-{consumed_samples}'
```
jstjohn, line 74: Same comment here about step/consumed samples being included to ensure unique names. Just make sure people think about this when they see this line incase they are considering changing it.

jstjohn (approved):
Approved with some requests for comments to avoid regression or user error.

sub-packages/bionemo-llm/src/bionemo/llm/train.py
skothenhill-nv: done

farhadrgh (approved):

PR #431, opened 2024-11-14
re-enable merge groups to trigger blossom-ci

pstjohn:
I'm hoping this doesn't brick our existing blossom workflow, but we'll need to quickly revert if it does. If we can get this merged to `main`, I'm hoping we can do the rest of our testing with merge groups with PRs to a seperate `main-merge-group` branch


trvachov (approved):
🤞

PR #449, opened 2024-11-18
BugFix: fix bugs on bionemo-size-aware-batching

guoqing-zhou:
## Summary
This PR fixes two bugs on bionemo-size-aware-batching, one from unit test, and one on the bucket batch sampler init method.

## Details
Bug 1: To address one unit test faling on GH200: torch.multinomial gives inconsistent results across different cpus, fixed with removing numerical checks and add logic checks.
Bug 2: in BucketBatchSampler, two arguments are default to None, while in the init method, they are expected as dictionaries, adding checking if they are None and if so set to empty dict.

## Usage
refer to `sub-packages/bionemo-size-aware-batching/README.md` for more details

## Testing
```
pytest -v sub-packages/bionemo-size-aware-batching
```

(**NOTE:** _also **delete** this checklist as you fill-out this PR template_)

**Most of the changes** to files with extensions `*.py`, `*.yaml`, `*.yml`, `Dockerfile*` or `requirements.txt` **DO REQUIRE both `pytest-` and `jet-` CI stages**.

- [x] Did you review the [Before your PR is "Ready for review" section](https://github.com/NVIDIA/bionemo-framework/-/blob/dev/CONTRIBUTING.md?ref_type=heads#before-pr-ready) before asking for review?
- [x] Did you make sure your changes have tests? Did you test your changes locally?
- [ ] Can you add [the `SKIP_CI` label](https://github.com/NVIDIA/bionemo-framework/-/blob/dev/CONTRIBUTING.md?ref_type=heads#skip-ci) to your PR?
- [ ] Can you add [the `PYTEST_NOT_REQUIRED` label](https://github.com/NVIDIA/bionemo-framework/-/blob/dev/CONTRIBUTING.md?ref_type=heads#skip-pytest) to your PR?
- [ ] Can you add [the `JET_NOT_REQUIRED` label](https://github.com/NVIDIA/bionemo-framework/-/blob/dev/CONTRIBUTING.md?ref_type=heads#skip-jet) to your PR?


sub-packages/bionemo-size-aware-batching/tests/bionemo/size_aware_batching/test_sampler.py
```diff
@@ -486,6 +486,8 @@ def test_iter_bucket_batch_sampler(sample_data):
 
 
 def test_iter_bucket_batch_sampler_with_shuffle(sample_data):
+    torch.set_default_device("cuda")
+
```
jstjohn, line 490: this is going to impact other tests, right?

jstjohn: My preference would be to make a context manager that has the flow:
1. get current default device (maybe "cpu" if unset)
2. set to user requested new default device
3. on exit restore default device regardless of error state etc in finally block

```python
from contextlib import contextmanager
@contextmanager
def default_device_ctx(new_device):
    try:
       original_device = torch.... # whatever this is
       torch.set_default_device(new_device)
       yield
    finally:
        torch.set_defulat_device(original_device)
```

jstjohn: Also why do you need this? maybe there are torch.tensor... inside of `BucketBatchSampler` that are not placed on the right device in the context that sizes and/or bucket_boundaries are tensors on some device? Seems like a hack to get stuff working potentially when there are other problems underneath.

sub-packages/bionemo-size-aware-batching/tests/bionemo/size_aware_batching/test_sampler.py
```diff
@@ -494,37 +496,37 @@ def test_iter_bucket_batch_sampler_with_shuffle(sample_data):
...
-        [1, 4],
+        [23, 24, 16, 21, 15],
+        [17, 20, 18, 22, 19],
+        [9, 13, 7],
+        [11, 6, 10],
+        [4, 5],
+        [2, 3],
+        [12, 8, 14],
+        [0, 1],
     ]
```
jstjohn, line 517: why is the data changing?

DejunL:
As an alternative to the proposed fix, could we simply tolerate the results if they are off by a permutation of the reference? e.g., the original failure case gives `batch_lists_first_iter`:
```
[[2, 5], [24, 17, 16, 22, 19], [3, 0], [15, 18, 20, 21, 23], [12, 10, 11], [7, 13, 6], [1, 4], [14, 9, 8]]
```
which can be sorted to give the `ref_batch_lists_first_iter`

sub-packages/bionemo-size-aware-batching/tests/bionemo/size_aware_batching/test_sampler.py
guoqing-zhou: As you suggested in the slack, we remove this part to avoid any device control, and refactor the check to not do numerical comparison, but check some basic logics.

sub-packages/bionemo-size-aware-batching/tests/bionemo/size_aware_batching/test_sampler.py
guoqing-zhou: previously, changing torch random call from cpu to gpu will lead to different random state, thus the reference data change.

We updated to avoid numerical check, this part is removed

jstjohn (approved):

malcolmgreaves (approved):

PR #451, opened 2024-11-19
Launchable notebooks in docs!

jstjohn:
Changes:
* Change default from PBSS to NGC for the `load` function.
* Expose a new env variable to change the default for the load function so we can have one in tests and an easy way for an external user to change the behavior in ci locally.
* Add new launchable link for the geneformer notebook.

Screenshot 2024-11-19 at 8 43 04 AM


malcolmgreaves (approved):

ci/scripts/run_pytest.sh
```diff
@@ -17,7 +17,8 @@
 
 set -xueo pipefail
 export PYTHONDONTWRITEBYTECODE=1
-
+# NOTE: if a non-nvidia user wants to run the test suite, just run `export BIONEMO_DATA_SOURCE=ngc` prior to this call.
+export BIONEMO_DATA_SOURCE="${BIONEMO_DATA_SOURCE:-pbss}"
 source "$(dirname "$0")/utils.sh"
```
skothenhill-nv, line 22: nice, this was my only question.

pstjohn (approved):
LGTM, thanks!

PR #453, opened 2024-11-19
Atom and bond features

sveccham:
---

Implements classes for featurizing atoms and bonds.

  - Implemented atom and bond featurization classes
  - Implemented classes can be used for featurizing atoms and bonds in a molecular graph in order to form inputs for graph neural networks
  - Usage example 1: `from bionemo.geometric.atom_featurizers import AromaticityFeaturizer, AtomicNumFeaturizer`
  - Usage example 2: `from bionemo.geometric.bond_featurizers import RingFeaturizer` 
---

## Usage
_How does a user interact with the changed code?_
```python
from bionemo.geometric.atom_featurizers import DegreeFeaturizer
df = DegreeFeaturizer()
mol = Chem.MolFromSmiles("CCCC")
feats = df.get_features(mol.GetAtomWithIdx(0))
```

## Testing

```shell
pytest -v tests/bionemo/geometric/test_atom_featurizers.py
pytest -v tests/bionemo/geometric/test_bond_featurizers.py
```


sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,400 @@
...
+        return 25
+
+    def get_period(self, atom: Chem.Atom) -> List[bool]:
+        """Returns one-hot encoded period of atom."""
+        atomic_number = atom.GetAtomicNum()
+        # The number of elements per period in the periodic table
+        period_limits = [2, 10, 18, 36, 54, 86, 118]
+
+        # Determine the period based on atomic number
+        for period, limit in enumerate(period_limits, start=1):
```
scal444, line 195: Would this return a -1 encoding for hydrogen or helium?

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,400 @@
...
+        """Initializes PeriodicTableFeaturizer class.
+
+        Args:
+            data_file: Path to the data file.
+        """
+        if data_file is None:
+            # Use default
+            root_path = Path(__file__).resolve().parent
+            data_file = root_path / "data" / "electronic_data.csv"
+            print(f"{data_file}")
```
scal444, line 240: leftover print

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
```diff
@@ -0,0 +1,105 @@
...
+# limitations under the License.
+
+
+from abc import ABC, abstractmethod, abstractproperty
+from typing import List
+
+from rdkit.Chem import Atom, Mol
+
+
+class BaseFeaturizer(ABC):
```
pstjohn, line 23: I'd vote for a `Protocol` definition rather than requiring a strict inheritance pattern here:

https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols

Probably worth having separate protocols for atom, bond, and molecule features

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,400 @@
...
+MAX_HYBRIDIZATION_TYPES = len(HybridizationType.values)
+MAX_NUM_HS = 5  # 4 + 1 (no hydrogens)
+
+
+class AtomicNumberFeaturizer(BaseFeaturizer):
+    """Class for featurizing atom by its atomic number."""
+
+    def __init__(self) -> None:
+        """Initializes AtomicNumberFeaturizer class."""
+        pass
```
pstjohn, line 62: not needed, just omit the whole function

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,400 @@
...
+
+    @property
+    def n_dim(self) -> int:
+        """Returns dimensionality of the computed features."""
+        return 6
+
+    def get_features(self, atom: Atom) -> List[bool]:
+        """Returns features of the atom."""
+        return one_hot_enc(atom.GetDegree(), self.n_dim)
```
pstjohn, line 88: Here, and elsewhere, might make sense to separate the one-hot transformation from the feature itself.

Something like
```python
OneHotDegreeFeaturizer = OneHot(DegreeFeaturizer)
```
that you could re-use in a bunch of places

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
```diff
@@ -0,0 +1,105 @@
...
+    """Abstract base featurizer class for all atom and bond featurization classes."""
+
+    @abstractproperty
+    def n_dim(self):
+        """Number of dimensions of compute feature."""
+        pass
+
+    def compute_features(self, mol: Mol) -> Mol:
```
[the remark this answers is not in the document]
sveccham: I will delete the modified reference to the same `mol` object. 

RE properties: 
I did not want the `Featurizer` classes to store any molecule-specific state. I want to instantiate one featurizer for all molecules. Storing molecule-specific information in the class would disallow me from using this pattern.
I like this pattern because some `Featurizers` require some one-time initialization (like [reading from property file](https://github.com/NVIDIA/bionemo-framework/blob/sveccham/impl_atom_bond_feats/sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py#L224) or [initializing SMARTS query molecules](https://github.com/NVIDIA/bionemo-framework/blob/sveccham/impl_atom_bond_feats/sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py#L314)) and I would like to do this once and not once per molecule.
Anyway this is possible?

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
sveccham: No. `period` would be set to 1. Hydrogen and helium would be encoded as `[True, False, False, False, False, False, False]`

scal444: Right, missed the start keyword, thanks

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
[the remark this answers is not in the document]
scal444: OK I see where you're coming from. Do you anticipate the bond and atom featurizers to iterate over all atoms and bonds? Or at least do that by default? In which case you could do something like.

```python
class MyFeaturizer(BaseFeaturizer): # where base has this abstract method
    def get_atom_features(self, mol, atomIndices: Optional[Iterable] = None):
        atomiterator = atomIndices if atomIndices else range(mol.GetNumAtoms())
        
        # Do any molecular preprocessing
        # ....

        res = np.zeros(self.n_dim,  len(atomIterator))
        for outIndex, atomIndex in enumerate(atomIterator):
            res[outIndex, :] = get_features(mol.GetAtomWithIdx(atomIndex))
        return res

```

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,398 @@
...
+MAX_ATOMIC_NUM = 100
+MAX_CHIRAL_TYPES = len(ChiralType.values)
+MAX_HYBRIDIZATION_TYPES = len(HybridizationType.values)
+MAX_NUM_HS = 5  # 4 + 1 (no hydrogens)
+
+
+class AtomicNumberFeaturizer(BaseFeaturizer):
+    """Class for featurizing atom by its atomic number."""
+
+    def __init__(self) -> None:
```
guoqing-zhou, line 59: would suggest to pass in the `MAX_ATOMIC_NUM` to init this function, or pass in a list of atomic numbers/types, which can generalize to cover cases like unknown atom types, masks. Also don't think we will use the full periodic table.

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,398 @@
...
+            if atomic_number <= limit:
+                return one_hot_enc(period - 1, 7)
+        return None
+
+    def get_group(self, atom: Chem.Atom) -> list[bool]:
+        """Returns one-hot encoded group of atom."""
+        group = self.pt.GetNOuterElecs(atom.GetAtomicNum())
+        return one_hot_enc(group - 1, 15)
+
+    def get_atomic_radii(self, atom: Chem.Atom) -> list[float]:
```
guoqing-zhou, line 204: better to add an option for normalization, either in the method call, or in the init

guoqing-zhou: same for other featurizers

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,398 @@
...
+
+    @property
+    def n_dim(self) -> int:
+        """Returns dimensionality of the computed features."""
+        return 2
+
+    def compute_features(self, mol):
+        """Compute atomic contributions to Crippen logP and molar refractivity."""
+        logp_mr_list = np.array(rdMolDescriptors._CalcCrippenContribs(mol))
+        logp_mr_list[:, 0] = np.clip(logp_mr_list[:, 0], a_min=MIN_LOGP, a_max=MAX_LOGP)
```
guoqing-zhou, line 379: This MIN_LOGP and MAX_LOGP are only used for this one right? if so, maybe avoid define them globally

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,398 @@
...
+    def __init__(self):
+        """Initializes ScaffoldFeaturizer class."""
+        pass
+
+    @property
+    def n_dim(self) -> int:
+        """Returns dimensionality of the computed features."""
+        return 1
+
+    def compute_features(self, mol):
```
guoqing-zhou, line 288: Seems like the user has to manually call `compute_features` before `get_features`. Not sure if there is any perf concern on set property on rdkit mol object. 
And the `compute_features` is for all atoms/bonds, while `get_features` is for a specific atom/bond, better to unify them, like 
```
def get_features(self, mol, atom_idx_list = None):
    prop = self.compute_features(mol)
    if atom_idx_list is None:
        # return features for all atoms
    else:
        # return features for selected atoms
    
```
 similar for bond type features. But I don't know if there are other cases not covered here. No need to refactor if you don't think this is reasonable.

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
sveccham: All features have been normalized to lie between 0 and 1. For example:https://github.com/NVIDIA/bionemo-framework/blob/sveccham/impl_atom_bond_feats/sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py#L204
Do you want me to expose a function to the user so that they can choose between normalized or unnormalized values?

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
sveccham: Makes sense. I will move them into the class.

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
sveccham: I can make them `Protocol`. How does this look?

`
class BaseAtomFeaturizer(Protocol):
    """Abstract base featurizer class for all atom featurizer classes."""

    @property
    def n_dim(self) -> int:
        """Number of dimensions of computed feature."""
        ...

    def get_atom_features(self):
        """Function for getting features."""
        ...

    def __call__(self):
        """Call featurizer on atom"""
        ...
class AtomicNumberFeaturizer:
    """Class for featurizing atom by its atomic number."""

    def __init__(self, max_atomic_num: Optional[int] = None) -> None:
        """Initializes AtomicNumberFeaturizer class."""
        self.max_atomic_num = max_atomic_num if max_atomic_num else MAX_ATOMIC_NUM

    def n_dim(self) -> int:
        """Returns dimensionality of the computed features."""
        return self.max_atomic_num

    def get_atom_features(self, mol: Mol, atom_indices: Optional[Iterable] = None) -> List[int]:
        """Computes features of atoms of all of select atoms."""
        _atom_indices = atom_indices if atom_indices else range(mol.GetNumAtoms())
        return [mol.GetAtomWithIdx(a).GetAtomicNum() for a in _atom_indices]

    
    def __call__(self, atom, atom_indices: Optional[Iterable] = None) -> List[int]:
        return self.get_atom_features(atom)`

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        val (int): An integer representing the value to be one-hot encoded.
+            Must be in the range [0, num_class - 1].
+        num_class (int): An integer representing the total number of classes or
+            possible classes.
+
+    Returns:
+        One-hot encoding of `val`.
+    """
+    one_hot = [False] * num_class
```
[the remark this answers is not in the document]
sveccham: Agreed.
I am deciding between directly returning `torch.Tensor` and `np.array`. Is there one you would recommend? Ultimately, everything would become a `torch.Tensor` but there might be some processing steps that the user might want to do (like normalization/scaling etc) which might be simpler with `np.array`.

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,400 @@
...
+
+ALL_ATOM_FEATURIZERS = [
+    "PeriodicTableFeaturizer",
+    "ElectronicPropertyFeaturizer",
+    "ScaffoldFeaturizer",
+    "SmartsFeaturizer",
+    "CrippenFeaturizer",
+]
+
+MAX_ATOMIC_NUM = 100
```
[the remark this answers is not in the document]
sveccham: No. I have changed it to 119 which is the maximum supported by RDkit

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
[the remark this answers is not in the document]
scal444: I vote for torch, easy for a user to convert themselves. So long as torch is a hard dependency, it should be used IMO.

sub-packages/bionemo-geometric/src/bionemo/geometric/base_featurizer.py
```diff
@@ -0,0 +1,120 @@
...
```
scal444, line 75: This could be torch too, right?

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,512 @@
...
+    @property
+    def n_dim(self) -> int:
+        """Returns dimensionality of the computed features."""
+        return 25
+
+    def get_period(self, atom: Chem.Atom) -> list[int]:
+        """Returns periodic table period of atom."""
+        atomic_number = atom.GetAtomicNum()
+        # The number of elements per period in the periodic table
+        period_limits = [2, 10, 18, 36, 54, 86, 118]
```
scal444, line 227: This should be in the init

scal444 (approved):

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,512 @@
...
+        Returns:
+            A torch.Tensor featurizing atoms by its atomic contribution to logP and molar refractivity.
+        """
+        logp_mr_list = torch.Tensor(rdMolDescriptors._CalcCrippenContribs(mol))
+        logp_mr_list = torch.clamp(logp_mr_list, min=self.min_val, max=self.max_val)
+        _atom_indices = atom_indices if atom_indices else range(mol.GetNumAtoms())
+        return logp_mr_list[_atom_indices, :]
+
+
+# # TODO Implement more features
```
guoqing-zhou, line 510: better to clean these comments

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
```diff
@@ -0,0 +1,512 @@
...
+    def __init__(self, max_atomic_num: Optional[int] = None) -> None:
+        """Initializes AtomicNumberFeaturizer class."""
+        MAX_ATOMIC_NUM = 118
+        self.max_atomic_num = max_atomic_num if max_atomic_num else MAX_ATOMIC_NUM
+
+    def n_dim(self) -> int:
+        """Returns dimensionality of the computed features."""
+        return self.max_atomic_num
+
+    def get_atom_features(self, mol: Mol, atom_indices: Optional[Iterable] = None) -> list[int]:
```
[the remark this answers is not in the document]
sveccham: I want to store all attributes related to atomic number encoding in the class itself. The one-hot featurizer would take information from `n_dim` to do one-hot encoding.
Similarly, for other classes which need scaling of features, it would also provide `min_val` and `max_val`. Does that make sense?

pstjohn (approved):

sub-packages/bionemo-geometric/src/bionemo/geometric/atom_featurizers.py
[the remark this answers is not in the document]
guoqing-zhou: If it is used for one hot encoding, why not one-hot encoding be part of this class? And same for min_val and max_val for normalization, the related methods should be part of this type of class.

guoqing-zhou (approved):

DejunL (approved):
LGTM. Just a general question, I'm wondering if there is an easy way to support batch featurization where potentially different molecules can be included in the batch

PR #457, opened 2024-11-20
bump version for pillow to avoid the CVE

polinabinder1:


pstjohn (commented):
Could you use requirements-cve.txt?

PR #461, opened 2024-11-20
set num_workers to 1 for esm2 tests

pstjohn:
Should help to address errors like https://prod.blsm.nvidia.com/bionemo-external-bionemo-fw/job/test_pytest/838/console 
```14:14:12  E       RuntimeError: DataLoader worker (pid 8445) is killed by signal: Aborted.```
when we use too many resources on the CI machine

### DIFF
--- a/sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune.py
+++ b/sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune.py
@@ -54,6 +54,7 @@ def pretrain_data_module(dummy_protein_dataset, dummy_parquet_train_val_inputs):
         micro_batch_size=4,
         min_seq_length=None,
         max_seq_length=1024,
+        num_workers=1,
     )
     yield data_module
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune.py [hunk 1]
```diff
@@ -54,6 +54,7 @@ def pretrain_data_module(dummy_protein_dataset, dummy_parquet_train_val_inputs):
         micro_batch_size=4,
         min_seq_length=None,
         max_seq_length=1024,
+        num_workers=1,
```
jstjohn, line 57: Why not 0? If you get a pin error you could add a `pin_memory=num_workers>0` in your datamodule kind of thing.

pstjohn: It interleaves data loading and computation, and might catch multiprocessing bugs? But I'm not picky either way

jstjohn: the benefit to 0 in testing would be that you can run pdb and land in a data thread if that's where an error happens. I guess you're right that it has that drawback on not testing a standard case as well though.

jstjohn: whatever you want to do is fine here.

jstjohn (approved):

sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune.py
pstjohn: yeah let's keep it at 1 for now -- if we do get a reproducible error here, we can always set it to zero for debugging.

polinabinder1 (approved):

PR #467, opened 2024-11-21
expose the ESM2 data random seed via CLI

pstjohn:
Exposes the ESM2DataModule's seed via the `train_esm2` CLI.


dorotat-nv (changes requested):
what about seeding plt, numpy or pytorch? Used to be achieved with  `pl.seed_everything(seed)`. Is it somewhere under the hood in our codebase?

PR #480, opened 2024-11-26
Fully Integrate SCDL into Geneformer

savitha-eng:
## Summary
In this PR we refactor the Geneformer `SingleCellDataset` class  to integrate the `SingleCellMemmapDataset`(SCDL). The goal of this is to streamline and increase readability of the dataset class. 
## Details
We make the following changes:
- Input Format: 
     - The `SingleCellDataset` now assumes that the input path to the data is a directory formatted in the `SingleCellMemmap` format. 
     -  The SingleCellModule now assumes that the train, val, and test input paths are to directories that are formatted in the `SingleCellMemmap` format 
 - Get Item:
     - `_get_item()` now leverages the get_row function from SCDL (so we eliminate the need to store and parse information in metadata.json)
- Error Handling for Genes not in the Tokenizer Vocabulary:
   - We add an optional parameter to SingleCellDataset and SingleCellDataModule called `bypass_tokenizer_vocab` which is by default `False`. So by default, we throw an error if a gene ID is not in the tokenizer vocabulary. If a user wants to bypass this, they can change `bypass_tokenizer_vocab` to `True`.
-  Error Handling for Genes with Zero Expression Values:
   - We throw an invalid input error in the cases that certain cells have no gene expression values (i.e. `sc_dataset.scdl.get_item()` returns `[]` for the gene data value) 
   
## Usage
The main change from a user perspective is to ensure that they convert their single cell h5ad files (or directories of h5ad files) to SingleCellMemmap format. 
1) For a single h5ad file, i.e. `data.h5ad`, they can simply run the following, where `output_path` is the file path the SingleCellMemmap directory should be written to:
`    SingleCellMemMapDataset(output_path, data.h5ad) `
2) For a directory of h5ad files, they can simply run the `convert_h5ad_to_scdl` script (more information available in the SCDL ReadMe). 

## Testing
We test that the updated SingleCellDataset produces the same output as the old dataset on synthetic samples and samples from the cellxsmall dataset. We also test for Megatron compatibility (as this dataset uses the MultiEpochDatasampler / Epoch Index) and for correct error handling of the above cases. 
```shell
pytest -vsub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_dataset.py
```
Note that we have also updated the following test files to use the MemMap dataset format + set bypass_tokenizer_vocab=True in them, because the cellxsmall dataset does have a few genes not in the HuggingFace tokenizer vocab and so the tests will error otherwise:
`sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py`
`scripts/singlecell/geneformer/test_train.py`
`sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_stop_and_go.py`


polinabinder1:
It would be useful to find the error handling cases to some sort of user documentation. I'm not sure of the best place.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -105,110 +105,30 @@ def __init__(  # noqa: D107
...
+        gene_data, col_idxs = np.array(values[0]), np.array(values[1])
+        if len(gene_data) == 0:
+            raise ValueError(
+                "SingleCellMemap data provided is invalid; the gene expression data parsed for the specified index is empty."
+            )
         return process_item(
             gene_data,
             col_idxs,
-            feature_ids,
+            feature_ids[0],
```
polinabinder1, line 131: Why are we only returning the first value here?

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_pydantic_train.py
```diff
@@ -77,9 +86,9 @@ def test_pretrain_cli_from_ckpt(tmpdir):
     assert (result_dir / "test-experiment").exists(), "Could not find test experiment directory."
 
 
-def test_pretrain_cli(tmpdir):
+def test_pretrain_cli(tmpdir, data_path):
     """trains from scratch"""
-    data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+    # data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
```
polinabinder1, line 91: remove comment?

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_train_geneformer.py
```diff
@@ -29,15 +29,21 @@
 from bionemo.testing import megatron_parallel_state_utils
```
polinabinder1, line 29: It would be helpful to add a test case with bypass_tokenizer_vocab=False and ensuring that the correct error is raised.

polinabinder1 (changes requested):

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -46,19 +45,19 @@ class SingleCellDataset(Dataset):
...
-            - `features.csv`: Feature subset associated with each sample.
+        data_path (str): Path where the single cell files are stored in SingleCell Memmap format. It should contain the following files:
+            - `metadata.json`: Path containing the number of rows int he dataset.
             - Gene expression matrix stored in CSR format as `numpy.memmap`:
-                - `gene_expression_data.npy`: Gene expression values.
-                - `gene_expression_ind.npy`: Gene indices associated with gene values.
-                - `gene_expression_ptr.npy`: Column indices for each sample.
+                - `data.npy`: Non-zero gene expression values.
+                - `col_ptr.npy`: Indices of the corresponding genes for each entry in data.npy.
+                - `row_ptr.npy`: Column index pointers for each cell sample.
```
skothenhill-nv, line 53: i love this change so much

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -105,110 +105,30 @@ def __init__(  # noqa: D107
...
         self._seed = seed
         self.eos_token = eos_token
-        # check if column indices are increasing for looking up genes. This is a way of spotting if the sc_memmap.py
-        #  script produced properly strctured sparse files.
-        self.assert_increasing_columns = assert_increasing_columns
-        path = Path(data_path)
 
-        # - metadata
-        metadata = json.load(open(path / "metadata.json", "r"))
+        self.scdl = SingleCellMemMapDataset(data_path)
```
skothenhill-nv, line 110: I'm confused by the semantics here, see my comment.

skothenhill-nv:
I can't leave a comment on the cell type classification notebook- but I see you are calling out to:
`!convert_h5ad_to_scdl --data-path {input_dir} --save-path {data_dir}`

But from reading the description of this PR, in the case of single datasets this is unnecessary.

```python
# For a single h5ad file, i.e. data.h5ad, they can simply run the following, where output_path is the file path the 
# SingleCellMemmap directory should be written to:
SingleCellMemMapDataset(output_path, data.h5ad)
```

More generally, I'm confused by the semantics here, in SingleCellDataset, we do this:

```
        self.scdl = SingleCellMemMapDataset(data_path)
```

which implies that data_path is populated with the scdl files already. If `SingleCellMemMapDataset` has a side effect of creating the scdl mmap files, we should make this flow extremely explicit (or maybe even a transformation of some kind). In anycase, could you add some clarity here? 

It seems like the intended workflow is:
```
convert_h5ad_to_scdl --data-path {input_dir} --save-path {data_dir}
self.scdl = SingleCellMemMapDataset({data_dir})
```

skothenhill-nv:
The project `README.md` needs to be updated with the new data-paths. What's there now is grabbing the files for the old `sc_memmap.py`, this goes for both `train_geneformer.py` and `bionemo-geneformer-recipe`. I suspect the finetuning examples need fixing up too. It might also be a good place to discuss the conversion script?

I think the inference script needs updating too (Line 54)
```
sub-packages/bionemo-geneformer/src/bionemo/geneformer/scripts/infer_geneformer.py
```

`docs/` looks correct, you fixed the only change (cell type classification notebook).

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -46,19 +45,19 @@ class SingleCellDataset(Dataset):
...
+                - `col_ptr.npy`: Indices of the corresponding genes for each entry in data.npy.
+                - `row_ptr.npy`: Column index pointers for each cell sample.
         tokenizer: The tokenizer to use for tokenizing the input data.
         median_dict (dict, optional): A dictionary containing median values for each gene. Defaults to None.
         max_len (int, optional): The maximum length of the input sequence. Defaults to 1024.
+        bypass_tokenizer_vocab (bool, optional): Allows you to bypass enforcing that all gene ensemble IDs in the dataset are in the tokenizer vocab. Defaults to False.
 
     Attributes:
-        data_path (str): Path where the single cell files are stored.
+        data_path (str): Path where the single cell files are stored in single cell memmap format.
```
skothenhill-nv, line 60: scdl memmap format? idk

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -84,7 +83,7 @@ class SingleCellDataset(Dataset):
 
     def __init__(  # noqa: D107
         self,
-        data_path: str | Path,
+        data_path: str,
```
skothenhill-nv, line 86: is this intentional?

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -105,110 +105,30 @@ def __init__(  # noqa: D107
...
-                self.dataset_map[i] = k
-                count += self.metadata[k]["shape"][0]
-            self.dataset_ccum[0] = -1
-        else:
-            # We can store a single feature_id vector for all datasets, and do not need to store the full metadata array.
-            logging.warning(
-                "Feature ids are the same across datasets. This is good, using the same feature_ids for all datasets."
-            )
-            self.feature_ids = feature_ids
-            self.metadata = None
```
skothenhill-nv, line 176: this feels fantastic

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -237,6 +159,8 @@ def _gather_medians(
             if normalize:
                 med = gene_median[tok]  # If not in the dictionary we default to no normalization (1)
                 medians.append(med)
+        elif not bypass_tokenizer_vocab:
+            raise ValueError("Provided gene id " + str(gene) + " not in tokenizer vocab.")
```
skothenhill-nv, line 163:
```suggestion
            raise ValueError(f"Provided gene id {str(gene)} not in tokenizer vocab.")
```

sub-packages/bionemo-geneformer/src/bionemo/geneformer/run/config_models.py
```diff
@@ -116,6 +116,7 @@ def construct_data_module(self, global_batch_size: int) -> SingleCellDataModule:
             persistent_workers=self.num_dataset_workers > 0,
             pin_memory=False,
             num_workers=self.num_dataset_workers,
+            bypass_tokenizer_vocab=True,
```
skothenhill-nv, line 119: do you think it makes sense to expose this in the config class? or would that add too much noise?

skothenhill-nv: Docstring of this class needs to be updated, it references `sc_memmap`, might make sense to do a search over the whole repo for references to `sc_memmap`

sub-packages/bionemo-geneformer/src/bionemo/geneformer/scripts/infer_geneformer.py
```diff
@@ -47,6 +47,7 @@ def infer_model(
     num_nodes: int = 1,
     num_dataset_workers: int = 0,
     config_class: Type[BioBertConfig] = GeneformerConfig,
+    bypass_tokenizer_vocab: bool = False,
 ) -> None:
     """Inference function (requires DDP and only training data that fits in memory)."""
     # This is just used to get the tokenizer :(
```
skothenhill-nv, line 53: the line right below this downloads data in the old format

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_pydantic_train.py
```diff
@@ -18,15 +18,25 @@
...
 from pathlib import Path
 
+import pytest
 from lightning.fabric.plugins.environments.lightning import find_free_network_port
 
 from bionemo.core.data.load import load
 
 
-data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+# data_path: Path = load("single_cell/testdata-20240506", source="pbss") / "cellxgene_2023-12-15_small" / "processed_data"
```
skothenhill-nv, line 27: can we remove this comment?

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_pydantic_train.py
```diff
@@ -117,9 +126,9 @@ def test_pretrain_cli(tmpdir):
     assert (result_dir / "test-experiment").exists(), "Could not find test experiment directory."
 
 
-def test_finetune_cli(tmpdir):
+def test_finetune_cli(tmpdir, data_path):
     """Uses CLI to invoke the entrypoint"""
-    data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+    # data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
```
skothenhill-nv, line 131: remove comment?

skothenhill-nv (approved):
Looks really good, no issues with any of the code.

There are some dangling references to the old data format around, I left them in a mix of comments outside this review and within the review. Unfortunately since your PR got pushed back, changes that were introduced between then probably made it hard to catch everything in the first pass.

Great job! Once you correct the various docstrings/load functions feel free to merge. 
:shipit:

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
jstjohn: My preference would be to allow either. If an internal method only works on one then you could do:

`self.data_path: str = str(data_path)` to make sure whichever a user passes in lands as a str internally.

jstjohn: on line 100 currently.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
jstjohn: I hope it's because feature_ids is a length 1 list always? If so I would add an assert len(feature_ids) == 1 or something before pulling out the 0th element. A comment about the shapes etc would be helpful.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_stop_and_go.py
```diff
@@ -46,7 +46,7 @@
 from bionemo.testing.harnesses.mode import Mode
 
 
-DATA_PATH: pathlib.Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+DATA_PATH: pathlib.Path = load("single_cell/testdata-memmap-format", source="pbss") / "cellxgene_2023-12-15_small_mmap"
```
jstjohn, line 49: Can you remove all of the `source="pbss"` from your tests etc. The default when you run can be changed with an environment variable. For example if you want to test locally and you want the test to use pbss then run:

```
export BIONEMO_DATA_SOURCE="pbss"
py.test [your cmd here]
```

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py
```diff
@@ -64,7 +64,7 @@
 nemo2_release_checkpoint_path: Path = load("geneformer/10M_240530:2.0")
 nemo_1_per_layer_outputs_path: Path = load("single_cell/nemo1-geneformer-per-layer-outputs")
 nemo_1_expected_values_path: Path = load("single_cell/nemo1-geneformer-golden-vals")
-data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+data_path: Path = load("single_cell/testdata-memmap-format", source="pbss") / "cellxgene_2023-12-15_small_mmap"
```
jstjohn, line 67: See comment on test_stop_and_go.py about not setting source manually here.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/conftest.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.fixture
+def cellx_small_directory() -> Path:
+    """Gets the path to the directory with with cellx small dataset in Single Cell Memmap format.
+
+    Returns:
+        A Path object that is the directory with the specified test data.
+    """
+    return load("single_cell/testdata-memmap-format", source="pbss") / "cellxgene_2023-12-15_small_mmap"
```
jstjohn, line 51: See comment near end about not setting `source="pbss"` manually in the test, and how you can change the behavior locally for yourself with an env variable.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/conftest.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.fixture
+def test_directory_feat_ids() -> Path:
+    """Gets the path to the directory with the synthetic single cell data (with the feature ids appended).
+
+    Returns:
+        A Path object that is the directory with specified test data.
+    """
+    return load("scdl/sample_scdl_feature_ids", source="pbss") / "scdl_data_with_feature_ids"
```
jstjohn, line 41: same around not setting source

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_train_geneformer.py
```diff
@@ -29,15 +29,21 @@
...
 
 
-data_path: Path = load("single_cell/testdata-20240506") / "cellxgene_2023-12-15_small" / "processed_data"
+@pytest.fixture
+def data_path() -> Path:
+    """Gets the path to the directory with with cellx small dataset in Single Cell Memmap format.
+    Returns:
+        A Path object that is the directory with the specified test data.
+    """
+    return load("single_cell/testdata-memmap-format", source="pbss") / "cellxgene_2023-12-15_small_mmap"
```
jstjohn, line 38: don't set source manually, use the environment variable instead. Note on how to do that as the last comment below.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_pydantic_train.py
```diff
@@ -18,15 +18,25 @@
...
 
 
-def test_bionemo2_rootdir():
+@pytest.fixture
+def data_path() -> Path:
+    """Gets the path to the directory with with cellx small dataset in Single Cell Memmap format.
+    Returns:
+        A Path object that is the directory with the specified test data.
+    """
+    return load("single_cell/testdata-memmap-format", source="pbss") / "cellxgene_2023-12-15_small_mmap"
```
jstjohn, line 36: same here about not setting source

docs/docs/user-guide/examples/bionemo-geneformer/geneformer-celltype-classification.ipynb
jstjohn, line 1: make sure you can run the example without setting `source=...` in the load lines. That will not work for users on brev.dev for example. Let's get these files into ngc.

sub-packages/bionemo-core/src/bionemo/core/data/resources/scdl.yaml
```diff
@@ -5,3 +5,9 @@
   sha256: 7a4237537bf535dfa00301ce8cc7073e0a23d5bc8aa902ad65db9f51b57a6df9 # pragma: allowlist secret
   owner: Polina Binder <pbinder@nvidia.com>
   description: Sample test data for SCDL.
+
+- tag: sample_scdl_feature_ids
+  pbss: "s3://bionemo-ci/test-data/scdl_feat_ids_sample_updated.tar.gz"
```
jstjohn, line 10: can you add any file that is part of a test or part of the notebook example into ngc? @dorotat-nv can help.

sub-packages/bionemo-core/src/bionemo/core/data/resources/single_cell.yaml
```diff
@@ -21,3 +21,9 @@
   sha256: ab038b184de52e53ff7bcea5e01d97d55944c507db88c0495bdf9e5e9e0303a4 # pragma: allowlist secret
   owner: John St John <jstjohn@nvidia.com>
   description: Golden values for geneformer QA model.
+
+- tag: testdata-memmap-format
+  pbss: "s3://bionemo-ci/test-data/singlecell/ordered_singlecell_mmap_test_data.tar.gz"
```
jstjohn, line 26: same here on adding to NGC. @dorotat-nv can help with that.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_stop_and_go.py
jstjohn: We do this in CI as well. Once you know stuff works and we like the files then you should work with @dorotat-nv to get them into NGC so the demo notebook etc still runs externally on brev.dev.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/run/config_models.py
savitha-eng: @skothenhill-nv @jstjohn  I assumed that since the data will probably have tokens not in the HuggingFaceTokenizer Vocab, it would be fine to pass it in as True, but can add it in if you think it would be helpful. Also, in general, do we want our default behavior to be bypass_tokenizer_vocab=True (instead of bypass_tokenizer_vocab=False), even when we do expose the API to make it user configurable?  @polinabinder1  was in favor of using ypass_tokenizer_vocab=True.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/scripts/infer_geneformer.py
savitha-eng: Yup, I believe it's just to generate the tokenizer so it shouldn't matter, but will go ahead and change it for consistency!

savitha-eng:
\build-ci

sub-packages/bionemo-scdl/tests/bionemo/scdl/conftest.py
```diff
@@ -29,7 +29,8 @@ def test_directory() -> Path:
     Returns:
         A Path object that is the directory with test data.
     """
-    return load("scdl/sample") / "scdl_data"
+    # return load("scdl/sample") / "scdl_data"
+    return load("scdl/sample_scdl_feature_ids", source="pbss") / "scdl_data_with_feature_ids"
```
trvachov, line 33: Do all the rest of our tests still use pbss as the data source? There was a recent push to move everything to public NGC.

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_pydantic_train.py
trvachov: nit picking github etiquette -- please just write "removed" or "fixed" and allow Steven to resolve, unless you both met and agreed that you (Savitha) can resolve it.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/scripts/infer_geneformer.py
```diff
@@ -233,6 +236,12 @@ def get_parser():
         help="Micro-batch size. Global batch size is inferred from this.",
     )
 
+    parser.add_argument(
+        "--skip-unrecognized-vocab-in-dataset",
+        action="store_false",
+        help="Set to False to verify whether all gene identifers are in the user supplied tokenizer vocab. Defaults to True which means that any gene identifier not in the user supplied tokenizer vocab will be excluded.",
```
trvachov, line 242: I think this docstring is very confusing...it's got some implied double negatives. Usually, when `--some_flag` is attached in the CLI, it implies "some_flag" = true. In short, even though "store_false" is an argparse option, I suggest changing this arg to be "store_true" and modifying the help message as well.

sub-packages/bionemo-geneformer/src/bionemo/geneformer/data/singlecell/dataset.py
```diff
@@ -277,6 +206,7 @@ def process_item(  # noqa: D417
         dirichlet_alpha (float): Alpha value for dirichlet sampling if set by `probabilistic_dirichlet_sampling`. Defaults to 0.5.
         same_length (bool): when true, sample the same length of genes as you originally had before the dirichlet sampler.
         recompute_globals (bool): when true, global arrays are always recomputed. this is only useful for testing.
+        skip_unrecognized_vocab_in_dataset (bool):  skip_unrecognized_vocab_in_dataset (bool, optional): Set to False to verify whether all gene identifers are in the user supplied tokenizer vocab. Defaults to True which means that any gene identifier not in the user supplied tokenizer vocab will be excluded.
```
trvachov, line 209: Again, suggest something like this:
```
If true, any gene identifier not in the user supplied tokenizer vocab will be excluded. If false, a hard-check is performed to verify all gene identifers are in the user supplied tokenizer vocab.

trvachov (changes requested):

trvachov:
Looks fine overall except there's one CLI arg where the understand the boolean logic can be better explained, I think.

trvachov (approved):

polinabinder1 (approved):

jstjohn (approved):

PR #482, opened 2024-11-26
Distributed Inference Support for ESM2/Geneformer

farhadrgh:
## Summary

This PR adds `PredictionWriter` callback of `BasePredictionWriter` that supposed to capture the `batch_indices` from dataloader's batch sampler and `predictions` from `predict_step` of the Lightning module. However, `batch_indices` are not being tracked properly and that prevents us from mapping the predictions to input sequences. Here I am listing the issues that are blocking distributed inference (writing to disk) in Bionemo 2.0. The first one does not have a clear workaround in Bionemo and requires changes in NeMo. The 2nd and 3rd issue listed below are not blocking right now.

Reference: #319 

----------------------
### 1.  NeMo `MegatronDataSampler` is not tracking the batch_indices.

`MegatronDataSampler` is [transforming the dataloaders](https://github.com/NVIDIA/NeMo/blob/1757ff9ed10272bf5ee7332d64fccd4bd9676f1b/nemo/lightning/pytorch/plugins/data_sampler.py#L67) by adding megatron batch sampler. But [Lightning prediction loop](https://github.com/Lightning-AI/pytorch-lightning/blob/8ad3e29816a63d8ce5c00ac104b14729a4176f4f/src/lightning/pytorch/loops/prediction_loop.py#L304) tracks `batch_indices` only if the batch sampler is an instance of `lightning.pytorch.overrides.distributed._IndexBatchSamplerWrapper`. To get this fixed I propose the following change in [`nemo.lightning.data.add_megatron_sampler`](https://github.com/NVIDIA/NeMo/blob/1757ff9ed10272bf5ee7332d64fccd4bd9676f1b/nemo/lightning/data.py#L218)

```python
def add_megatron_sampler(...)
    ...
    return DataLoader(
        dataloader.dataset,
        batch_sampler=_IndexBatchSamplerWrapper(batch_sampler),
        num_workers=dataloader.num_workers,
        pin_memory=dataloader.pin_memory,
        persistent_workers=dataloader.persistent_workers,
        collate_fn=dataloader.collate_fn,
    )
```
### Solution:
Fixed by NVIDIA/NeMo/pull/10934

### 2. Lightning `_PredictionLoop` not storing data for prediction writer callback

Again in [Lightning prediction loop](https://github.com/Lightning-AI/pytorch-lightning/blob/8ad3e29816a63d8ce5c00ac104b14729a4176f4f/src/lightning/pytorch/loops/prediction_loop.py#L235) if `data_fetcher` is an instance of `lightning.pytorch.loops.fetchers._DataLoaderIterDataFetcher`, data (including `batch_indices`) for prediction writer is not being stored. This is always the case since [NeMo Megatron Strategy always wraps the data fetcher](https://github.com/NVIDIA/NeMo/blob/1757ff9ed10272bf5ee7332d64fccd4bd9676f1b/nemo/lightning/pytorch/strategies/megatron_strategy.py#L833) with `_DataLoaderIterDataFetcher`. This is also a question for Lightning team as I don't undrestand the reason behind skipping datastore for this type of data fetcher.

### Solution:
A temporary workaround I proposed for this issue is to get the indices from `trainer` (`trainer.predict_dataloaders.batch_sampler.seen_batch_indices`). However, this only works for "epoch" writer intervals. We now can optionally return `input_ids` with the predictions (`--include-input-ids`) that should be used as a reliable method to map predictions to input sequences instead.

### 3. Lightning bug when return_predictions is False in `trainer.predict`

This is also related to the 2nd issue listed here. If the `data_fetcher` is an instance of `_DataLoaderIterDataFetcher`, lightning prediction loop skips creating the `any_on_epoch` which is referenced later in the loop:

```bash
File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/prediction_loop.py", line 124, in run
  self._predict_step(batch, batch_idx, dataloader_idx, dataloader_iter)
File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/prediction_loop.py", line 271, in _predict_step
  if self._return_predictions or any_on_epoch:
UnboundLocalError: local variable 'any_on_epoch' referenced before assignment
```
### Solution:
This is also a lightening bug. No solution at the moment. We are avoid setting `restrun_predictions=False`


jstjohn (approved):
Great work! Can you also verify that the examples in the notebooks can make use of this change? If you wouldn't mind adding the same change to geneformer that would be helpful as well.

sub-packages/bionemo-llm/src/bionemo/llm/utils/callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+import torch
+from lightning.pytorch.callbacks import BasePredictionWriter
+
+from bionemo.llm.lightning import batch_collator
+
+
+IntervalT = Literal["epoch", "batch"]
+
+
+class PredictionWriter(BasePredictionWriter, pl.Callback):
```
pstjohn, line 30: Can you add tests for this somewhere? Ideally we'd use this in both the DDP and single-device paths, otherwise it's going to be hard to write tests for this that run in CI

farhadrgh: yes, makes sense, added here: [6a18c19](https://github.com/NVIDIA/bionemo-framework/pull/482/commits/6a18c19d8bbaf98c3612b450024584de099d779f)

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/infer_esm2.py
```diff
@@ -135,11 +145,13 @@ def infer_model(
...
-    non_none_keys = [key for key, val in results_dict.items() if val is not None]
-    print(f"Writing output {str(non_none_keys)} into {results_path}")
-    torch.save(results_dict, results_path)
+    if devices > 1:
+        trainer.predict(module, datamodule=datamodule)  # return_predictions=False lightning bug
+    else:
+        predictions = batch_collator(trainer.predict(module, datamodule=datamodule, return_predictions=True))
+        non_none_keys = [key for key, val in predictions.items() if val is not None]
+        print(f"Writing output {str(non_none_keys)} into {results_path}")
+        torch.save(predictions, results_path)
```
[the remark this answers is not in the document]
farhadrgh: [bf43836](https://github.com/NVIDIA/bionemo-framework/pull/482/commits/bf43836a19ff55ee2905d9de2e1c74ab7a11cd60)

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/infer_esm2.py
```diff
@@ -101,12 +104,19 @@ def infer_model(
         find_unused_parameters=True,
     )
 
+    callbacks = []
+    if devices > 1:
```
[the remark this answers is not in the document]
farhadrgh: [bf43836](https://github.com/NVIDIA/bionemo-framework/pull/482/commits/bf43836a19ff55ee2905d9de2e1c74ab7a11cd60)

sub-packages/bionemo-llm/src/bionemo/llm/utils/callbacks.py
```diff
@@ -0,0 +1,100 @@
...
+    """A callback that writes predictions to disk at specified intervals during training."""
+
+    def __init__(self, output_dir: str | os.PathLike, write_interval: IntervalT):
+        """Initializes the callback.
+
+        Args:
+            output_dir (str): The directory where predictions will be written.
+            write_interval (str): The interval at which predictions will be written. (batch, epoch)
```
pstjohn, line 39: maybe just remove the type hints in these docstring, you already have them in the function def

pstjohn (commented):
I mean, this looks ok, but is this really the API we're forced into? Having to write out pytorch files per rank and per batch / epoch seems like a pretty unfriendly usage pattern.

It would be good to revisit this if we could get pytorch metrics working; I'd like to see what a 'distributed evaluation' workflow looks like. Do we need to write these tensors all to disk, and then load them and calculate the evaluation metric? How would this work as some sort of gRPC / triton inference service?

pstjohn (approved):

farhadrgh:
I pushed a change to fix the lightning bug when `return_prediction=False`: https://github.com/Lightning-AI/pytorch-lightning/pull/20484
Once that is merged we can use the `batch` write interval to do online prediction properly.

PR #485, opened 2024-11-27
update collate mask_value

pstjohn:
We use a mask value of -100 elsewhere for these masked tokens, so we should be consistent in the collate function. This will be especially important if we use torchmetrics, which only wants a single value to ignore (rather than all negative values)


farhadrgh (approved):

jstjohn (approved):
Nice default.

PR #492, opened 2024-12-03
Disable validation/test stages in ESM-2 and Geneformer

sichu2023:
## Summary
Allow `--limit-val-batches=0.` to disable validation. Same applies for testing in geneformer.

## Details
Previously validation dataset cannot be created when `limit_val_batches = 0.`. Now this is caught and the validation dataset is set to `None`. Same applies to test dataset.

## Tests
Implemented in `sub-packages/bionemo-geneformer/tests/bionemo/esm2/scripts/test_train_esm2.py::test_val_dataloader_in_main_runs_with_limit_val_batches` and that in geneformer.

## Out of scope
- The e2e test function only test whether it runs. It does not test whether validation is actually skipped.
- Currently there is no testing on finetuning.

### DIFF
--- a/sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_train_geneformer.py
+++ b/sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_train_geneformer.py
@@ -46,12 +46,16 @@ def test_val_dataloader_in_main_runs_with_limit_val_batches(tmpdir, limit_val_ba
             data_dir=data_path,
             num_nodes=1,
             devices=1,
-            seq_length=128,
+            seq_length=8,
             result_dir=result_dir,
             wandb_project=None,
             wandb_offline=True,
             num_steps=5,
+<<<<<<< HEAD
             limit_val_batches=1,
+=======
+            limit_val_batches=limit_val_batches,
+>>>>>>> c7291696a (reduce geneformer e2e test burden)
             val_check_interval=2,
             num_dataset_workers=0,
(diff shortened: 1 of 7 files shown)


jstjohn (approved):
Can you also add a test to lock in this support?

sichu2023:
> Can you also add a test to lock in this support?

Will do. Just disabled auto-merge.

farhadrgh (changes requested):
Would you add this support to infer/fintetune datamodule [here](https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/datamodule.py) as well?

sichu2023:
> Would you add this support to infer/fintetune datamodule [here](https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/datamodule.py) as well?

Done!

farhadrgh (approved):

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/scripts/test_train_geneformer.py
```diff
@@ -37,7 +37,8 @@ def test_bionemo2_rootdir():
     assert data_path.is_dir(), "Test data directory is supposed to be a directory."
 
 
-def test_main_runs(tmpdir):
```
pstjohn, line 40: can we make sure this is running a super small version of the model? Any other CLI options we can change to make this run as fast as possible would be great.

pstjohn (approved):

PR #502, opened 2024-12-05
Set up our pytest suite for the codecov integration

pstjohn:
This will require @dorotat-nv to add the CODECOV_TOKEN secret to our blossom runners.

### DIFF
--- a/ci/scripts/run_pytest.sh
+++ b/ci/scripts/run_pytest.sh
@@ -15,7 +15,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+# Parse arguments
+CODECOV_DISABLED=0
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --no-codecov) CODECOV_DISABLED=1 ;;
+        --help)
+            echo "Usage: $0 [--no-codecov] [--help]"
+            echo "  --no-codecov    Disables installing and uploading coverage results to codecov"
+            echo "  --help          Display this help message"
+            exit 0
+            ;;
+    esac
+    shift
+done
+
 set -xueo pipefail
+
+
 export PYTHONDONTWRITEBYTECODE=1
 # NOTE: if a non-nvidia user wants to run the test suite, just run `export BIONEMO_DATA_SOURCE=ngc` prior to this call.
 export BIONEMO_DATA_SOURCE="${BIONEMO_DATA_SOURCE:-pbss}"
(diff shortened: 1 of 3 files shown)


ci/scripts/run_pytest.sh
```diff
@@ -25,13 +25,33 @@ if ! set_bionemo_home; then
...
 
+# integrity check
+curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
+curl -Os https://cli.codecov.io/latest/linux/codecov
+curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM
+curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM.sig
+gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
+
+shasum -a 256 -c codecov.SHA256SUM
+sudo chmod +x codecov
```
trvachov, line 39: any way we can make this work without sudo? I worry `run_pytest.sh` will be run in some CI environment where sudo isn't available.

trvachov: Thinking out loud, you might want to make this bash script with a command like arg to optionally disable codecov

ci/scripts/run_pytest.sh
```diff
@@ -25,13 +25,33 @@ if ! set_bionemo_home; then
...
+    pytest -v --nbval-lax --cov=bionemo --cov-append --junitxml=$(basename $dir).junit.xml -o junit_family=legacy $dir
 done
 
-python -m coverage combine
-python -m coverage report --show-missing
+# Merge all sub-directory test results into a single xml file.
+junitparser merge *.junit.xml combined.junit.xml
+
+# Upload test analytics to codecov.
+./codecov do-upload --report-type test_results --file combined.junit.xml
```
trvachov, line 53: maybe I don't know how codecov works, but where is this being uploaded?

ci/scripts/run_pytest.sh
pstjohn: might be easier to just install `codecov-cli` in our requirements-test.txt

ci/scripts/run_pytest.sh
pstjohn: Yeah IIUC this will upload by default to the standard codecov cloud, using our CODECOV_TOKEN to identify / authenticate

ci/scripts/run_pytest.sh
dorotat-nv, line 39: I would set integration with `codedev` as optional, controlled by a flag to the script, ie `--codedev`

dorotat-nv (changes requested):

ci/scripts/run_pytest.sh
pstjohn: ok, added a `--no-codecov` option

malcolmgreaves (approved):

PR #505, opened 2024-12-06
[Draft] Unify inference interface - request for comments

skothenhill-nv:
# Overview

Unifies the inference interface between esm2 and geneformer. Unlike training, user configuration is limited to parallelism arguments, which has no impact on the output, and inference specific arguments (checkpoint, data, etc). Due to this, I think we should stick with argparse style parsing and instead focus on validating the inputs we can and unifying the inference workflows for our models.

## Proposed workflow:
```
bionemo-infer --checkpoint path/to/ckpt
    --model-config-t  ESM2Config  # (Exposed or not is fine here)
    --data-config-t InMemoryCSVDatasetConfig
    --data-args arg1=val arg2=val arg3=val 
    --checkpoint-path /path/to/ckpt
    --precision
    --pipeline-model-parallel
    ... 
    (other runtime args)
```

 ## Pseudocode (in python) then looks like:
 ```
 args = parse_args()
 data_config[args.DataConfigT](args.data1, args.data2... etc)
 parallel_config = ParallelConfig(args.parallel_arg1, parallel_arg2... etc)

infer(args.ckpt, data_config, parallel_config, result_path, include_embeddings/logits/etc)
```

### DIFF
--- a/sub-packages/bionemo-esm2/src/bionemo/esm2/run/config_models.py
+++ b/sub-packages/bionemo-esm2/src/bionemo/esm2/run/config_models.py
@@ -223,6 +223,7 @@ class ExposedESM2PretrainConfig(ExposedModelConfig[ESM2Config]):
         ), "Must set variable_seq_lengths to True when min_seq_length != max_seq_length under pipeline or tensor parallelism."
         return global_cfg
 
-    def model_class(self) -> Type[ESM2Config]:
+    @classmethod
+    def model_class(cls) -> Type[ESM2Config]:
         """Returns the model class associated with this configuration."""
         return ESM2Config
--- a/sub-packages/bionemo-llm/src/bionemo/llm/infer.py
+++ b/sub-packages/bionemo-llm/src/bionemo/llm/infer.py
@@ -1,0 +1,164 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-Apache2
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+from pathlib import Path
+import pathlib
+from typing import Dict, Optional, Sequence, Type, cast, get_args
+
+import torch
+
+from bionemo.llm.run import config_models
+from bionemo.llm.model.biobert import model
+
+from bionemo.llm.run.config_models import DataConfig, ExperimentConfig, ExposedModelConfig, OptimizerSchedulerConfig, ParallelConfig, TrainingConfig, DataModuleT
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: sub-packages/bionemo-llm/src/bionemo/llm/infer.py [hunk 1]
```diff
@@ -0,0 +1,164 @@
...
+    include_hiddens: bool = False,
+    include_embeddings: bool = False,
+    include_logits: bool = False,
+    include_input_ids: bool = False,
+):
+
+    if isinstance(results_path, str):
+        results_path = pathlib.Path(results_path)
+
+    if os.path.isdir(results_path):
```
farhadrgh, line 63: This will change as we will require `results_path` to be a directory for storing results.

farhadrgh (commented):
Thanks @skothenhill-nv. This looks solid to me and makes a lot of sense overall. I like the approach and I don’t see any major gaps.

I don’t think this impacts the proposal, but I want to point out a difference between the ESM2 and Geneformer datamodule/dataset. For ESM2, I started with an in-memory CSV dataset and a datamodule that’s separate from the pre-training datamodule (which uses pairs of .db and parquet files) to simplify the inference process.

skothenhill-nv:
> Thanks @skothenhill-nv. This looks solid to me and makes a lot of sense overall. I like the approach and I don’t see any major gaps.
> 
> I don’t think this impacts the proposal, but I want to point out a difference between the ESM2 and Geneformer datamodule/dataset. For ESM2, I started with an in-memory CSV dataset and a datamodule that’s separate from the pre-training datamodule (which uses pairs of .db and parquet files) to simplify the inference process.

Thanks Farhad, just waiting on feedback from @jstjohn before moving forward with this. Glad to hear we are aligned and thanks for the extra context.

sub-packages/bionemo-esm2/src/bionemo/esm2/run/config_models.py
```diff
@@ -223,6 +223,7 @@ def model_validator(self, global_cfg: MainConfig) -> MainConfig:
         ), "Must set variable_seq_lengths to True when min_seq_length != max_seq_length under pipeline or tensor parallelism."
         return global_cfg
 
-    def model_class(self) -> Type[ESM2Config]:
+    @classmethod
+    def model_class(cls) -> Type[ESM2Config]:
```
jstjohn, line 227: I'd say either make it a class method and return `cls` or static and return the one class? Right now there's no change in terms of correct usage of class method vs instance method. Static would be the lowest featured method type that still works here.

sub-packages/bionemo-llm/src/bionemo/llm/infer.py
```diff
@@ -0,0 +1,164 @@
...
+        config=bionemo_model_config, tokenizer=datamodule.tokenizer, optimizer=None
+    )
+    # setup results
+    pathlib.Path(data_config.result_dir).mkdir(parents=True, exist_ok=True)
+    # setup lightning module
+
+    # TODO will want to bring in Farhad's work, eventually we should rebase on his PR
+    # Does this have to be a DataModule or can it be a DataLoader?
+    predictions = trainer.predict(module, datamodule=datamodule, return_predictions=True)
+    # TODO evaluate this section. I worry about gathering predictions in memory.
```
jstjohn, line 136: So in the future we should be able to say `return_predictions=False` and use a prediction writer. I think you were getting at that with mention of Farhad's PR but maybe specifically say that we should get return_predictions=False working alongside a prediction writer.

sub-packages/bionemo-esm2/src/bionemo/esm2/run/config_models.py
skothenhill-nv: oh good callout!

malcolmgreaves (approved):

PR #523, opened 2024-12-11
Remove quotes for Jupyter command on startup in init guide

tshimko-nv:
Fix: Remove problematic quotes from startup command

This PR fixes an issue where the docker run command for BioNeMo 2 failed due to incorrect quoting of the JupyterLab startup command. The fix removes the quotes around the command.


cspades (approved):
string to script, lgtm

dorotat-nv (approved):

PR #524, opened 2024-12-11
Flexible memory management to avoid fragmentation-related CUDA OOM

farhadrgh:
The environment variable `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` is part of PyTorch’s CUDA memory management configuration introduced to address memory fragmentation issues. When PyTorch allocates GPU memory, it reserves chunks of memory from the GPU for its own use. Over time, the GPU memory can become fragmented—even though there may be enough total memory available, the memory might not be contiguous.


docs/docs/user-guide/examples/bionemo-esm2/mutant-design.ipynb
pstjohn, line 1: I'm not a huge fan of the behavior changing quietly depending on the hardware; it seems like this could bite us later. Could we just run 650m for all hardware? I'd prefer just to have a commented out line saying "load 3b if you have a larger GPU" or something like that.

farhadrgh: Sure, I will add a note and switch to 650M

farhadrgh: [5fa6e9b](https://github.com/NVIDIA/bionemo-framework/pull/524/commits/5fa6e9b879d21c0077cc9d7cbcb5cb95e06001a1)

trvachov (approved):

pstjohn (approved):

PR #531, opened 2024-12-13
RowFeatureIndex Optimization

polinabinder1:
With this PR, if an identical dataframe is appended to RowFeatureIndex as the last one it, instead of storing it, the counter corresponding to the previous dataframe is incremented. This becomes an issue in very large datasets. 

Previously, if we had dataframe A corresponding to row [0,2] and we wanted to add the same dataframe A corresponding to 4 more rows, we would store dataframe A twice, with the first copy corresponding to rows [0,2] and the second to [2,6].

Now, we would store dataframe A once and it would correspond to rows [0,6].


sub-packages/bionemo-scdl/tests/bionemo/scdl/index/test_row_feature_index.py
```diff
@@ -113,6 +176,28 @@ def test_feature_index_internals_on_append(create_first_RowFeatureIndex):
...
 
+def test_feature_index_internals_on_append_same_features(create_first_RowFeatureIndex):
+    one_feats = {"feature_name": np.array(["FF", "GG", "HH"]), "feature_int": np.array([1, 2, 3])}
+    create_first_RowFeatureIndex.concat(create_first_RowFeatureIndex)
+    # append(8, two_feats, len(two_feats["feature_name"]), "MY_DATAFRAME")
+    assert len(create_first_RowFeatureIndex) == 1
+    assert create_first_RowFeatureIndex.number_vars_at_row(1) == 3
+    assert create_first_RowFeatureIndex.number_vars_at_row(13) == 3
+    assert create_first_RowFeatureIndex.number_vars_at_row(19) == 3
+    assert create_first_RowFeatureIndex.number_vars_at_row(2) == 3
```
skothenhill-nv, line 187: i think im still missing something, but isnt the key thing checking that theyre the same `RowFeatureIndex`? or (thinking outloud) was the behavior before something like:

[FF, GG, HH, FF, GG, HH], 0=>6

and now its
[FF,GG,HH] => 3 
?

skothenhill-nv:
Is the Megatron-LM change intentional? or do we need to do the `git submodule update recursive` thing. it shows a diff again main, which seems weird.

sub-packages/bionemo-scdl/tests/bionemo/scdl/index/test_row_feature_index.py
polinabinder1: They are all the same. I can check this with the for loop.

polinabinder1: Before, it would have been 3 at every position. But it would have been stored as [[FF,GG,HH] , [FF,GG,HH] ]

sub-packages/bionemo-scdl/src/bionemo/scdl/index/row_feature_index.py
```diff
@@ -28,6 +28,20 @@
 __all__: Sequence[str] = ("RowFeatureIndex",)
 
 
+def are_dicts_equal(dict1, dict2):
```
pstjohn, line 31: could you move the type hints to the definition line? Might also be better to re-phrase the function name to make it more explicit what it returns
```suggestion
def dicts_are_equal(dict1: dict[str, np.ndarray], dict2: dict[str, np.ndarray]) -> bool:
```

sub-packages/bionemo-scdl/tests/bionemo/scdl/index/test_row_feature_index.py
```diff
@@ -20,7 +20,53 @@
...
+    assert are_dicts_equal(dict1, dict4) is False
+
+
+def test_different_lengths(dict1):
+    smaller_dict = {"a": np.array([1, 2, 3])}
+    assert are_dicts_equal(dict1, smaller_dict) is False
+
+
+def test_empty_dicts(empty_dict):
+    assert are_dicts_equal(empty_dict, empty_dict) is True
```
pstjohn, line 69: Don't use fixtures here, I know it reduces duplication but it also makes you have to search to find the dicts that are being compared. Just redefine the dicts each time and make each test take no fixture inupts

pstjohn (approved):

skothenhill-nv (approved):

PR #545, opened 2024-12-18
Run all pytests even if submodules fail

pstjohn:
Attempt to run pytests in all submodules even if individual packages fail.


farhadrgh (approved):

trvachov (approved):
Thank you for the quick action!

jstjohn (approved):

PR #547, opened 2024-12-18
xFail known bad tests on H100 and fix CVEs

gagank1:
Known issue on H100 (and GH200) with loading checkpoints. Also fixing CVE in ARM container


cspades (approved):
LGTM!

PR #549, opened 2024-12-18
xFail known bad tests on H100 and fix CVEs

gagank1:


cspades (approved):

Dockerfile.arm
```diff
@@ -312,6 +312,7 @@ COPY --from=rust-env /usr/local/rustup /usr/local/rustup
 
 
 # RUN rm -rf /usr/local/cargo /usr/local/rustup
+RUN rm -rf /root/.cache/bazel
```
pstjohn, line 315: If the idea was to reduce image size, you should put this command in the same `RUN` where the cache is generated. Docker layers are additive; so that `.cache` directory would still be shipped; alongside the layer removing it 😄

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py
```diff
@@ -260,6 +260,7 @@ def __getitem__(self, idx):
         return {"text": self.input_ids[idx], "attention_mask": self.mask[idx]}
 
 
+@pytest.mark.xfail(reason="Known issue on H100 GPUs")
```
pstjohn, line 263:
```suggestion
@pytest.mark.xfail("H100" in torch.cuda.get_device_name(0), reason="Known issue on H100 GPUs")
```
Can we xfail this only if we're on an H100? (confirm this works)

pstjohn (approved):

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py
jstjohn: Agreed. Also you can do something like this instead:

```
from transformer_engine.pytorch.utils import get_cudnn_version
from transformer_engine.pytorch.utils import get_device_compute_capability

@pytest.mark.xfail(get_cudnn_version() < (9, 5, 0) and get_device_compute_capability() >= (9, 0), reason="Known issue on hopper GPUs with older cuDNN versions pre 9.5.")
```

jstjohn: When we're on hopper we see really good precision with cudnn >= 9.5.0 at least, so I'm guessing it would work in that case.

jstjohn (approved):
Approve but please make the test only xfail when not on H100 (or see my suggestion about also testing for cudnn version since I think it should work again with newer cuDNN, eg when we upgrade the base pytorch container to 24.10-py3.)

skothenhill-nv (approved):

Dockerfile.arm
gagank1: This was actually to fix CVEs from the bazel cache

sub-packages/bionemo-geneformer/tests/bionemo/geneformer/test_model.py
gagank1: Done, checking for Hx00 GPU now

PR #553, opened 2024-12-23
update base image to 24.12

pstjohn:
Attempts to update the base image to the most recent 24.12 release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 #   https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/blob/main/.gitlab-ci.yml
 #  We should keep versions in our container up to date to ensure that we get the latest tested perf improvements and
 #   training loss curves from NeMo.
-ARG BASE_IMAGE=nvcr.io/nvidia/pytorch:24.10-py3
+ARG BASE_IMAGE=nvcr.io/nvidia/pytorch:24.12-py3
 
 FROM rust:1.82.0 as rust-env
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,11 +73,6 @@ RUN rm -rf /build
 
 # Addressing Security Scan Vulnerabilities
 RUN rm -rf /opt/pytorch/pytorch/third_party/onnx
-RUN apt-get update  && \
-  apt-get install -y openssh-client=1:8.9p1-3ubuntu0.10 && \
-  rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


trvachov:
Did you build-ci work given that this is a fork? I actually am not certain our ci will work on forks (...this is bad if it doesn't, since fork PRs should probably be the norm so that we dont have the branch explosion that we have today).

pstjohn:
> Did you build-ci work given that this is a fork? I actually am not certain our ci will work on forks (...this is bad if it doesn't, since fork PRs should probably be the norm so that we dont have the branch explosion that we have today).

Yeah it works on forks; that's pretty essential given that we're open source and not adding write-access external collaborators. IIUC that should be the recommended way to make PRs at this point, otherwise we clog up the main repo with a bunch of forks

Dockerfile
```diff
@@ -108,12 +103,23 @@ ENV PATH="/usr/local/cargo/bin:/usr/local/rustup/bin:${PATH}"
...
+# in the source distribution doesn't match the expected pypi version.
 RUN --mount=type=bind,source=./.git,target=./.git \
   --mount=type=bind,source=./requirements-test.txt,target=/requirements-test.txt \
   --mount=type=bind,source=./requirements-cve.txt,target=/requirements-cve.txt \
   <<EOF
 set -eo pipefail
-uv pip install maturin --no-build-isolation && uv pip install --no-build-isolation \
+uv pip install maturin --no-build-isolation --break-system-packages
+
+pip install --use-deprecated=legacy-resolver  --no-build-isolation --break-system-packages \
```
trvachov, line 115: can you please add a comment on why we need this, and also add a global comment to the dockerfile as to why `--break-system-packages` is used (I know it's because of the new python venv requirement...but let's just document it to make it easier for everyone.

trvachov (changes requested):

trvachov (approved):

Dockerfile
pstjohn: So I added a single `UV_BREAK_SYSTEM_PACKAGES` env variable; looks like the NGC torch image team already set the equivalent pip variable; 
```Dockerfile
ENV PIP_BREAK_SYSTEM_PACKAGES=1
```
https://gitlab-master.nvidia.com/dl/dgx/pytorch/-/blob/main/Dockerfile.base#L43

pstjohn:
sudo-merging after CI passed due to folks being out of the office

PR #555, opened 2024-12-24
[BUFIX] fail when passed fastas with duplicate sequence ids

skothenhill-nv:
When NvFaidx was used on Fasta files containing duplicate sequence ids, which violates the FASTA spec, it would silently fail and use the last-seen sequence as an entry.

This PR fails by default and exposes a parameter to ignore sequence_ids and use integer indexing instead.


cspades (approved):

edawson (approved):
Looks good, and appreciate the tests - approved.

PR #560, opened 2024-12-31
Use self-hosted azure runners for bionemo unit tests

pstjohn:


pablo-garay (approved):

pablo-garay (approved):

pstjohn:
Looks like `test-results-action` isn't working correctly, since this PR is coming from a fork rather than an internal branch and doesn't have access to repository secrets.
So I think the options are 
1. Only run test analytics on `main` branch runs
2. figure out tokenless test analytics uploads
3. use `on: pull_request_target` and make sure we understand the security implications there

ci/scripts/run_pytest.sh
```diff
@@ -36,13 +36,6 @@ EOF
     exit "${1:-0}"
 }
 
-# Function to handle cleanup on script exit
-cleanup() {
-    local exit_code=$?
-    [ -n "${coverage_files[*]:-}" ] && rm -f "${coverage_files[@]:-}"
-    exit "$exit_code"
-}
-
```
pstjohn, line 45: We need these coverage files to persist to the next step in the GHA actions pipeline so we can upload them to codecov. Do these stick around on the blossom workers?

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,119 @@
...
+      volumes:
+        - /home/azureuser/actions-runner-bionemo/cache:/root/.cache
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: ${{ github.run_id }}
+
+      - name: Run tests
+        run: ./ci/scripts/run_pytest.sh --no-nbval
```
pstjohn, line 104: we could eventually make this configurable based on the PR label 🤷

dorotat-nv, line 104: could you add here mechanism for PR labels: SKIP_CI and INCLUDE_NOTEBOOKS_TESTS to define the scope of tests or omit them completely?

dorotat-nv (commented):
Reviewed, and added ask to add mechanism for PR labels, otherwise LGTM

pstjohn:
Testing from a branch rather than a fork in #587

PR #556, opened 2024-12-25
Duplicates of Peter's changes for CI

trvachov:
Duplicate of Peter's PR (https://github.com/NVIDIA/bionemo-framework/pull/553) for CI testing.

PR #566, opened 2025-01-02
update devcontainer for new ubuntu base image

pstjohn:
The new ubuntu base container contains a couple of changes that breaks the (untested in CI) base container:
1. it now has a default 1000:1000 `ubuntu` user we can use, instead of creating a new bionemo user.
2. it uses python 3.12, which changes some of our copy paths.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -24,7 +24,7 @@
 		"NUMBA_CACHE_DIR": "/tmp/"
 	},
 	"postCreateCommand": "./.devcontainer/postCreateCommand.sh",
-	"remoteUser": "bionemo",
+	"remoteUser": "ubuntu",
 	"customizations": {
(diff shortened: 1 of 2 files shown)


trvachov:
Peter since you're patching various docker things, can you add this? https://github.com/NVIDIA/bionemo-framework/pull/556/files#r1900994582

Fixes security issues.

trvachov (changes requested):
Request to add rm -rf bazel line into docker image to fix security issues coming from that build garbage (I think it introduces 1 CRIT and 16 HIGH vulns, based on last scan).

trvachov (approved):
LGTM. Let's look at container scan results before merge...I expect ~2 HIGH CVEs.

jstjohn (approved):

dorotat-nv (approved):

PR #567, opened 2025-01-02
don't eagerly download esm2 checkpoints

pstjohn:
Adding the load calls in the global namespace leads to these checkpoints being downloaded even when we don't use them. We really should call `load` as lazily as possible to avoid materializing data for tests we don't need.


jstjohn (approved):
ship

PR #569, opened 2025-01-03
run pytest with or without docs and notebooks in run_pytest.sh

dorotat-nv:
### Description
We've streamlined our testing workflow to improve CI efficiency. By default, documentation tests and notebook validations are skipped to reduce pipeline duration.

### Type of changes

- [x]  Refactor
- [x]  Documentation update
- [x]  Other (please describe): user guideline, updated to CI workflow

### CI Pipeline Configuration
Configure CI behavior by checking relevant boxes below. This will automatically apply labels.

- [ ] [SKIP_CI](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/user-guide/contributing/contributing.md#skip_ci) - Skip all continuous integration tests
- [x] [INCLUDE_NOTEBOOKS_TESTS](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/user-guide/contributing/contributing.md#include_notebooks_tests) - Execute notebook validation tests in pytest

> [!NOTE]
### Usage
When running tests locally, use these flags to control test execution:
```bash
./ci/scripts/run_pytest.sh --skip-docs
./ci/scripts/run_pytest.sh --no-nbval
./ci/scripts/run_pytest.sh --no-nbval --skip-docs
```
Control test pipeline execution by adding these labels to your PR:
`RUN_DOCS` - Enables documentation testing
`RUN_NOTEBOOKS_VALIDATION` - Enables notebook validation

### Pre-submit Checklist


ci/scripts/run_pytest.sh
```diff
@@ -23,17 +23,33 @@ export BIONEMO_DATA_SOURCE="${BIONEMO_DATA_SOURCE:-pbss}"
...
 
 python -m coverage erase
 
 error=false
+# Build pytest options dynamically
+pytest_options="-v --durations=0 --durations-min=60.0"
+if [ "$USE_NBVAL" = true ]; then
+    pytest_options="$pytest_options --nbval-lax"
+fi
 for dir in docs/ ./sub-packages/bionemo-*/; do
```
pstjohn, line 49: I think the better approach might be to just not run pytest in the `docs/` directory?
That might give us a better separation of "example notebooks" that we don't want to run in pre-merge testing, vs. testing functionality that we expect to work inside a jupyter notebook.

dorotat-nv: good point, replaced --no-nbval flag with --skip-docs

pstjohn (approved):
awesome, LGTM, thanks!

ci/scripts/run_pytest.sh
```diff
@@ -23,14 +23,30 @@ export BIONEMO_DATA_SOURCE="${BIONEMO_DATA_SOURCE:-pbss}"
...
+
+# Loop through directories and skip docs/ if the flag is set
 for dir in docs/ ./sub-packages/bionemo-*/; do
+    if [ "$SKIP_DOCS" == true ] && [ "$dir" == "docs/" ]; then
+        echo "Skipping pytest in $dir because --skip-docs flag is set"
+        continue
+    fi
     echo "Running pytest in $dir"
     python -m coverage run --parallel-mode --source=bionemo \
     -m pytest -v --nbval-lax --durations=0 --durations-min=60.0 "$dir" || error=true
```
jstjohn, line 52: This will still run the `examples` long notebook tests that are within regular packages due to the `--nbval-lax` argument. Is the intention not to run notebooks? If so maybe make that the argument and if so skip both docs and drop `--nbval-lax`.

jstjohn (approved):
Approved, but see my comment about the `--nbval-lax` argument. Maybe change the option to `--skip-notebooks` and if that's set then drop that kwarg.

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,56 @@
...
+    types: [opened, edited, synchronize]
+
+jobs:
+  manage-labels:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v3
```
pstjohn, line 14: do we need this checkout line?

.github/pull_request_template.md
```diff
@@ -1,45 +1,33 @@
...
+- [ ]  Bug fix (non-breaking change which fixes an issue)
+- [ ]  New feature (non-breaking change which adds functionality)
+- [ ]  Refactor
+- [ ]  Documentation update
+- [ ]  Other (please describe):
 
-_High level summary of changes. Try to keep this as short and informative as possible: less is more._
+### CI Pipeline Configuration
+Check the boxes below to configure CI (continuous integration) behavior. These will automatically apply labels.
+> **Note:** Leave boxes unchecked by default, check only if you want to modify default behavior.]
```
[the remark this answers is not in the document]
dorotat-nv: thanks!, will update

.github/workflows/pr-labels.yml
dorotat-nv: I double checked and no, we dont need that, I will correct

jstjohn:
One thought: when you run docs, there is _only_ notebook validation in there. I am a bit worried a user will add a notebook to docs, set `RUN_NOTEBOOKS_VALIDATION` and not realize they _also_ need to add `RUN_DOCS_VALIDATION` if their notebook happens to be in the docs folder rather than a sub-module. Basically you currently support:

* No notebook tests (default)
* Only notebooks tests in sub-packages (if you say `RUN_NOTEBOOKS_VALIDATION`)
* All notebooks tests (if you say both `RUN_NOTEBOOKS_VALIDATION` and `RUN_DOCS_VALIDATION`)

Note that only saying `RUN_DOCS_VALIDATION` would be more or less a no-op since the only tests in there are notebook tests so without `RUN_NOTEBOOKS_VALIDATION` also it would not do anything. 

My proposal would be to, for now, only have `RUN_NOTEBOOKS_VALIDATION` and have that option check in docs as well as sub-packages for notebooks to test. Thoughts? Also I don't think the argument would _only_ run notebooks, it would include notebooks. So maybe the least confusing thing would be to say `INCLUDE_NOTEBOOKS_VALIDATION` and then it's clearer that all tests would run but notebooks would be included.

PR #570, opened 2025-01-03
pin importlib-resources

pstjohn:
Temporary pin for importlib-resources due to an undeclared dependency in a recently published version (this morning) in importlib-resources.

python/importlib_resources#323


trvachov (approved):
can you please just add the explicit keyword "TODO" here? otherwise lgtm.

pstjohn:
Fixed with new importlib version; no longer needed

PR #573, opened 2025-01-06
Refactor dockerfile for better caching and avoid pbss download in notebook test

pstjohn:
Re-orders some of the Dockerfile commands to avoid re-building `tensorstore` from source every docker build.
Also removes the pbss download from `test_load_notebook.ipynb` since this can't succeed outside the nvidia VPN


dorotat-nv (approved):

dorotat-nv:
the pbbs download was replaced with ngc in`test_load_notebook`??

jstjohn (approved):

pstjohn:
> the pbbs download was replaced with ngc in`test_load_notebook`?? 


We always had both. The NGC one previously failed due to their use of the async library, but we patched that to work (hence the test). But if we want this to work on the azure runners we shouldn't test pbss

PR #582, opened 2025-01-08
default to overlap_param_gather

sichu2023:
Turn on `overlap_param_gather` for an additional 4% MFU gain on 15B model pretraining

Dependent on NeMo2 fix on `overlap_param_gather` https://github.com/NVIDIA/NeMo/commit/bde672e75f1ac45ead08e2b977920a28eb81448e.


pstjohn (approved):
Looks fine. I groan a bit at the double negative `not no-overlap-param-gather`, but I suppose it's consistent.

The alternative would be to make --overlap-param-gather not `store_true` and default to True, so you would pass `--overlap-param-gather False` as the CLI argument

jstjohn (approved):
Looks good!

PR #588, opened 2025-01-10
add codecov badge

pstjohn:
Adds a codecov badge to the README and removes a duplicate pre-commit workflow.

Because jobs can only depend on other jobs in the same workflow file (IIUC), we put the pre-commit workflow in the new azure CI definition. That way, the docker build and test steps only trigger if the pre-commit results pass


cspades (approved):

jstjohn (approved):

.github/workflows/pre-commit.yml
```diff
@@ -1,26 +0,0 @@
-name: pre-commit
```
jstjohn, line 1: So... what's this removal doing? Doesn't seem to be related to code-cov? Worth a comment in the PR.

pstjohn: Yeah my apologies, added a description to the PR description. But essentially we moved this to the new azure workflow file, and it's still (correctly) being picked up as a required task since the name is the same

PR #626, opened 2025-01-21
hotfix for some failing python tests due to NGC files being moved around

pstjohn:
Fixes failing tests due to NGC resource files being moved around, and renames some of the ESM-2 resources to better delineate checkpoints we trained on our internal data from those converted from the original ESM-2 model weights.

### DIFF
--- a/sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune_regressor.py
+++ b/sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune_regressor.py
@@ -26,15 +26,9 @@ from bionemo.esm2.model.finetune.finetune_regressor import (
 from bionemo.testing import megatron_parallel_state_utils
 
 
-# To download a 8M internally pre-trained ESM2 model
-@pytest.fixture
-def pretrain_ckpt_path():
-    return load("esm2/nv_8m:2.0")
-
-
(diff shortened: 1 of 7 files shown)


sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune_regressor.py
```diff
@@ -27,7 +27,9 @@
 
 
 # To download a 8M internally pre-trained ESM2 model
-pretrain_ckpt_path = load("esm2/nv_8m:2.0")
+@pytest.fixture
+def pretrain_ckpt_path():
```
dorotat-nv, line 31: maybe it can be moved to conftest?

dorotat-nv (approved):

sub-packages/bionemo-esm2/tests/bionemo/esm2/model/finetune/test_finetune_regressor.py
pstjohn: I actually went the other way and just moved the load calls into the tests themselves. I don't really like fixtures; they sort of obfuscate where the data is coming from, and in this case the `load()` call itself is less code than having to pass that fixture around

jstjohn (approved):

PR #631, opened 2025-01-22
Loss disagreement between TP=1 and TP=2

sichu2023:
### Description
Loss values from `unreduced_token_loss_fn` disagrees even though the `logits` are the same when TP=1 and TP=2. Interestingly, this introduces a nearly-constant offset in training curve and does not impact `grad_norm`. The offset slowly shrinks over 100 training steps.

Suspected root cause:
`vocab_parallel_cross_entropy` takes into account of `logits` at the padded vocab dimensions while it should have ignored them. As training progresses, more probabilities are then assigned to "real vocabs" so the padding vacabs slowly becomes less relevant, leading to the closing gap between TP=1 and TP=2 loss curves.

`logits` dimensions also expanded from 128 to 256 when replacing `ColumnParallelLinear` with `torch.nn.Linear` although the reason is still unclear.

On a related issue, `unreduced_token_loss_fn` introduces inplace operation on the token logits.
https://github.com/NVIDIA/bionemo-framework/blob/sichu/loss-curve-tp-shift/sub-packages/bionemo-llm/tests/bionemo/llm/model/test_loss.py#L158

### Type of changes

### Usage

```bash
DATADIR=$(download_bionemo_data esm2/testdata_esm2_pretrain:2.0 --source pbss)/2024_03_sanity

NUM_DEVICES=${1}
TP=${NUM_DEVICES}

test_data_flags="""
    --train-cluster-path=${DATADIR}/train_clusters_sanity.parquet \
    --train-database-path=${DATADIR}/train_sanity.db \
    --valid-cluster-path=${DATADIR}/valid_clusters.parquet \
    --valid-database-path=${DATADIR}/validation.db
"""
hparam_flags_8m="""
    --num-layers=6 \
    --hidden-size=320 \
    --num-attention-heads=20 \
    --ffn-hidden-size=1280 \
    --micro-batch-size=4
"""
test_run_flags="""
    --num-nodes=1 \
    --num-gpus=${num_devices} \
    --num-steps=100 \
    --limit-val-batches=2
"""

launcher="torchrun --nproc-per-node=${NUM_DEVICES}"
script="sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/train_esm2.py"
cmd="""
${launcher} $script \
    $test_data_flags \
    $hparam_flags_8m \
    $test_run_flags \
    --log-every-n-steps=5 \
    --val-check-interval=5 \
    --tensor-model-parallel-size=${TP}
"""
eval $cmd
```

### Pre-submit Checklist

### DIFF
--- a/sub-packages/bionemo-esm2/src/bionemo/esm2/model/model.py
+++ b/sub-packages/bionemo-esm2/src/bionemo/esm2/model/model.py
@@ -21,7 +21,6 @@ from typing import Callable, Literal, Optional, Sequence, Type, TypeVar
 
 import torch
 import torch.distributed
-from megatron.core import tensor_parallel
 from megatron.core.models.bert.bert_lm_head import BertLMHead
 from megatron.core.models.bert.pooler import Pooler
 from megatron.core.models.common.embeddings.rotary_pos_embedding import RotaryEmbedding
@@ -178,16 +177,21 @@ class ESM2Model(MegatronBioBertModel):
                 config,
             )
 
-            self.output_layer = tensor_parallel.ColumnParallelLinear(
-                config.hidden_size,
-                self.vocab_size,
-                config=config,
-                init_method=config.init_method,
-                bias=True,
-                skip_bias_add=False,
-                gather_output=not self.parallel_output,
-                skip_weight_param_allocation=pre_process and share_embeddings_and_output_weights,
-            )
+            # TODO replace tensor_parallel.ColumnParallelLinear with torch.nn.Linear to debug; remove once complete
+            # self.output_layer = tensor_parallel.ColumnParallelLinear(
+            #     config.hidden_size,
+            #     self.vocab_size,
+            #     config=config,
+            #     init_method=config.init_method,
+            #     is_expert=False,
+            #     bias=True,
+            #     skip_bias_add=False,
+            #     gather_output=not self.parallel_output,
+            #     skip_weight_param_allocation=pre_process and share_embeddings_and_output_weights,
+            #     embedding_activation_buffer=self.embedding_activation_buffer,
+            #     grad_output_buffer=self.grad_output_buffer,
+            # )
+            self.output_layer = torch.nn.Linear(config.hidden_size, self.vocab_size, bias=True)
 
             self.binary_head = None
             if self.add_binary_head:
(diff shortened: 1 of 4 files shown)


sichu2023:
Loss curve when initializing `torch.nn.Linear` with `torch.init.zeros_` under `output_layer_init_method`.
W B Chart 1_21_2025, 6_40_00 PM
W B Chart 1_21_2025, 6_40_05 PM

sub-packages/bionemo-esm2/src/bionemo/esm2/model/model.py
```diff
@@ -178,16 +177,21 @@ def __init__(
...
+            #     init_method=config.init_method,
+            #     is_expert=False,
+            #     bias=True,
+            #     skip_bias_add=False,
+            #     gather_output=not self.parallel_output,
+            #     skip_weight_param_allocation=pre_process and share_embeddings_and_output_weights,
+            #     embedding_activation_buffer=self.embedding_activation_buffer,
+            #     grad_output_buffer=self.grad_output_buffer,
+            # )
+            self.output_layer = torch.nn.Linear(config.hidden_size, self.vocab_size, bias=True)
```
jstjohn, line 194: This probably has bugs with `bias` getting out of sync between TP ranks during training FYI. See https://jirasw.nvidia.com/browse/BIONEMO-668 and https://jirasw.nvidia.com/browse/BIONEMO-666 and https://nvidia.slack.com/archives/C074Z808N05/p1737508003987919 and https://nvidia.slack.com/archives/C0434FDLPQV/p1733963545314469

Also if your concern is when you do TP=2 that the logit dim is 1/2 that may be because columnparallellinear splits along the logit vocab dimension, and ideally vocab parallel cross entropy knows how to reduce across this.

sichu2023: Yup. Only temporarily placed `torch.nn.Linear` to debug. Will revert back to `ColumnParallelLinear` after so.

sichu2023: I am less concerned about "1/2 logits dim" (128 dim) but more concerned about `torch.nn.Linear` giving 256 dim on TP=2.

malcolmgreaves (approved):

PR #635, opened 2025-01-22
catch ngc api key validation errors and default to not using an api key

pstjohn:
Should fix issues like https://nvbugspro.nvidia.com/bug/5063178, where the SLURM environment picks up a stray and incorrect NGC api key.


trvachov (approved):
awesome peter thanks for being proactive in fixing these things.

farhadrgh (approved):

PR #638, opened 2025-01-22
Instructions for uploading a package to pypi

polinabinder1:
These are instructions for uploading a package to pypi.


jwilber:
Nice!
I wonder if we should include some note with the minimum expected things we should have for PyPi, e.g.:
- LICENSE
- `pyproject.toml`-specific metadata (`author`, `license`, `description`, etc.)
- etc.

(but that may be outside the scope of this file)

jwilber (approved):

jstjohn (approved):

PR #640, opened 2025-01-22
Draft AMPLIFY PR

pstjohn:


sub-packages/bionemo-amplify/tests/bionemo/amplify/test_convert.py
```diff
@@ -0,0 +1,270 @@
...
+        zip(hf_results["encoder_block_outputs"], nemo_results["encoder_block_outputs"], strict=True)
+    ):
+        assert_cosine_similarity(
+            hf_block_output,
+            nemo_block_output,
+            attention_mask.cpu(),
+            rtol=1e-4,
+            atol=1e-4,
+            msg=f"Encoder block output {i}",
+        )
```
pstjohn, line 87: So yeah, I have really no idea what to make of this. The accuracy starts degrading steadily through each attention block; this test fails after the second block with an error of ~0.01.

I tried using a non-TE attention block, as well as turning off all the fusions with no luck. Some other things to try would be omitting the mask tokens in the test inputs and see if that helps figure out where this is coming from.

sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
```diff
@@ -0,0 +1,330 @@
...
+            self.embedding = LanguageModelEmbedding(
+                config=self.config,
+                vocab_size=self.vocab_size,
+                max_sequence_length=self.max_sequence_length,
+                position_embedding_type=position_embedding_type,
+                num_tokentypes=num_tokentypes,
+            )
+
+        if self.position_embedding_type == "rope":
+            self.rotary_pos_emb = RotaryEmbedding(
```
cspades, line 172: Random question, where is this RoPE used if it's not passed into `TransformerBlock`?

cspades (approved):
Very clean, mostly reviewed `model.py` and `convert.py`, LGTM.

So now we just need to add training, inference, and dataset/dataloader?

sub-packages/bionemo-amplify/tests/bionemo/amplify/test_convert.py
farhadrgh: Was this fixed by using `DotProductAttention` instead of `TEDotProductAttention` in layer spec?

sub-packages/bionemo-amplify/src/bionemo/amplify/convert.py
```diff
@@ -0,0 +1,163 @@
...
+
+from bionemo.amplify.model import AMPLIFYConfig
+from bionemo.amplify.tokenizer import BioNeMoAMPLIFYTokenizer
+from bionemo.llm.lightning import BionemoLightningModule
+from bionemo.llm.model.biobert.lightning import biobert_lightning_module
+
+
+@io.model_importer(BionemoLightningModule, "hf")
+class HFAMPLIFYImporter(io.ModelConnector[AutoModel, BionemoLightningModule]):
+    """Converts a Hugging Face ESM-2 model to a NeMo ESM-2 model."""
```
jstjohn, line 33: Should be amplify? Or will this also work for esm2?

sub-packages/bionemo-amplify/src/bionemo/amplify/convert.py
```diff
@@ -0,0 +1,163 @@
...
+    padding_rows = torch.zeros(num_padding_rows, source_embed.size(1))
+    return torch.cat((source_embed, padding_rows), dim=0)
+
+
+@io.state_transform(
+    source_key="lm_head.bias",
+    target_key="output_layer.bias",
+)
+def _pad_bias(ctx: io.TransformCTX, source_bias):
+    """Pad the embedding layer to the new input dimension."""
```
jstjohn, line 114: This is vocab right? Maybe call out that this supports changing vocab sizes or padding along the vocab dimension?

sub-packages/bionemo-amplify/src/bionemo/amplify/convert.py
```diff
@@ -0,0 +1,163 @@
...
+    nemo_embedding_dimension = ctx.target.config.make_vocab_size_divisible_by
+    hf_embedding_dimension = source_bias.size(0)
+    output_bias = torch.zeros(nemo_embedding_dimension, dtype=source_bias.dtype, device=source_bias.device)
+    output_bias[:hf_embedding_dimension] = source_bias
+    return output_bias
+
+
+@io.state_transform(
```
jstjohn, line 122: See if this function is already present in the Nemo transforms class, if it's a standard transform it might be good to use standard functions. There's an example of this for combining the w1 and w2 linear projections in EVO2

sub-packages/bionemo-amplify/src/bionemo/amplify/convert.py
```diff
@@ -0,0 +1,163 @@
...
+    # transpose weights
+    # [sequence length, batch size, num_splits_model_parallel * attention head size * #attention heads]
+    # --> [sequence length, batch size, attention head size * num_splits_model_parallel * #attention heads]
+    concat_weights = concat_weights.view(3, np, -1, query.size()[-1])
+    concat_weights = concat_weights.transpose(0, 1).contiguous()
+    concat_weights = concat_weights.view(*input_shape)
+    return concat_weights
+
+
+@io.state_transform(
```
jstjohn, line 144: Same, double check if this is already available in the class Evo2 imports for its HF importers. A lot of these are pretty common transforms.

sub-packages/bionemo-amplify/src/bionemo/amplify/dataset.py
```diff
@@ -0,0 +1,147 @@
...
+        # Initialize a random number generator with a seed that is a combination of the dataset seed, epoch, and index.
+        rng = np.random.default_rng([self.seed, index.epoch, index.idx])
+        if index.idx >= len(self):
+            raise IndexError(f"Index {index.idx} out of range [0, {len(self)}).")
+
+        sequence = self.protein_dataset[int(index.idx)]["sequence"]
+
+        # We don't want special tokens before we pass the input to the masking function; we add these in the collate_fn.
+        tokenized_sequence = self._tokenize(sequence)
+        cropped_sequence = _random_crop(tokenized_sequence, self.max_seq_length, rng)
```
jstjohn, line 118: This could use a comment.

sub-packages/bionemo-amplify/src/bionemo/amplify/hf_rotary.py
```diff
@@ -0,0 +1,111 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Code copied from https://huggingface.co/chandar-lab/AMPLIFY_350M/blob/main/rotary.py
+# Original license:
+# MIT License
+
+# Copyright (c) [year] [fullname]
```
jstjohn, line 20: This was the original license? Seems unlikely

jstjohn: I would expect the person who wrote the code's name and whatever date they used, right?

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
```diff
@@ -0,0 +1,337 @@
...
+from bionemo.llm.model.biobert.model import BiobertSpecOption
+from bionemo.llm.model.config import TorchmetricsConfig
+from bionemo.llm.utils.datamodule_utils import infer_global_batch_size
+from bionemo.llm.utils.logger_utils import WandbConfig, setup_nemo_lightning_logger
+
+
+app = typer.Typer()
+
+
+@app.command()
```
jstjohn, line 47: So this ingests the function and makes a cli? How is it?

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+    )
+
+    callbacks = [
+        RichModelSummary(max_depth=4),
+        LearningRateMonitor(),
+        nl_callbacks.PreemptionCallback(),
+        TimingCallback(),
+    ]
+    if nsys_profiling:
```
jstjohn, line 207: Maybe also add a Boolean cli option for the flops callback?

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
```diff
@@ -0,0 +1,337 @@
...
+        num_workers=num_dataset_workers,
+        random_mask_strategy=random_mask_strategy,
+        tokenizer=tokenizer,
+    )
+
+    # Configure the model
+    train_metric = None
+    is_model_parallel = tensor_model_parallel_size * pipeline_model_parallel_size > 1
+    if is_model_parallel:
+        valid_metric = None  # metric logging under model parallelism is not supported yet
```
jstjohn, line 254: You still get loss right? Maybe add a comment calling out that loss will still be logged.

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
```diff
@@ -0,0 +1,337 @@
...
+        tokenizer=tokenizer,
+        optimizer=MegatronOptimizerModule(
+            config=OptimizerConfig(
+                lr=lr,
+                optimizer="adam",  # fused_adam not supported
+                use_distributed_optimizer=False,
+                weight_decay=0.01,
+                adam_beta1=0.9,
+                adam_beta2=0.95,
+                clip_grad=1.0,
```
jstjohn, line 292: These are so important to tune. I'd recommend pulling these out as options with these defaults to main.

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
```diff
@@ -0,0 +1,337 @@
...
+            ),
+            lr_scheduler=nl.lr_scheduler.CosineAnnealingScheduler(
+                min_lr=0.1 * lr,
```
jstjohn, line 295: I would recommend making this an option to main with default None. If None then set it like this based on the user supplied (or default) lr

sub-packages/bionemo-amplify/tests/bionemo/amplify/test_convert.py
```diff
@@ -0,0 +1,298 @@
...
+
+def test_convert_amplify_120M(tmp_path):
+    model_tag = "chandar-lab/AMPLIFY_120M"
+    module = biobert_lightning_module(config=AMPLIFYConfig())
+    io.import_ckpt(module, f"hf://{model_tag}", tmp_path / "nemo_checkpoint")
+    with megatron_parallel_state_utils.distributed_model_parallel_state():
+        assert_amplify_equivalence(tmp_path / "nemo_checkpoint", model_tag, atol=1e-4, rtol=1e-4)
+
+
+def test_convert_amplify_120M_bf16(tmp_path):
```
jstjohn, line 277: Do you also have a downstream accuracy test? Some kind of biological function interpretation or zero shot or something task? Those are important since sometimes you can have good cosine similarity and trash downstream accuracy or vice versa.

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
pstjohn: It's great! 😄 . Definitely nicer than doing the whole thing via argparse

sub-packages/bionemo-amplify/tests/bionemo/amplify/test_convert.py
pstjohn: actually no -- it was fixed by me passing the wrong attention mask into the huggingface model 🤦 . 
They raised a warning about it here: https://huggingface.co/chandar-lab/AMPLIFY_120M/commit/e27bb0dff186f9508d13b6ad977c30f8b2e050ed

sub-packages/bionemo-llm/src/bionemo/llm/model/biobert/transformer_specs.py
```diff
@@ -221,5 +222,36 @@ def get_biobert_spec(  # noqa: D417
             )
             return esm2_bert_layer_local_spec
 
+        case BiobertSpecOption.amplify_with_transformer_engine_spec:
+            if core_attention is None:
+                core_attention = DotProductAttention
```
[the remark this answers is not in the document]
pstjohn: I can actually change that back to TEDotProductAttention

sub-packages/bionemo-amplify/src/bionemo/amplify/convert.py
pstjohn: but we need to tell it what keys to transform into what, right? via these transforms? I suppose I could skip the decorator and just call this function on an imported function directly

sub-packages/bionemo-amplify/src/bionemo/amplify/dataset.py
pstjohn: added, but this was largely copied from the ESM-2 dataset implementation.

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
pstjohn: this was copied from `train_esm2`. we need a common train entrypoint script that makes it easier to add models without copying and pasting this whole files. But I have no idea what this means tbh.

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
pstjohn: this is just the esm2 train script

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
pstjohn: it's clearer to the user if the default is just 1E-3 right? what's the benefit of having None as the default?

sub-packages/bionemo-amplify/src/bionemo/amplify/train_amplify.py
pstjohn: easy enough to do later if/when we need to change them. If AMGEN doesn't tune them i don't think we really need to either. Ideally we could do this as part of a config refactor, rather than having uncountably infinite CLI options

sub-packages/bionemo-amplify/tests/bionemo/amplify/test_convert.py
pstjohn: we could calculate the loss for a real protein perhaps

PR #648, opened 2025-01-23
fix devcontainer initialize script

pstjohn:
If this script exits unsuccessfully, the devcontainer launch stops. This makes sure that we exit successfully regardless of the status of ~/.netrc


pstjohn:
just going to force-merge this, it's not covered by any CI tests and is currently breaking the devcontainer

PR #650, opened 2025-01-23
edit to approval workflow to avoid marking a failed action

pstjohn:
Rather than having a bunch of red 'X's on PRs that haven't gotten the required +2 review, we could have a conditional step that only runs when the required approvals are met. If we mark this as the required step in our branch rules it has the same effect, without cluttering our PR status with a bunch of failed status checks

### DIFF
--- a/.github/workflows/approvals.yml
+++ b/.github/workflows/approvals.yml
@@ -1,6 +1,7 @@
 name: Enforce Tiered Approvals
 
 on:
+  pull_request: # TODO: remove
   pull_request_review:
 
 env:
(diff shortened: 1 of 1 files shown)


jomitchellnv (approved):

trvachov (approved):

.github/workflows/approvals.yml
```diff
@@ -7,9 +7,10 @@ env:
   TIER2_REVIEWERS: "jstjohn,trvachov,pstjohn"
 
 jobs:
```
pstjohn, line 9: test

PR #654, opened 2025-01-24
ESM2 Interactive inference

farhadrgh:
### Type of changes

- [ ]  Refactor
- [ ]  Documentation update

### CI Pipeline Configuration

> [!NOTE]

### Usage

```python
TODO: Add code snippet
```

### Pre-submit Checklist

### DIFF
--- a/sub-packages/bionemo-esm2/src/bionemo/esm2/testing/compare.py
+++ b/sub-packages/bionemo-esm2/src/bionemo/esm2/testing/compare.py
@@ -14,16 +14,14 @@
 # limitations under the License.
 
 
-import gc
 from pathlib import Path
 
 import torch
-from megatron.core.transformer.module import Float16Module
 from transformers import AutoModelForMaskedLM
 
 from bionemo.core.utils.dtypes import PrecisionTypes, get_autocast_dtype
 from bionemo.esm2.data.tokenizer import get_tokenizer
-from bionemo.esm2.model.model import ESM2Config
+from bionemo.esm2.testing.eval import eval_esm2
 
 
 def assert_model_equivalence(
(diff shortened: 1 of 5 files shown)


sub-packages/bionemo-esm2/src/bionemo/esm2/testing/eval.py
```diff
@@ -0,0 +1,86 @@
...
+import gc
+from pathlib import Path
+
+import torch
+from megatron.core.transformer.module import Float16Module
+
+from bionemo.core.utils.dtypes import PrecisionTypes, get_autocast_dtype
+from bionemo.esm2.data import tokenizer
+from bionemo.esm2.model.model import ESM2Config
+from bionemo.testing import megatron_parallel_state_utils
```
farhadrgh, line 26: I will have to move `megatron_parallel_state_utils` from `bionemo.testing` for this to work.
CC: @jstjohn @pstjohn

sub-packages/bionemo-esm2/src/bionemo/esm2/testing/compare.py
```diff
@@ -53,34 +51,14 @@ def assert_model_equivalence(
...
-        fp16=dtype is torch.float16,
-    )
-
-    nemo_model = nemo_config.configure_model(tokenizer).to("cuda").eval()
 
-    if dtype is torch.float16 or dtype is torch.bfloat16:
-        nemo_model = Float16Module(nemo_config, nemo_model)
-
-    nemo_output = nemo_model(input_ids, attention_mask)
+    nemo_output = eval_esm2(ckpt_path, test_proteins, precision=precision)
```
pstjohn, line 55: IMO I like to ensure the tensor inputs to the model are the same, rather than doing that twice here. Might make sense to decompose your eval_esm2 function into more modular pieces so that's possible

sub-packages/bionemo-esm2/src/bionemo/esm2/testing/eval.py
pstjohn: Yeah I had this same issue and almost put a thread together for it. The other option (which honestly might be better) is just to move the context manager outside of `eval_esm2`) and just ensure that `eval_esm2` is called inside one. Thats what I do with `assert_model_equivalence`

pstjohn: I.e., I think we can call into `bionemo.testing` in the example notebooks without tach yelling at us, and it makes the creation of the megatron context more explicit

farhadrgh: I figured we might run into cases where we iterate on a dataloader, so I separated the setup/teardown to avoid configuring the model every time we call forward. 

Now I can also add a method to check the input to eval and tokenize if they are sequences. But I don't think that would look clean

sub-packages/bionemo-esm2/src/bionemo/esm2/testing/compare.py
farhadrgh: I figured we might run into cases where we iterate on a dataloader, so I separated the setup/teardown to avoid configuring the model every time we call forward.

Now I can also add a method to check the input to eval and tokenize if they are sequences. But I don't think that would look clean

sub-packages/bionemo-esm2/src/bionemo/esm2/testing/eval.py
farhadrgh: yeah, I moved the context manager outside and into the notebooks, no complain from tach!

farhadrgh:
@pstjohn any idea why the golden value tests are not passing? I am getting `AssertionError: Tensor-likes are not close!` from `sub-packages/bionemo-esm2/tests/bionemo/esm2/model/test_convert.py`

pstjohn:
@farhadrgh looks like you're changing the model comparison eval function; I'm guessing we're not getting the same values out. Maybe try running through a debugger to see where the change is happening?

PR #655, opened 2025-01-24
Possible changes for biobert config changes

pstjohn:


sub-packages/bionemo-llm/src/bionemo/llm/model/biobert/model.py
```diff
@@ -129,32 +129,12 @@ class BioBertOutput(BioBertOutputCore, total=False):
...
 
-    def __init__(  # noqa: D107
+    def __init__(
         self,
-        config: TransformerConfig,
+        config: "BioBertConfig",
         num_tokentypes: int,
-        transformer_layer_spec: ModuleSpec,
-        vocab_size: int,
-        max_sequence_length: int,
```
pstjohn, line 157: Nearly all these arguments that we're passing around are _also_ in the `config` object, which we've mis-typed as a vanilla `TransformerConfig`

PR #659, opened 2025-01-24
Dependency graph

polinabinder1:


scripts/dependency_graph.py
```diff
@@ -0,0 +1,115 @@
...
+def find_pyproject_files(base_dir):
+    """Find all pyproject.toml files in subdirectories."""
+    pyproject_files = []
+    for root, _, files in os.walk(base_dir):
+        if "pyproject.toml" in files:
+            pyproject_files.append(os.path.join(root, "pyproject.toml"))
+    return pyproject_files
+
+
+def parse_dependencies(pyproject_path):
```
pstjohn, line 35: Do you think we could also parse tach.toml and possibly warn if the dependency graphs are different? The tach check actually enforces this separation during CI, so it's probably more accurate.

dorotat-nv: Is there a check in place to ensure that the tach.toml and pyproject.toml files are up-to-date and valid in terms of dependencies? For example, does PyPI installation and importing all subpackages automatically verify this?

We could implement regular checks or enforcement in the CI pipeline. If the above method isn't sufficient, we can create a script that parses the project.toml files of the main project and its subpackages, extracts the import paths used in Python scripts under the src directory of each subpackage, and verifies all imports starting with `from bionemo`.

dorotat-nv (commented):
I am not sure is pyproject.toml is the up to date source of dependency information anymore. I am not sure how it is maintained. 

If no, I thin we should implement a script that parses dependency graph for subpackages from the scripts, ie parse the project.toml files of the main project and its subpackages, extract the import paths used in Python scripts under the src directory of each subpackage, and verifies all imports starting with `from bionemo`.

scripts/dependency_graph.py
trvachov: Dorota -- I suggest we constrain scope for now to just drawing the dependency graph.

I agree it's a good idea to do what you're describing, but it would increase the scope of this substantially and at the moment there's other stuff we gotta do :)

trvachov:
To me this looks like a good start just need to document "how to run" script in github PR description. I don't necessary need this to do any more than in currently does ( @pstjohn  @dorotat-nv , I suggest we constrain our review just to "graph drawing" rather than any sort of py file parsing + CI enforcement.

scripts/dependency_graph.py
polinabinder1: The new code gets the tach.toml dependencies and checks that the code imports in the sub-packages are correct based on what is in the pyproject.toml and tach.toml files.

polinabinder1:
> Could this script be relocated under internal/scripts?

Done!

pstjohn (commented):
This looks great, the module was laid out nicely and you have a lot of clear, reusable functions.

I do think we need unit tests for these functions though. Copilot could likely do that pretty quickly. Ideally any public function / class / method should get a test for all its expected behavior and edge cases.

I also think we should add those resulting images to our documentation; along with a command of how to run this to regenerate them.

dorotat-nv (approved):

sub-packages/bionemo-fw/src/bionemo/fw/dependency_graph.py
```diff
@@ -0,0 +1,219 @@
...
+    )
+    found_imports = {}
+    for dir_name in directories:
+        directory = base_dir / dir_name / "src"
+        subpackages = set()
+
+        for root, _, files in os.walk(directory):
+            for file in files:
+                if file.endswith(".py"):
+                    file_path = os.path.join(root, file)
```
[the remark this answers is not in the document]
polinabinder1: That is easier. Fixed!

docs/docs/assets/images/sub_package_graphs/dependency_file_imports.png
pstjohn, line 1: would be cool to try a topo-sort plot if it's possible: https://networkx.org/documentation/stable/auto_examples/graph/plot_dag_layout.html

pstjohn (approved):
LGTM, feel free to self-resolve other comments. Thanks!

PR #672, opened 2025-01-31
Add initial amplify package structure

pstjohn:
Adds an empty sub-package to house future AMPLIFY model code. Also adds a missing __init__ file to the ESM test directory to distinguish future tests with similar names.

PR #673, opened 2025-01-31
Add model and tokenizer smoke tests

pstjohn:
Adds an initial AMPLIFY model class and tokenizer, and smoke tests that exercise these classes. 

Depends on #672

### DIFF
--- a/sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
+++ b/sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
@@ -1,0 +1,347 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-Apache2
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from dataclasses import dataclass
+from typing import Any, Callable, Literal, Optional, Sequence, Type, TypeVar
+
+import torch
+from megatron.core import tensor_parallel
+from megatron.core.models.bert.pooler import Pooler
+from megatron.core.models.common.embeddings.rotary_pos_embedding import RotaryEmbedding
+from megatron.core.transformer import spec_utils
+from megatron.core.transformer.enums import ModelType
+from megatron.core.transformer.identity_op import IdentityOp
+from megatron.core.transformer.module import MegatronModule
+from megatron.core.transformer.transformer_block import TransformerBlock
+from megatron.core.transformer.transformer_config import TransformerConfig
+from megatron.core.transformer.utils import get_linear_layer
+from torch import Tensor
+from torch.nn.functional import silu
+from torch.optim import Optimizer
+
+# from bionemo.amplify.data.tokenizer import BioNeMoAMPLIFYTokenizer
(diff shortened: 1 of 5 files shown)


sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
```diff
@@ -0,0 +1,354 @@
...
+        """Check compatibility between biobert_spec_option and apply_query_key_layer_scaling post initialization."""
+        super().__post_init__()
+        if self.biobert_spec_option == BiobertSpecOption.esm2_bert_layer_with_transformer_engine_spec:
+            self.apply_query_key_layer_scaling = False
+            if self.gated_linear_unit:
+                # To keep the number of parameters and the amount of computation constant, we reduce the number of
+                # hidden units by a factor of 2/3 (https://arxiv.org/pdf/2002.05202.pdf) and make it a multiple of 8 to
+                # avoid RuntimeError due to misaligned operand
+                multiple_of = 8
+                self.ffn_hidden_size = int(2 * self.ffn_hidden_size / 3)
```
ynashed, line 351: Just a heads up that after I tried this model with v2.2, I found out that this `__post_init__` function was being called twice. Leading to `ffn_hidden_size` being rescaled twice, leading to a model size that's roughly half the intended size. I fixed this by moving the rescaling to the model `__init__` https://github.com/ynashed/bionemo-framework/blob/7fe9c465c578c55f0ed938436306b43f1369ff52/sub-packages/bionemo-amplify/src/bionemo/amplify/model/model.py#L150. Maybe there's a better place.

sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
```diff
@@ -0,0 +1,354 @@
...
+    # core attention
+    use_esm_attention: bool = False  # Skip ESM2 custom attention for TE acceleration. Still passes golden value test.
+    attention_softmax_in_fp32: bool = False
+    normalize_attention_scores: bool = False
+
+    # From megatron.core.models.gpt.bert_model.GPTModel
+    fp16_lm_cross_entropy: bool = False  # Move the cross entropy unreduced loss calculation for lm head to fp16
+    parallel_output: bool = True
+    share_embeddings_and_output_weights: bool = True
+    make_vocab_size_divisible_by: int = 1
```
pstjohn, line 307: @ynashed, any reason you didn't make this 128 like we did in ESM-2? @farhadrgh, did we profile this and find that it's faster to have the padded vocab even at this tiny size?

ynashed: I did it just to be equivalent with the published architecture. I doubt it makes any difference. If it's faster to pad, then pad

sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
```diff
@@ -0,0 +1,354 @@
...
+
+    # core attention
+    use_esm_attention: bool = False  # Skip ESM2 custom attention for TE acceleration. Still passes golden value test.
+    attention_softmax_in_fp32: bool = False
+    normalize_attention_scores: bool = False
+
+    # From megatron.core.models.gpt.bert_model.GPTModel
+    fp16_lm_cross_entropy: bool = False  # Move the cross entropy unreduced loss calculation for lm head to fp16
+    parallel_output: bool = True
+    share_embeddings_and_output_weights: bool = True
```
pstjohn, line 306: @ynashed, I think this is actually supposed to be `False` for AMPLIFY? At least in their huggingface checkpoint, `encoder.weight` and `decoder.weight` are different tensors

ynashed: I tried with on and off and it did not make any significant difference in the training loss curves, but I agree we should turn it off to match the published model.

sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
pstjohn: Fixed, thanks!

sub-packages/bionemo-amplify/src/bionemo/amplify/model.py
pstjohn: Sounds good

PR #679, opened 2025-02-05
remove development target, install all packages in postcreatecommand

pstjohn:
Restructures the dockerfile setup

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
-for sub in ./3rdparty/*/ ./sub-packages/bionemo-*/; do
-    uv pip install --no-deps --no-build-isolation --editable $sub
-done
+set -eo pipefail
+
+uv pip install --no-build-isolation --editable \
+  ./3rdparty/* \
+  ./sub-packages/bionemo-*
+
+rm -rf /tmp/*
+rm -rf ./sub-packages/bionemo-noodles/target
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN rustup set profile minimal && \
   rustup target add x86_64-unknown-linux-gnu && \
   rustup default 1.82.0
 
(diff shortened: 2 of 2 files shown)


skothenhill-nv (approved):
This deletes interna/scripts. If this is intentional, could you please:

- update the README with alternative instructions for building images and developer environment.
- update the PR description to describe the changes to developer workflow (e.g. instead of using "build_dev_image.sh, we prefer... blah"
- describe the major changes to Dockerfile in a bullet list? I think it would be easier to look at the change and see something like, 1) removed development target, 2) release is built from dev 3) 

Or something else if it makes more sense. Ideally both the PR description and the README would make clear both the changes from this script and how we expect workflows to change.

cspades:
Discussed with @pstjohn, ideally we can preserve our `docker build` development workflow, instead of relying on a release container without editable installs, or `devcontainers` which requires... `devcontainers`.

Implementation is super flexible (e.g. a third build path in the Dockerfile called `oss-dev`), but the requirements if we can do this are:

- To run a development setup with editable installs of all of our sub-packages using a simple `docker run --rm -it -v <code> -v <data> -v <results>` (or similar, etc.) and build the setup with only `docker build --target oss-dev` as in [build_dev_image.sh](https://github.com/NVIDIA/bionemo-framework/blob/main/internal/scripts/build_dev_image.sh).
  - This also provides a clear and simple recipe for our OSS developers to build containers.
- To either setup the ENV correctly, or mount the relevant configs (`.aws` for PBSS, `.ngc` for NGC, `.netrc` for W&B, `.cache` for pip cache installs, `.ssh` for credentials) that allow us to rapidly get up to speed after running the container, which is identical to what devcontainer supports, except with only `docker build`.

Originally, this PR was meant to fix an issue where `bionemo-noodles` had some of its installation dependencies deleted during the build: https://github.com/NVIDIA/bionemo-framework/blob/main/Dockerfile#L178
```
RUN <<EOF
  set -eo pipefail
  rm -rf /usr/local/lib/python3.12/dist-packages/bionemo*
  pip uninstall -y nemo_toolkit megatron_core
EOF
```

PR #701, opened 2025-02-25
disable metric when model parallel

sichu2023:
### Description
Previously metric logging is blocking model parallel. Disable instead of raising error.

### Type of changes


farhadrgh (approved):

dorotat-nv (approved):
Would be great to have it unit tested, but LGTM

PR #726, opened 2025-03-06
remove zstandard to address nvbug 5149698

pstjohn:
Removes `zstandard` after installing NeMo to avoid issues like the following downloading large binary files with ngcsdk. 

https://nvbugspro.nvidia.com/bug/5149698

```
    raise DecodeError(
urllib3.exceptions.DecodeError: ('Received response with content-encoding: zstd, but failed to decode it.', ZstdError('cannot use a decompressobj multiple times'))
...
requests.exceptions.ContentDecodingError: ('Received response with content-encoding: zstd, but failed to decode it.', ZstdError('cannot use a decompressobj multiple times'))
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -111,9 +111,10 @@ uv pip install --no-build-isolation \
   -r /requirements-cve.txt \
   -r /requirements-test.txt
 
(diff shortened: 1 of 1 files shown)


trvachov (approved):
LGTM

Do you plan on attempting to remove lm-eval?

dorotat-nv (approved):

PR #739, opened 2025-03-10
If desired, training can be stopped on a specific step without impacting the LR curve.

jstjohn:
See new option `--early-stop-on-step` which may be used in JET tests to stop on a specific step without impacting the LR curve by changing `--max-steps`

### DIFF
--- a/sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py
+++ b/sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py
@@ -16,10 +16,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
 import os
 import re
+import shlex
 import subprocess
 import sys
+from contextlib import redirect_stderr, redirect_stdout
 
 import pytest
 import torch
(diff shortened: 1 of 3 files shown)


dorotat-nv (approved):
Still, would be nice to add a unit tests with label `slow`

sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py
```diff
@@ -73,6 +74,60 @@ def test_train_evo2_runs(tmp_path, num_steps=5):
     assert result.returncode == 0, "train_evo2 command failed."
 
 
+@pytest.mark.timeout(256)  # Optional: fail if the test takes too long.
+@pytest.mark.slow
+def test_train_evo2_stops(tmp_path, num_steps=500000, early_stop_steps=3):
+    """
+    This test runs the `train_evo2` command with mock data in a temporary directory.
```
dorotat-nv, line 81: if you would use train test definition from  test_train_single_gpu it is a bit quicker since subprocess import the whole codebase and it takes a few sec

cspades (approved):

PR #741, opened 2025-03-10
Geneformer gene embedding calculation

jyin-bst:
### Description

We implemented the gene embedding calculation for BioNeMo Geneformer model. The gene embedding function is essential for the Geneformer model to extract gene embeddings for downstream tasks, such as gene coexpression network generation and in silico perturbation.

Only one file is changed to summarize gene embedding using the hidden state and input ids from the model.
sub-packages/bionemo-llm/src/bionemo/llm/utils/callbacks.py

### Type of changes

- [x]  New feature (non-breaking change which adds functionality)
- [ ]  Refactor
- [ ]  Documentation update

### CI Pipeline Configuration

> [!NOTE]

### Usage

When users call infer_geneformer.py from BioNeMo (sub-packages/bionemo-geneformer/src/bionemo/geneformer/scripts/infer_geneformer.py) with --no-embeddings as False (as default in the current script), the script will write the gene embeddings into the result file.

```python
!infer_geneformer \
    --data-dir {data_dir} \
    --checkpoint-path {geneformer_10m} \
    --results-path {result_path_10m} \
    --micro-batch-size {micro_batch_size} \
    --seq-len 2048 \
    --num-dataset-workers 10 \
    --num-gpus 1 \
    --include-input-ids \
    --no-embeddings False

```

### Pre-submit Checklist


sub-packages/bionemo-llm/src/bionemo/llm/utils/callbacks.py
```diff
@@ -108,6 +108,56 @@ def write_on_epoch_end(
             collate_kwargs["seq_dim_key_defaults"] = self.seq_dim_key_defaults
         prediction = batch_collator([item for item in predictions if item is not None], **collate_kwargs)
 
+        # handle gene embeddings
```
jstjohn, line 111: Hmm so this function is used by a number of different tools where gene embeddings would not make sense. Here is my proposal. In the geneformer branch, make a `utils/callbacks.py` and import this callback and subclass it, adding your stuff the subclass. It will look like the following:

```
class GeneformerPredictionWriter(PredictionWriter):
    def __init__(self, *args, include_gene_embeddings:bool = False, **kwargs):
        super().__init__(*args, **kwargs)
        # now save whether the user wants gene embeddings
        self.include_gene_embeddings = include_gene_embeddings
    def write_on_epoch_end(
        self,
        trainer: pl.Trainer,
        pl_module: pl.LightningModule,
        predictions: Any,
        batch_indices: Sequence[int],
    ) -> None:
    ** copy docstring here**
    # handle gene embeddings
    if self.include_gene_embeddings:
        ... the rest of your new code here
 ```
 
 Then in the geneformer predict function, add a cli option to argparse:
 ```
 parser.add_argument("--include-gene-embeddings", action="store_true", help="Add gene embeddings to predictions")
 ```
 
 And use this new prediction writer, passing that option from the argparser through. 
 
 Let me know if this makes sense! If you do it this way, the current behavior is not changed, but you should have support for your new feature.

jstjohn (changes requested):

sub-packages/bionemo-llm/src/bionemo/llm/utils/callbacks.py
jyin-bst: Thank you John. I will try it and update you later.

jyin-bst:
I will start a new pull request for the gene embedding calculation based on John's suggestions.

PR #751, opened 2025-03-13
bump ruff to 0.9.10 and reformat files as necessary

pstjohn:
Bumps the ruff version we use to format files. Ruff now will check and format jupyter notebooks, which means we need a couple small edits.

The ruff vscode extension has deprecrated the old ruff-lsp, so the old version is giving me a warning every time I open the container.


dorotat-nv (approved):

nvdreidenbach (approved):
moco changes lgtm

jstjohn (approved):

PR #758, opened 2025-03-14
Updated file paths for images in SCDL README

polinabinder1:


sub-packages/bionemo-scdl/README.md
```diff
@@ -136,11 +136,11 @@ convert_h5ad_to_scdl --data-path hdf5s --save-path example_dataset
 
 The runtime and memory usage are examined on a CellXGene Dataset with ~1.5 million rows and a size of 24 GB. On this dataset, there is a 4.9x memory speed up.
 
-<img src="images/throughput.png" alt="Throughput Image" width="600">
+<img src="https://raw.githubusercontent.com/NVIDIA/bionemo-framework/main/sub-packages/bionemo-scdl/images/throughput.png" alt="Throughput Image" width="600">
```
jstjohn, line 139: Is there something like a relative path that would work for all of the ways this could be displayed?
* Github nav render (prob works fine)
* Compiled docs (the github main docs page would be fine, since that points at main like this link)
* Versioned compiled docs (this would not work here since it points to the latest image).

jstjohn (approved):
Ok with this for now, See discussion in slack thread about the need for a better long term solution: https://nvidia.slack.com/archives/C074Z808N05/p1741975424499369

jwilber (approved):
- updated mkdocs render logic by adding get_subpackage_assets function to correctly copy assets over using mkdocs_gen_files
- renamed images/ dir to assets/
- udpated reademe so that the  tags were replaced with markdown image syntax

PR #775, opened 2025-03-21
Add AMPLIFY inference

pstjohn:
Adds a new mock datamodule class and a new infer entrypoint for AMPLIFY. Also moves the `mock_xformers` logic from `test_convert.py` to `convert.py`, so it can be used in the infer script to infer directly from a huggingface tag.


farhadrgh (approved):
LGTM, on a slightly unrelated note, I just faced an issue while using the checkpoints save in the following format to run inference with ESM2:
```python
filename="{epoch}-{val_loss:.2f}-{step}-{consumed_samples}"
```
The problem is due to `{val_loss:.2f}` - Different ranks have slightly different validation loss values due to distributed training, leading to different directories per rank. We can avoid this by switching to the following. This aggregates the rank specific data into the same dir which we use for inference:
```python
filename="checkpoint-{step}-{consumed_samples}"
````

jstjohn (approved):

PR #776, opened 2025-03-21
[cye/subpack-gpu-testing] Add GPU runner to testing job.

cspades:
### Description

- Add GPU runners to BioNeMo Sub-Package CI workflow.
- https://jirasw.nvidia.com/browse/BIONEMO-1340

### Details

- GHA GPU Runner Documentation: https://docs.gha-runners.nvidia.com/runners/#gpu-runners
- NVIDIA CUDA Container Images: https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda/tags
  - Installing CUDA is presumed out-of-scope of BioNeMo Framework.
- BioNeMo Sub-Package Updates - GPU Testing PASS Checklist
  - `bionemo-moco` ✅ @nvdreidenbach 
  - `bionemo-webdatamodule` ✅ @DejunL 
  - `bionemo-size-aware-batching` ✅ @DejunL 
  - `bionemo-core` ✅ 
  - `bionemo-testing` ✅ 
- TODO after TransformerEngine is installed, otherwise they are not functional. (https://jirasw.nvidia.com/browse/BIONEMO-1341)
  - `bionemo-llm` 🚧 
    - `bionemo-amplify`
    - `bionemo-esm2`
    - `bionemo-evo2`
    - `bionemo-geneformer`

### Usage

- Refer to the BioNeMo Sub-Package CI documentation to run the GHA workflow: https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/user-guide/contributing/contributing.md#contributing-python-sub-packages-to-bionemo-framework

### Testing

- Latest: https://github.com/NVIDIA/bionemo-framework/actions/runs/14030539337

### DIFF
--- a/.github/workflows/bionemo-subpackage-ci.yml
+++ b/.github/workflows/bionemo-subpackage-ci.yml
@@ -21,7 +21,7 @@ on:
 
 env:
   # TODO(@cspades): Skip pytest for packages that require GPU's or special installation recipes until resolved.
-  SUBPACKAGE_SKIP_TEST: '["bionemo-noodles","bionemo-geometric","bionemo-llm","bionemo-evo2","bionemo-esm2","bionemo-amplify","bionemo-geneformer","bionemo-testing","bionemo-example_model","bionemo-fw"]'
+  SUBPACKAGE_SKIP_TEST: '["bionemo-noodles","bionemo-geometric","bionemo-evo2","bionemo-esm2","bionemo-amplify","bionemo-geneformer","bionemo-testing","bionemo-example_model","bionemo-fw"]'
   # TODO(@cspades): Skip publishing for packages that require special installation recipes or should not be published to PyPI.
   SUBPACKAGE_SKIP_PUBLISH: '["bionemo-noodles","bionemo-geometric","bionemo-testing","bionemo-example_model"]'
 
@@ -115,7 +115,8 @@ jobs:
         run: |
           apt-get update
           apt-get install -y git
-          apt-get install -y lsb-release
+          apt-get install -y lsb-release      # No longer pre-installed in Ubuntu>=22.04.
+          apt-get install -y build-essential  # For installing C build tools, like GCC and make.
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
--- a/.github/workflows/bionemo-subpackage-ci.yml
+++ b/.github/workflows/bionemo-subpackage-ci.yml
@@ -21,7 +21,6 @@ env:
   SUBPACKAGE_SKIP_TEST: '["bionemo-moco","bionemo-noodles","bionemo-geometric","bionemo-webdatamodule","bionemo-size-aware-batching","bionemo-testing","bionemo-llm","bionemo-evo2","bionemo-esm2","bionemo-geneformer","bionemo-example_model","bionemo-fw"]'
   # TODO(@cspades): Skip publishing for packages that require special installation recipes or should not be published to PyPI.
   SUBPACKAGE_SKIP_PUBLISH: '["bionemo-noodles","bionemo-geometric","bionemo-testing","bionemo-example_model"]'
-  NVIDIA_VISIBLE_DEVICES: "all"
 
 jobs:
   configure-workflow-packages:
@@ -74,6 +73,9 @@ jobs:
 
   install-and-test:
     needs: configure-workflow-packages
+    # Check if the previous job has any staged packages to test and publish.
+    # If not, skip this job and subsequent jobs that directly or indirectly depend on this job.
+    if: ${{ needs.configure-workflow-packages.outputs.workflow_packages != '[]' }}
     outputs:
       test_packages: ${{ needs.configure-workflow-packages.outputs.workflow_packages }}
     strategy:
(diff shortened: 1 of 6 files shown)


.github/workflows/bionemo-subpackage-ci.yml
```diff
@@ -80,7 +81,11 @@ jobs:
         package: ${{ fromJson(needs.configure-workflow-packages.outputs.workflow_packages) }}
       fail-fast: false  # Prevent all matrix jobs from failing if one fails.
     name: "[${{ matrix.package }}] Install and test sub-package."
-    runs-on: ubuntu-latest
+    runs-on: linux-amd64-gpu-l4-latest-1
+    container: # GPU jobs must run in a container
+      image: nvidia/cuda:12.0.0-base-ubuntu22.04
+      env:
+        NVIDIA_VISIBLE_DEVICES: ${{ env.NVIDIA_VISIBLE_DEVICES }} # GPU jobs must set this container env variable
```
pstjohn, line 88: I think that's actually outdated info now

.github/workflows/bionemo-subpackage-ci.yml
```diff
@@ -80,7 +81,11 @@ jobs:
         package: ${{ fromJson(needs.configure-workflow-packages.outputs.workflow_packages) }}
       fail-fast: false  # Prevent all matrix jobs from failing if one fails.
     name: "[${{ matrix.package }}] Install and test sub-package."
-    runs-on: ubuntu-latest
+    runs-on: linux-amd64-gpu-l4-latest-1
+    container: # GPU jobs must run in a container
+      image: nvidia/cuda:12.0.0-base-ubuntu22.04
```
pstjohn, line 86: Do we need to use a cuda base image? if you install pytorch into a standard `ubuntu` container, it brings along it's own cuda libraries. I think the questions are
1. Do we expect to be able to install + run in a container without existing cuda libraries? (i.e., with python installing the cuda bits we need)
2. If we do need a system CUDA library, do we also need a cuda compiler? (i.e., do we need the devel or runtime image)?

cspades: I try it out and report back, this is just copypasta from the GHA docs. We would ideally want to emulate a user who has a fresh ubuntu install, so I'll probably have to make it work.

cspades: I'm currently still using a CUDA base image since it has the NVIDIA GPU drivers already pre-installed.
```
>           torch._C._cuda_init()
E           RuntimeError: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx
```
And it seems to work really well with that. Can we just assume that users will have to figure out how to install GPU drivers, and that the CUDA image is a sufficient approximation of what users will experience downloading and installing hardware drivers?

.github/workflows/bionemo-subpackage-ci.yml
cspades: I think so too, though only when you use their CUDA image. Otherwise, you need to install hardware drivers and set this ENV variable yourself.

Removed, resolved.

sub-packages/bionemo-size-aware-batching/pyproject.toml
```diff
@@ -11,9 +11,7 @@ requires-python = ">=3.10"
 license = {file = "LICENSE"}
 dynamic = ["version"]
 dependencies = [
-    # bionemo sub-packages
```
jstjohn, line 14: I liked these... It's a nice way to let people know how you intended to organize.

jstjohn (approved):
Approved but would prefer not doing the changes to pyproject to remove section suggestion comments from the dependency blocks in pyprojects.

cspades:
> Approved but would prefer not doing the changes to pyproject to remove section suggestion comments from the dependency blocks in pyprojects.

I'll put them back, I deprecated the entire "PR trigger" stuff, it was too fancy and we don't need it anymore!

jomitchellnv (approved):

.github/workflows/bionemo-subpackage-ci.yml
pstjohn: Sure that works, yup, I forgot the GHA runners were sensitive to the base containers used

sub-packages/bionemo-llm/pyproject.toml
```diff
@@ -13,6 +13,7 @@ dynamic = ["version"]
 dependencies = [
     # bionemo sub-packages
     'bionemo-core',
+    'bionemo-testing',
```
pstjohn, line 16: wait... this isn't right. we need some optional requirements here; or just manually install bionemo-testing when we run the test suite? I.e.,

```
[project.optional-dependencies]
test = [
    "bionemo-testing"
]
```

pstjohn (approved):
LGTM after the optional-dependencies bit

sub-packages/bionemo-llm/pyproject.toml
cspades: Added, and now `pip install package[test]` for the testing job. Ran it locally and it just works even if `test` isn't defined, which is convenient.

PR #813, opened 2025-04-09
Have dependabot update our docker base image

pstjohn:
An attempt to get dependabot to trigger PRs to update our pytorch docker base image


dorotat-nv (approved):
that would be exciting

PR #830, opened 2025-04-14
Draft: outline push-to-gitlab-registry step

jwilber:
### Description
This corresponds to ticket https://jirasw.nvidia.com/browse/BIONEMO-642: add pushing images to gitlab registry from github actions.

Currently, it's a draft, as I don't have repo rights to add the required authentication tokens.

Changes:
- Added new `push-to-gitlab-registry` job

With steps to:
- debug info
- login to docker
- checkout repo
- verify docker image tags from prev build step
- check cpu architecture
- once I've verified this works + have authentication, I'll add the `buildx imagetools create` step to publish

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -243,6 +243,7 @@ jobs:
           username: ${{ vars.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+      # todo, get access to repo and set up auth for below login
       # - name: Login to GitLab Registry
       #   uses: docker/login-action@v3
       #   with:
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -223,3 +223,59 @@ jobs:
             echo "All test jobs have completed successfully or been skipped!"
             exit 0
           fi
+
+  push-to-gitlab-registry:
```
pstjohn, line 227: Couldn't we just add a tag to `build-push-action`? Maybe add a separate docker-metadata-action that only runs on push events if we want to get out the sha. I'm not sure we need to separate this out as a separate step

Issue #757, opened 2025-03-14
[BUG] ESM2 training does not resume from checkpoint using train_esm2


dorotat-nv:
### BioNeMo Framework Version

c61ef42b0bff5efc79b3f873c71f39893921f7a9

### Bug Description

The ESM2 training  breaks when it resumes from the checkpoint 

### Steps to Reproduce

1. run first training with num_steps=20 and val_check_interval=10
```
python sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/train_esm2.py --train-cluster-path=/data/train_clusters.parquet --train-database-path=/data/train.db --valid-cluster-path=/data/valid_clusters.parquet --valid-database-path=/data/validation.db --micro-batch-size=16 --num-nodes=1 --num-gpus=1  --limit-val-batches=1  --min-seq-length=1024 --max-seq-length=1024 --num-layers=33 --hidden-size=1280 --num-attention-heads=20 --ffn-hidden-size=5120 --create-tensorboard-logger --wandb-project=fix-tensorboard-logs --val-check-interval 10 --wandb-group main --num-steps=20 --resume-if-exists
```
2. Run again with --num-steps 30
```
python sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/train_esm2.py --train-cluster-path=/data/train_clusters.parquet --train-database-path=/data/train.db --valid-cluster-path=/data/valid_clusters.parquet --valid-database-path=/data/validation.db --micro-batch-size=16 --num-nodes=1 --num-gpus=1 --limit-val-batches=1  --min-seq-length=1024 --max-seq-length=1024 --num-layers=33 --hidden-size=1280 --num-attention-heads=20 --ffn-hidden-size=5120 --create-tensorboard-logger --wandb-project=fix-tensorboard-logs --val-check-interval 10 --wandb-group main --num-steps=30 --resume-if-exists
```

### Error Messages and Logs

```shell
[...]

PR #842, opened 2025-04-23
fixing the ESM2 checkpointing issue

polinabinder1:
### Description
This addresses: https://github.com/NVIDIA/bionemo-framework/issues/757
### Type of changes

In the original code the optimizer was not saved in the checkpoint, but it is expected in the megatron strategy. Saving the optimizer is added to the checkpoint callback and the test has been updated. Changes were made to have the checkpointing callback in the nemo logger. This is the same as the training path in sub-packages/bionemo-llm/src/bionemo/llm/train.py.


sub-packages/bionemo-esm2/tests/bionemo/esm2/scripts/test_train_esm2.py
```diff
@@ -403,53 +398,25 @@ def test_main_stop_at_num_steps_and_continue(tmp_path, dummy_protein_dataset, du
...
         experiment_name=experiment_name,
     )
-    stdout_second_run, stderr_second_run, returncode_second_run = run_command_in_subprocess(
-        command=command_second_run, path=str(tmp_path)
-    )
+    stdout_second_run = run_command_in_subprocess(command=command_second_run, path=str(tmp_path))
 
-    # Verify that the command failed with a non-zero exit code
-    # This assertion will fail if the resume functionality gets fixed, prompting us to update the test
-    assert returncode_second_run != 0, (
```
trvachov, line 412: Should we keep a return code check?

trvachov (commented):
@jwilber can you review as well please?

skothenhill-nv:
There is similar logic here: https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-llm/src/bionemo/llm/train.py#L59

for the pydantic based workflow. but checkpoints in this code-path are passed into `setup_nemo_lightning_logger` but without the additional arguments to the callback- seems like we do not see the same error.

Do you understand what's going on here? What are the tradeoffs when passing the ModelCheckpoint callback via `Trainer` vs `setup_nemo_lightning_logger`. It makes me uncomfortable that we have hit a divergent point.

skothenhill-nv (approved):
Approved- but please address the divergence with sub-packages/bionemo-llm/src/bionemo/llm/train.py .

Ideally all of our models would use the same workflow. We should understand the difference and if its a reason we saw this bug.

Or maybe we should document in the description of this PR the differences two code paths?

sub-packages/bionemo-esm2/tests/bionemo/esm2/scripts/test_train_esm2.py
polinabinder1: We look at stdout_second_run later (line 405)

sub-packages/bionemo-esm2/tests/bionemo/esm2/scripts/test_train_esm2.py
```diff
@@ -185,6 +185,7 @@ def get_command_line_args(arg_name, arg_value) -> str:
         return arg_str
 
     cmd = "train_esm2 " + " ".join(get_command_line_args(arg_name, arg_value) for arg_name, arg_value in args.items())
+    print("CMD", cmd)
```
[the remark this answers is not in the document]
polinabinder1: yep, will remove it

trvachov:
Nit: can you fix the type in the PR title?

sub-packages/bionemo-esm2/tests/bionemo/esm2/scripts/test_train_esm2.py
```diff
@@ -185,6 +185,7 @@ def get_command_line_args(arg_name, arg_value) -> str:
         return arg_str
 
     cmd = "train_esm2 " + " ".join(get_command_line_args(arg_name, arg_value) for arg_name, arg_value in args.items())
```
jwilber, line 187: Looks like we're using f-strings everywhere else, so I'd make this

```
cmd = f"train_esm2 {' '.join(get_command_line_args(arg_name, arg_value) for arg_name, arg_value in args.items())}"
```

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/train_esm2.py
```diff
@@ -339,8 +316,20 @@ def main(
             filename="{epoch}-{step}-{consumed_samples}",
```
jwilber, line 316: This isn't part of your changes, but reviewing the file, I think `filename` here is missing an f-string prefix?

jwilber: Oh nevermind, ptl doing some magic under the hood, ignore!: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html

jwilber (approved):
Added style comment, otherwise LGTM

PR #848, opened 2025-04-25
inference ddp improvements

skothenhill-nv:
Adds a constraint to the inference constraint that limits 'epoch' writing to 1 device. This is due to a non-deterministic ordering that occurs when the number of devices is > 1. In the case of 'batch' frequency, the batch index is recorded and thus the original ordering may be recovered.


skothenhill-nv:
wrong base branch.

PR #862, opened 2025-05-07
remove unused dependencies from bionemo-core

pstjohn:
Removes unused dependencies from bionemo-core's pyproject.toml so it's a lighter-weight install via pip


sub-packages/bionemo-core/pyproject.toml
```diff
@@ -17,15 +17,12 @@ dependencies = [
...
     "torch>=2.2.1",
-    'boto3',
-    'lightning>=2.2.1',
     'nest_asyncio',
     'ngcsdk',
     'pooch',
     'pydantic[email]>=2.7.0',
     'pyyaml',
     'tqdm',
-    'uv',
```
cspades, line 28: Just some historical context, I added this here mainly because I think I have some documentation in one or more of the sub-package README.md that taught you how to use `uv` to pip install TE or something as a post-install thing without needing to install the base dependencies, which is something only supported in `uv` and not `pip`.

We can remove it, but that means to use that documentation, they have to manually install `uv`, and then install TE without re-installing everything else.

cspades (approved):

sub-packages/bionemo-core/pyproject.toml
pstjohn: we should put that in the bionemo image then, right? not here; i just want to `load(esm2/testdata)`

jstjohn (approved):

PR #867, opened 2025-05-08
Exposes TflopsCallback to pydantic workflow

skothenhill-nv:
training_config now has an additional property -- include_tflops_callback. This matches the workflow in the `train_geneformer` and `train_esm2` entrypoints. This parameter may also be set via CLI when submitting a config. Additionally exposes the option to disable the use of distributed optimizers. this allows for use of non-adam optimizers such as SGD.


skothenhill-nv:
still uses the wrong base branch.

PR #872, opened 2025-05-12
adding .obs to scdl

polinabinder1:


Intron7:
most of obs is either numerical or categorical. wouldnt it make for sense to just save the codes of the categorical data and than store the labels once?

sub-packages/bionemo-scdl/src/bionemo/scdl/io/single_cell_memmap_dataset.py
```diff
@@ -523,6 +522,27 @@ def regular_load_h5ad(
...
+                        mode=Mode.CREATE_APPEND,
+                    )
+
+                else:
+                    self.obs[obs_name] = np.memmap(
+                        f"{str(Path(self.data_path).absolute())}/obs/{obs_name}",
+                        dtype=adata.obs[obs_name].dtype,
+                        shape=(num_rows + 1,),
+                        mode=Mode.CREATE_APPEND,
+                    )
```
jstjohn, line 543:
```suggestion
                if adata.obs[obs_name].dtype == "category":
                    obs_dtype = "S20"
                else:
                    obs_dtype = adata.obs[obs_name].dtype
                self.obs[obs_name] = np.memmap(
                        f"{str(Path(self.data_path).absolute())}/obs/{obs_name}",
                        dtype=obs_dtype,
                        shape=(num_rows + 1,),
                        mode=Mode.CREATE_APPEND,
                    )
```
Maybe somethign like the above since it looks like the commands are the same other than setting the dtype?

jstjohn:
Can you add a test that demonstrates some obs extraction? Maybe something like cell type?

Also is there any chance that there are conflicting integer categorical labels between anndata files for the same string value? I don't think the assignment of categorical label to string is deterministic as a function of a particular string. I think it at least depends on (in addition) the order of other strings observed? I don't _think_ it does something like a hash or a sort as part of the logic of assigning category ids.

Hmm... but maybe you are dealing with that by saving the string reps and not the category ids (eg you use S20 which should be a 20 byte long string)

Note that it's also possible that the above concern is a real issue when you get 3 anndata files from 3 sequencing runs, but not when you query anndata files from a giant global TileDB object like you do through cellxgene. Maybe the cellxgene derived data already shares the same categorical labels but some random set of 3 anndata files made by different labs would not, even if a standard vocabulary were used.

malcolmgreaves (approved):

PR #900, opened 2025-05-29
Fix masked token loss reductions

skothenhill-nv:
### Description
Fixes a bug introduced by a reworking of masked_token_loss where losses were reduced across worker groups while tokens were not. This solution does as global all reduce over both tokens and losses, and then does the per-token calculation on a single worker. This is slightly different than the previous implementations, where losses per token were computed in each data paralell/context parallel group, and then were averaged. This implementation matches how it is performed in the refactored BERTMLMLoss in NeMo.

Next step should be to match the pattern and structure used by NeMo, where the reduction now actually occurs in the `reduce` method.

### Type of changes

- [ ]  Refactor
- [ ]  Documentation update

### Pre-submit Checklist

 - [x] All existing tests pass successfully


sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
```diff
@@ -155,6 +153,7 @@ def __init__(
         super().__init__()
         self.validation_step = validation_step
         self.val_drop_last = val_drop_last
+        self.rank = parallel_state.get_data_parallel_rank()
```
jstjohn, line 156: Maybe get rid of this? doesn't look like it's used or logged anywhere, and makes initialization a little harder since you need to have distributed initialized afaik.

jstjohn: You could add an `@property` `def rank(self) -> int` method that does the same thing as this lazily if anyone wants it, and you want it for debugging etc.

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
```diff
@@ -174,43 +173,42 @@ def forward(
...
 
-        # Get the context parallel size for some normalizations and reductions.
-        cp_size = parallel_state.get_context_parallel_world_size()
+        # We can compute the loss for the group by dividing the summed loss by the number of valid tokens in the group.
+        loss_for_microbatch = group_token_loss_sum / num_valid_tokens_in_microbatch
 
         # If we do not drop the last partial batch of validation, we need to do fancy reduction handling to support
         #  reducing the loss across the data parallel group.
         if self.validation_step and not self.val_drop_last:
-            if loss_for_microbatch.isnan():
```
jstjohn, line 195: I don't think you need this anymore with the new formulation. Follow the pattern of https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/megatron_parallel.py#L1733-L1788 which should work for either the drop last and regular case? I think that's part of the appeal.

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
```diff
@@ -219,21 +217,29 @@ def forward(
...
+                ]
+            )
+            # Reduce over all groups
+            torch.distributed.all_reduce(
+                loss_sum_and_microbatch_size_all_gpu,
+                group=parallel_state.get_data_parallel_group(with_context_parallel=True),
+                op=torch.distributed.ReduceOp.SUM,
+            )
+            # Compute the reduced loss over all groups
+            reduced_loss = loss_sum_and_microbatch_size_all_gpu[0] / loss_sum_and_microbatch_size_all_gpu[1]
```
jstjohn, line 236: This is a confusing flow to me now.

Looks like:
1) if we are doing validation and there is drop last then we do not do reduction in forward and let the reduction happen elsewhere
2) if there is drop last, then we _do_ the reduction in forward and return avg which is used downstream.

AFAIK we should just rewrite this whole thing to be more like the new NeMo flow which is a lot simpler with fewer if/else conditionals. We should pick one place to put the reductions and not reduce in forward one way and in reduce anotherway. IMO that's prob why we have so many bugs now https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/megatron_parallel.py#L1733-L1788.

cspades: This exact thing confused the heck out of me too when I was fixing it the first time!

I was like WTF is going on here?!

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
skothenhill-nv: fixed

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
skothenhill-nv: done

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
```diff
@@ -219,21 +217,29 @@ def forward(
                 op=torch.distributed.ReduceOp.SUM,
             )
 
-            # Return the loss tensor multiplied by the context parallel size,
-            # and the data & context parallel reduced loss sum.
-            return loss_for_microbatch * cp_size, {
-                "loss_sum_and_microbatch_size": loss_sum_and_microbatch_size_all_gpu
-            }
+            return val_loss_for_microbatch, {"loss_sum_and_microbatch_size": loss_sum_and_microbatch_size_all_gpu}
```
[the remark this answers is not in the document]
skothenhill-nv: our reduce does not do the all reduce, our forward does, but i modified to match nemo as discussed.

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
skothenhill-nv: just to pile on, this also confused me, i couldnt find a functional difference between the two. The new apporach seems much cleaner.

jstjohn (approved):

cspades (approved):

trvachov:
🍾

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
```diff
@@ -62,178 +60,77 @@ class DataParallelGroupLossAndIO(TypedDict):
...
+                else torch.tensor([0.0, 0.0], device=torch.cuda.current_device())
+            )
             torch.distributed.all_reduce(
-                loss_sum_and_microbatch_size_all_gpu,
+                loss,
                 group=parallel_state.get_data_parallel_group(with_context_parallel=True),
-                op=torch.distributed.ReduceOp.SUM,
             )
+            # average over the total number of tokens across the global batch.
+            loss = loss[0] / loss[1]
```
trvachov, line 129: suspect? div by 0?

sub-packages/bionemo-llm/src/bionemo/llm/model/loss.py
```diff
@@ -62,178 +60,77 @@ class DataParallelGroupLossAndIO(TypedDict):
...
+
+        if self.validation_step and not self.val_drop_last and loss_sum.isnan():
+            assert num_valid_tokens == 0, "Got NaN loss with non-empty input"
+            if batch["loss_mask"].count_nonzero() != 0:
+                raise ValueError("Got NaN loss with non-empty input")
+            loss_sum = torch.zeros_like(num_valid_tokens)
+
+        num_valid_tokens = num_valid_tokens.clone().detach().to(torch.int)
+        loss_sum_and_ub_size = torch.cat([loss_sum.clone().detach().view(1), num_valid_tokens.view(1)])
+        return loss_sum, num_valid_tokens, {"loss_sum_and_ub_size": loss_sum_and_ub_size}
```
trvachov, line 100: forward func return: loss_sum=tensor(0., device='cuda:1', grad_fn=<SumBackward0>),num_valid_tokens=tensor(0, device='cuda:1', dtype=torch.int32),loss_sum_and_ub_size=tensor([0., 0.], device='cuda:1')

num_valid_tokens=0 is possible here.

PR #922, opened 2025-06-04
Add cli interface for esm2 checkpoint conversion

pstjohn:
Follow-up to #918 with some minor improvements, additional testing, and the addition of a typer CLI command to initiate checkpoint conversions

You can now convert checkpoints to huggingface with
```
convert_esm2 convert-nemo-to-hf /path/to/existing/nemo/checkpoint /path/to/new/hf/checkpoint
```

### DIFF
--- a/sub-packages/bionemo-esm2/src/bionemo/esm2/model/convert.py
+++ b/sub-packages/bionemo-esm2/src/bionemo/esm2/model/convert.py
@@ -18,7 +18,8 @@ from pathlib import Path
 
 import torch
 import typer
-from nemo.lightning import io, teardown
+from megatron.core.dist_checkpointing.validation import StrictHandling
+from nemo.lightning import MegatronStrategy, Trainer, io, teardown
 from nemo.lightning.pytorch.utils import dtype_from_hf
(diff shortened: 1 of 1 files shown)


sub-packages/bionemo-esm2/src/bionemo/esm2/model/convert.py
```diff
@@ -340,3 +343,41 @@ def _import_qkv_bias(ctx: io.TransformCTX, query, key, value):
     concat_biases = concat_biases.transpose(0, 1).contiguous()
     concat_biases = concat_biases.view(*input_shape)
     return concat_biases
+
+
+app = typer.Typer(pretty_exceptions_enable=False)
```
jstjohn, line 348: wow, this is magic. Can you add a comment above this saying something like "# create command line interface automatically with Typer"

jstjohn (approved):

sub-packages/bionemo-esm2/src/bionemo/esm2/model/convert.py
pstjohn: we use this elsewhere too 😁

PR #959, opened 2025-06-26
Fix esm2 finetune loss

yzhang123:
### Description
fixed esm2 finetuning loss for token classification, regression, and sequence classification to work with data parallelism
and classification to work with vocab tp parallelism

updated fine-tuning notebook to have test evaluation

### Type of changes

- [ ]  Refactor
- [ ]  Documentation update

### CI Pipeline Configuration

> [!NOTE]

#### Authorizing CI Runs

* If a pull request is opened by a trusted user and contains only trusted changes, the pull request's code will
* If a pull request is opened by an untrusted user or contains untrusted changes, an NVIDIA org member must leave an

### Usage

```python
TODO: Add code snippet
```

### Pre-submit Checklist

### DIFF
--- a/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
+++ b/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
@@ -23,8 +23,6 @@ from torch import Tensor
 
 from bionemo.llm.model.loss import (
     BERTMLMLossWithReduction,
-    PerTokenLossDict,
-    SameSizeLossDict,
     unreduced_token_loss_fn,
 )
 
@@ -43,42 +41,26 @@ class RegressorLossReduction(BERTMLMLossWithReduction):
 
     def forward(
         self, batch: Dict[str, Tensor], forward_out: Dict[str, Tensor]
-    ) -> Tuple[Tensor, PerTokenLossDict | SameSizeLossDict]:
-        """Calculates the loss within a micro-batch. A micro-batch is a batch of data on a single GPU.
+    ) -> Tuple[torch.Tensor, torch.Tensor, Dict[str, torch.Tensor]]:
+        """Calculates the sum of squared errors within a micro-batch. A micro-batch is a batch of data on a single GPU. The averaging of the loss is done in super().reduce().
 
         Args:
             batch: A batch of data that gets passed to the original forward inside LitAutoEncoder.
             forward_out: the output of the forward method inside classification head.
-
-        Returns:
-            A tuple containing [<loss_tensor>, ReductionT] where the loss tensor will be used for
-                backpropagation and the ReductionT will be passed to the reduce method
-                (which currently only works for logging.).
         """
         regression_output = forward_out["regression_output"]
         targets = batch["labels"].to(dtype=regression_output.dtype)  # [b, 1]
 
+        num_valid_tokens = torch.tensor(targets.numel(), dtype=torch.int, device=targets.device)
+
         cp_size = parallel_state.get_context_parallel_world_size()
         if cp_size == 1:
-            loss = torch.nn.functional.mse_loss(regression_output, targets)
+            loss_sum = ((regression_output - targets) ** 2).sum()  # [b, 1]
         else:
             raise NotImplementedError("Context Parallel support is not implemented for this loss")
 
-        return loss, {"avg": loss}
-
-    def reduce(self, losses_reduced_per_micro_batch: Sequence[SameSizeLossDict]) -> Tensor:
-        """Works across micro-batches. (data on single gpu).
-
-        Note: This currently only works for logging and this loss will not be used for backpropagation.
-
-        Args:
-            losses_reduced_per_micro_batch: a list of the outputs of forward
-
-        Returns:
-            A tensor that is the mean of the losses. (used for logging).
-        """
-        losses = torch.stack([loss["avg"] for loss in losses_reduced_per_micro_batch])
-        return losses.mean()
+        loss_sum_and_ub_size = torch.cat([loss_sum.clone().detach().view(1), num_valid_tokens.view(1)])
+        return loss_sum, num_valid_tokens, {"loss_sum_and_ub_size": loss_sum_and_ub_size}
 
 
 class ClassifierLossReduction(BERTMLMLossWithReduction):
@@ -90,15 +72,11 @@ class ClassifierLossReduction(BERTMLMLossWithReduction):
     def forward(
         self, batch: Dict[str, Tensor], forward_out: Dict[str, Tensor]
     ) -> Tuple[torch.Tensor, torch.Tensor, Dict[str, torch.Tensor]]:
-        """Calculates the loss within a micro-batch. A micro-batch is a batch of data on a single GPU.
+        """Calculates the loss within a micro-batch. A micro-batch is a batch of data on a single GPU. The averaging of the loss is done in super().reduce().
 
         Args:
             batch: A batch of data that gets passed to the original forward inside LitAutoEncoder.
             forward_out: the output of the forward method inside classification head.
-
-        Returns:
-            A tuple where the loss tensor will be used for backpropagation and the dict will be passed to
-            the reduce method, which currently only works for logging.
         """
         targets = batch["labels"].squeeze()  # [b] or [b, s] for sequence-level or token-level classification
         loss_mask = batch["loss_mask"]
--- a/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
+++ b/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
@@ -18,9 +18,15 @@ from typing import Dict, Sequence, Tuple
 
 import torch
 from megatron.core import parallel_state
+from nemo.lightning.megatron_parallel import masked_token_loss
 from torch import Tensor
 
-from bionemo.llm.model.loss import BERTMLMLossWithReduction, PerTokenLossDict, SameSizeLossDict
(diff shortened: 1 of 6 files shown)


sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
```diff
@@ -83,50 +71,37 @@ class ClassifierLossReduction(BERTMLMLossWithReduction):
...
+            classification_output = classification_output.unsqueeze(0)  # change to [1, b, num_class]
+            targets = targets.unsqueeze(1)  # change to [b, 1]
+            loss_mask = torch.ones((targets.shape[0], 1), dtype=loss_mask.dtype, device=loss_mask.device)
         else:
-            raise NotImplementedError("Context Parallel support is not implemented for this loss")
+            raise ValueError(f"Unexpected classification output dimension: {classification_output.dim()}")
 
-        return loss, {"avg": loss}
+        # NOTE: token_logits is [sequence, batch] but labels and other fields, including the loss are [batch, sequence]
+        unreduced_token_loss = unreduced_token_loss_fn(classification_output, targets)  # [b s]
```
jstjohn, line 96: Good catch that this is already equivalent to F.cross_entropy, just with a different classification "vocabulary".

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
```diff
@@ -37,42 +41,26 @@ class RegressorLossReduction(BERTMLMLossWithReduction):
 
     def forward(
         self, batch: Dict[str, Tensor], forward_out: Dict[str, Tensor]
-    ) -> Tuple[Tensor, PerTokenLossDict | SameSizeLossDict]:
-        """Calculates the loss within a micro-batch. A micro-batch is a batch of data on a single GPU.
+    ) -> Tuple[torch.Tensor, torch.Tensor, Dict[str, torch.Tensor]]:
+        """Calculates the sum of squared errors within a micro-batch. A micro-batch is a batch of data on a single GPU. The averaging of the loss is done in super().reduce().
```
jstjohn: """Calculates the mean squared error across micro-batches when paired with the megatron scheduler. A micro-batch is a batch of data on a single GPU. The averaging of the loss for backprop happens in the megatron scheduler while the averaging for logging is handled in super().reduce().

jstjohn: See https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/pipeline_parallel/schedules.py#L303-L308 where your "loss_sum" is divided by "num_valid_tokens" which leads to the MSE. This is also where the backwards happens prior to reduction.

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
```diff
@@ -37,42 +41,26 @@ class RegressorLossReduction(BERTMLMLossWithReduction):
...
         """
         regression_output = forward_out["regression_output"]
         targets = batch["labels"].to(dtype=regression_output.dtype)  # [b, 1]
 
+        num_valid_tokens = torch.tensor(targets.numel(), dtype=torch.int, device=targets.device)
+
         cp_size = parallel_state.get_context_parallel_world_size()
         if cp_size == 1:
-            loss = torch.nn.functional.mse_loss(regression_output, targets)
+            loss_sum = ((regression_output - targets) ** 2).sum()  # [b, 1]
```
jstjohn: Are we sure by this point that the final linear layer is not column parallel? I assume probably not since the final dim is something like `[b, s, 1]` prior to this sum reduction, right? Also see my suggestion below, as is you are also reducing over the batch with the default `.sum()`
```suggestion
            loss_sum = ((regression_output - targets) ** 2).sum(dim=-1)  # [b, 1]
```

jstjohn: My question about column parallel is related to whether or not this loss is valid for tensor-parallel models? The cross entropy function you use in the classification loss _is_ TP safe, so that should work, but the F.cross_entropy that used to be there was not TP safe. This may still be TP unsafe, unless we know for sure that the different TP ranks are all getting the same linear layer result. We may need to do something like te.Linear and set `weight.sequence_parallel=True` or something so it gets all reduced across TP ranks... There's hackery like this inside of the Evo2 custom layers if we want to verify.

jstjohn: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/gpt/model/megatron/hyena/hyena_utils.py#L507-L516 setting this magic parameter to `True` tells Megatron that the parameter _should not_ be reduced, so... maybe by default things _will_ be reduced?

yzhang123: the last layer is NOT using TP , see https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/sequence_model.py#L42. The idea is that is that the head is a very small MLP layer. 

The code is correct, the comment was wrong. i fixed it. it should be of size [1,1] since the loss_sum should be float

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
yzhang123: Is this a request for change ? Since the goal is to eventually calculate MSE loss, which is broken down into sum of squared errors in forward and the reduction+averaging in super().reduce(), see https://github.com/NVIDIA/NeMo/blob/6a78ab8cf0c67f4d704a245e2b4243dfd968019b/nemo/lightning/megatron_parallel.py#L362

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
yzhang123: actually i just debugged on the minimal example and it looks correct how it is right now, when i use TP=2 the loss is the same as TP=1 across multiple steps, and when i use DP=2 the loss is 1/2 as DP=1. This also means loss for logging stays the same for all

jstjohn (approved):

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
jstjohn: Sounds good, let's keep our eyes on the slack thread and if it turns out it is subtly different, or keeps things from drifting better, then we can go back to it. I've seen things where grads are _slightly_ different between ranks if they are not explicitly mean reduced, and that may not be happening now. It may just work for now because they start the same and run on the same data.

jstjohn: https://nvidia.slack.com/archives/C0434FDLPQV/p1751059626701659 or https://nvidia.slack.com/archives/C01JPHFK7DH/p1751054240388109

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
trvachov: @jstjohn

pstjohn (approved):

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
```diff
@@ -37,42 +41,26 @@ class RegressorLossReduction(BERTMLMLossWithReduction):
...
-        Returns:
-            A tuple containing [<loss_tensor>, ReductionT] where the loss tensor will be used for
-                backpropagation and the ReductionT will be passed to the reduce method
-                (which currently only works for logging.).
         """
         regression_output = forward_out["regression_output"]
         targets = batch["labels"].to(dtype=regression_output.dtype)  # [b, 1]
 
+        num_valid_tokens = torch.tensor(targets.numel(), dtype=torch.int, device=targets.device)
+
```
skothenhill-nv, line 55: is there a per-token regression target w/ masked tokens? If so we should add a guard/conditional case in the situation `num_valid_tokens` is 0.

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
```diff
@@ -83,50 +71,37 @@ class ClassifierLossReduction(BERTMLMLossWithReduction):
...
-                masked_loss = torch.masked_select(losses, loss_mask)
-                loss = masked_loss.sum() / loss_mask.sum()
-            else:
-                loss = losses.mean()  # sequence-level single value classification
+            classification_output = classification_output.permute(1, 0, 2).contiguous()  # change to [s, b, num_class]
+        elif classification_output.dim() == 2:
+            # NOTE: this is for sequence-level classification, we artificially create a sequence dimension to use the same code path as token-level classification
+            classification_output = classification_output.unsqueeze(0)  # change to [1, b, num_class]
+            targets = targets.unsqueeze(1)  # change to [b, 1]
+            loss_mask = torch.ones((targets.shape[0], 1), dtype=loss_mask.dtype, device=loss_mask.device)
```
skothenhill-nv, line 91: nice

sub-packages/bionemo-esm2/src/bionemo/esm2/model/finetune/loss.py
yzhang123: no there is not

PR #621, opened 2025-01-17
set min seq len by default

pstjohn:
### Description
In https://nvbugspro.nvidia.com/bug/5060664 they notice a warning message about performance when pretraining with variable sequence lengths. This is largely an oversight since our test scripts didn't set both minimum and maximum seq_lens. We should have the default if min_seq_length is omitted be to just pad to the maximum sequence length for performance reasons.

### Type of changes

- [x]  Bug fix (non-breaking change which fixes an issue)
- [ ]  New feature (non-breaking change which adds functionality)
- [ ]  Refactor
- [ ]  Documentation update
- [ ]  Other (please describe):

### CI Pipeline Configuration
Configure CI behavior by applying the relevant labels:

- [SKIP_CI](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/user-guide/contributing/contributing.md#skip_ci) - Skip all continuous integration tests
- [INCLUDE_NOTEBOOKS_TESTS](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/user-guide/contributing/contributing.md#include_notebooks_tests) - Execute notebook validation tests in pytest

> [!NOTE]
> By default, the notebooks validation tests are skipped unless explicitly enabled.

### Usage

```python
TODO: Add code snippet
```

### Pre-submit Checklist

 - [x] I have tested these changes locally
 - [x] I have updated the documentation accordingly
 - [x] I have added/updated tests as needed
 - [x] All existing tests pass successfully

PR #964, opened 2025-07-01
[DRAFT] Add ESM2 Finetuning Benchmark Configuration

nvmvle:
### Description

**⚠️ This is a DRAFT PR - Work in Progress**

This PR adds a new YAML configuration file for benchmarking ESM2 finetuning across three different task types: sequence classification, token classification, and sequence regression.

**Current Status:**
- ✅ Basic YAML structure created following existing patterns
- ✅ Three finetuning variants configured (seq_classification, token_classification, seq_regression)
- ✅ Data copying to local storage implemented
- ✅ Dynamic command generation based on task type
- 🚧 Step counts reduced to 100 for testing (original values commented out)
- 🚧 Test assertions temporarily removed for initial validation
- 🚧 Need to validate data paths and checkpoint locations

### Type of changes

- [ ] Refactor
- [ ] Documentation update

### CI Pipeline Configuration

> [!NOTE]

#### Authorizing CI Runs

* If a pull request is opened by a trusted user and contains only trusted changes, the pull request's code will
* If a pull request is opened by an untrusted user or contains untrusted changes, an NVIDIA org member must leave an

### Usage

```bash
# The benchmark will be triggered through the CI pipeline
# Currently configured for quick validation runs (100 steps)
# Production values are (maybe):
# - Sequence Classification: 1485 steps
# - Token Classification: 1500 steps  
# - Sequence Regression: 6980 steps
```

### TODO Before Merging
- [ ] Restore original step counts after validation
- [ ] Add test assertions back with appropriate thresholds
- [ ] Verify data paths exist on CI infrastructure
- [ ] Confirm checkpoint path accessibility
- [ ] Run full benchmark to validate expected metrics

### Pre-submit Checklist

### Notes for Reviewers
- Based on validated training commands from the ESM2 finetuning work
- Time limit increased to 14400s (4 hours) to accommodate full training runs
- Currently using reduced steps (100) for quick iteration and testing
- Test assertions have been temporarily removed to allow initial validation runs
- Please review the command structure and parameter mapping for correctness

### DIFF
--- a/ci/benchmarks/partial-conv/esm2_finetuning.yaml
+++ b/ci/benchmarks/partial-conv/esm2_finetuning.yaml
@@ -158,24 +158,3 @@ script: |-
 
   # Execute the command
   eval $CMD
-tests:
-  - logic_type: dynamic
-    logic_spec:
-      exit_codes:
-        - 0
-      baselines:
-        # For sequence classification
-        - variant: seq_classification
-          val_acc:
-            operator: gt
-            value: 0.77
-        # For token classification
-        - variant: token_classification
-          val_acc:
-            operator: gt
-            value: 0.86
-        # For sequence regression
-        - variant: seq_regression
-          val_mse:
-            operator: lt
-            value: 50.0
--- a/ci/benchmarks/partial-conv/esm2_finetuning.yaml
+++ b/ci/benchmarks/partial-conv/esm2_finetuning.yaml
@@ -1,5 +1,5 @@
 scope: partial-conv
-time_limit: 7200
+time_limit: 14400
 key_segments:
   # Modify keys to be renamed (str) or excluded (False) from run identifier. By default, all args under script_args are included.
   train_data_path: False
@@ -36,7 +36,8 @@ script_args:
       mlp_ft_dropout: 0.25
       mlp_hidden_size: 256
       mlp_target_size: 10
-      num_steps: 1485
+      # num_steps: 1485
+      num_steps: 100
       experiment_name: seq-level-classification
       lr: 0.0005
       batch_size: 64
@@ -54,7 +55,8 @@ script_args:
       cnn_dropout: 0.25
       cnn_hidden_size: 32
       cnn_num_classes: 3
-      num_steps: 1500
+      # num_steps: 1500
+      num_steps: 100
       experiment_name: token-level-classification
       lr: 0.0005
       batch_size: 64
@@ -72,7 +74,8 @@ script_args:
       mlp_ft_dropout: 0.1
       mlp_hidden_size: 128
       mlp_target_size: 1
-      num_steps: 6980
+      # num_steps: 6980
+      num_steps: 100
       experiment_name: seq-level-regression
       lr: 0.0005
       batch_size: 32
--- a/ci/benchmarks/partial-conv/esm2_finetuning.yaml
+++ b/ci/benchmarks/partial-conv/esm2_finetuning.yaml
@@ -1,0 +1,181 @@
+scope: partial-conv
+time_limit: 7200
+key_segments:
+  # Modify keys to be renamed (str) or excluded (False) from run identifier. By default, all args under script_args are included.
+  train_data_path: False
+  valid_data_path: False
+  checkpoint_path: False
+  num_workers: False
+  limit_val_batches: False
+  limit_test_batches: False
+  val_check_interval: False
+  log_every_n_steps: False
+script_args:
+  # All arguments referenced in the script string must be specified here.
+  # Arguments not referenced in the script string must have the 'arg' field specified.
+  # See jet/core/configs.py for the specification of the configuration class
+  workspace: /workspace/bionemo2
+  checkpoint_path: /data/checkpoints/esm2nv650m_v2.0
+  data_base_path: /data/FLIP
+  nodes: [1]
+  gpus: 1
+  precision: [bf16-mixed]
+  encoder_frozen: true
+  num_workers: 8
+  limit_val_batches: 1000
+  limit_test_batches: 1000
+  log_every_n_steps: 1
+  accumulate_grad_batches: 2
+  products:
+    - variant: seq_classification
+      train_data_path: scl/train/x000.csv
+      valid_data_path: scl/val/x000.csv
+      task_type: classification
+      config_class: ESM2FineTuneSeqConfig
+      dataset_class: InMemorySingleValueDataset
+      mlp_ft_dropout: 0.25
+      mlp_hidden_size: 256
+      mlp_target_size: 10
+      num_steps: 1485
+      experiment_name: seq-level-classification
+      lr: 0.0005
+      batch_size: 64
+      label_column: scl_label
+      val_check_interval: 100
+      min_seq_length: null
+      labels_mask_column: null
+      expected_val_acc: 77
+    - variant: token_classification
+      train_data_path: secondary_structure/train/x000.csv
+      valid_data_path: secondary_structure/val/x000.csv
+      task_type: classification
+      config_class: ESM2FineTuneTokenConfig
+      dataset_class: InMemoryPerTokenValueDataset
+      cnn_dropout: 0.25
+      cnn_hidden_size: 32
+      cnn_num_classes: 3
+      num_steps: 1500
+      experiment_name: token-level-classification
+      lr: 0.0005
+      batch_size: 64
+      label_column: 3state
+      val_check_interval: 100
+      min_seq_length: 1024
+      labels_mask_column: resolved
+      expected_val_acc: 86
+    - variant: seq_regression
+      train_data_path: meltome/train/x000.csv
+      valid_data_path: meltome/val/x000.csv
+      task_type: regression
+      config_class: ESM2FineTuneSeqConfig
+      dataset_class: InMemorySingleValueDataset
+      mlp_ft_dropout: 0.1
+      mlp_hidden_size: 128
+      mlp_target_size: 1
+      num_steps: 6980
+      experiment_name: seq-level-regression
+      lr: 0.0005
+      batch_size: 32
+      label_column: target
+      val_check_interval: 100
+      min_seq_length: null
+      labels_mask_column: null
+      expected_val_mse: 50
+script: |-
+  # Set up paths
+  FULL_TRAIN_PATH="${data_base_path}/${train_data_path}"
+  FULL_VALID_PATH="${data_base_path}/${valid_data_path}"
+
+  # Copy data to local storage for faster access
+  COPY_FLAG="/tmp/copy_done_${{SLURMD_NODENAME}}_${variant}";
+  NEW_DATA_PATH="/dev/shm/flip_${variant}_${{SLURMD_NODENAME}}";
+  if [ "$SLURM_LOCALID" = "0" ]; then
+      df -h;
+      echo "Copying data to $NEW_DATA_PATH";
+      mkdir -p $(dirname $NEW_DATA_PATH/${train_data_path});
+      mkdir -p $(dirname $NEW_DATA_PATH/${valid_data_path});
+      time cp $FULL_TRAIN_PATH $NEW_DATA_PATH/${train_data_path};
+      time cp $FULL_VALID_PATH $NEW_DATA_PATH/${valid_data_path};
+      touch $COPY_FLAG
+  fi
+  # All ranks wait until copy is done
+  while [ ! -f $COPY_FLAG ]; do
+      sleep 1
+  done
+
+  # Build the command with common arguments
+  CMD="WANDB_API_KEY=$BIONEMO_WANDB_API_KEY python ${workspace}/sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py \
+    --train-data-path=$NEW_DATA_PATH/${train_data_path} \
+    --valid-data-path=$NEW_DATA_PATH/${valid_data_path} \
+    --task-type=${task_type} \
+    --restore-from-checkpoint-path=${checkpoint_path} \
+    --config-class=${config_class} \
+    --dataset-class=${dataset_class} \
+    --num-steps=${num_steps} \
+    --experiment-name=${experiment_name}_${batch_size}bs_${nodes}node_${gpus}gpu_${num_steps}s_${precision}prec \
+    --lr=${lr} \
+    --result-dir=${tensorboard_dir} \
+    --micro-batch-size=${batch_size} \
+    --limit-val-batches=${limit_val_batches} \
+    --limit-test-batches=${limit_test_batches} \
+    --precision=${precision} \
+    --label-column=${label_column} \
+    --num-gpus=${gpus} \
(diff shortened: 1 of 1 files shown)


ci/benchmarks/partial-conv/esm2_finetuning.yaml
```diff
@@ -0,0 +1,163 @@
...
+  gpus: 1
+  precision: [bf16-mixed]
+  encoder_frozen: true
+  num_workers: 8
+  limit_val_batches: 1000
+  limit_test_batches: 1000
+  log_every_n_steps: 1
+  accumulate_grad_batches: 2
+  products:
+    - variant: seq_classification
```
dorotat-nv: could you rename variant="finetune" and and replace it with task: "seq_classification"?

ci/benchmarks/partial-conv/esm2_finetuning.yaml
```diff
@@ -0,0 +1,163 @@
...
+  log_every_n_steps: False
+script_args:
+  # All arguments referenced in the script string must be specified here.
+  # Arguments not referenced in the script string must have the 'arg' field specified.
+  # See jet/core/configs.py for the specification of the configuration class
+  workspace: /workspace/bionemo2
+  checkpoint_path: /data/checkpoints/esm2nv650m_v2.0
+  data_base_path: /data/FLIP
+  nodes: [1]
+  gpus: 1
```
dorotat-nv: please add model="esm2" and variant="finetune" and config_name="650M"

ci/benchmarks/partial-conv/esm2_finetuning.yaml
```diff
@@ -0,0 +1,163 @@
...
+    --precision=${precision} \
+    --label-column=${label_column} \
+    --num-gpus=${gpus} \
+    --num-nodes=${nodes} \
+    --accumulate-grad-batches=${accumulate_grad_batches} \
+    --val-check-interval=${val_check_interval} \
+    --log-every-n-steps=${log_every_n_steps} \
+    --num-dataset-workers=${num_workers} \
+    --wandb-project=${wandb_project_name} \
+    --wandb-group=esm2_finetune_${variant}__${target} \
```
dorotat-nv: could you add here ${model}__${variant}__${config_name}__${task}__${target}

ci/benchmarks/partial-conv/esm2_finetuning.yaml
```diff
@@ -0,0 +1,163 @@
+scope: partial-conv
+time_limit: 14400
```
dorotat-nv, line 2: is you set it to 3600 while working on the config, then it would be allocated faster

nvmvle:
/ok to test https://github.com/NVIDIA/bionemo-framework/pull/964/commits/6a7c952c39756823f6ecbc162639cf7694fdafad

3rdparty/Megatron-LM
pstjohn, line 1: did you intend to bump these sub-modules? 😅 
If so, I wouldn't do that in the same PR as adding the benchmark configuration

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
```diff
@@ -166,6 +168,28 @@ def get_parser():
     parser.add_argument("--lora-checkpoint-path", type=Path, default=None, help="LoRA checkpoint path")
     parser.add_argument("--lora-finetune", action="store_true", help="Use LoRA fine-tuning")
 
+    parser.add_argument(
+        "--disable-checkpointing",
+        action="store_false",
+        default=True,
+        dest="create_checkpoint_callback",
+        help="Disable creating a ModelCheckpoint callback.",
+    )
```
pstjohn, line 177: is there a reason we need to turn off checkpointing? even in a jet test, it might be nice to validate that creation still works even if they just get garbage collected at the end of a run. Or just turn the checkpointing frequency high enough so that it doesnt' get called

3rdparty/Megatron-LM
jstjohn: yeah I'm worried about this change because we recently made some changes to Evo2 that require megatron+nemo being pretty new.

3rdparty/NeMo
jstjohn, line 1: I'm worried about this change because we recently made some changes to Evo2 that require megatron+nemo being pretty new.

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
```diff
@@ -362,7 +390,7 @@ def train_model(
         autocast_dtype=get_autocast_dtype(precision),  # setting this speeds things up a lot
         tensor_model_parallel_size=tensor_model_parallel_size,
         pipeline_model_parallel_size=pipeline_model_parallel_size,
-        initial_ckpt_path=str(restore_from_checkpoint_path),
+        initial_ckpt_path=str(restore_from_checkpoint_path) if restore_from_checkpoint_path is not None else None,
```
jstjohn: Wait... what does fine-tuning mean if you do not have a `restore_from_checkpoint_path`? If this change was just for typing, better would be to make `restore_from_checkpoint_path` required... Please add a code comment on when a user would want to not have a checkpoint supplied during fine-tuning.

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
```diff
@@ -411,24 +439,44 @@ def train_model(
...
+            config,
+            data_module,
+            "bert",
+        )
+        callbacks.append(flop_meas_callback)
 
     trainer = nl.Trainer(
         devices=num_gpus,
-        max_steps=num_steps,
+        max_steps=num_steps if early_stop_on_step is None else early_stop_on_step,
```
jstjohn, line 479: If you want the same behavior as other repos use the early stop callback. The idea with `--early-stop-on-step` is to leave `max_steps` unchanged for things like logging/LR curves/etc, and use the callback to stop training partway through for testing early. Like... https://github.com/NVIDIA/bionemo-framework/blob/main/sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py#L775 is also wrong and explains some weirdness I saw in testing. I wonder why we got rid of that callback...

3rdparty/Megatron-LM
jwilber: +1. I have a meeting with My later where I'll make sure that our submodule pointers here align with ToT (and things work as expected) before merge

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
jwilber: Good question. This is the case for all of our [benchmarking scripts](https://github.com/NVIDIA/bionemo-framework/tree/main/ci/benchmarks). I don't remember if it's because of jet-specific storage bloat or wanting to eliminate the i/o overhead of the ephemeral state (or something mandated by jet?). That said, section 1.2.1 in the [SADD](https://docs.google.com/document/d/1PGIijuObov12J8MWJdx07oJmHr__AR9Xg0g_cl9_OII/edit?tab=t.0) does mention saving checkpoints, so this is something we may want to revisit cc @dorotat-nv

ci/benchmarks/perf/esm2_finetuning.yaml
```diff
@@ -0,0 +1,90 @@
...
+  train_data_path: scl/train/x000.csv
+  valid_data_path: scl/val/x000.csv
+  task_type: classification
+  config_class: ESM2FineTuneSeqConfig
+  dataset_class: InMemorySingleValueDataset
+  max_steps: 30000
+  stop_steps: 300
+  experiment_name: seq-level-classification
+  val_check_interval: 100
+  products:
```
jwilber, line 43: Do we want one of these variants to use `pp: 2` (which would then require adding the `--decoder-first-pipeline-num-layers=17` flag)? Is this useful to track for esm2? cc @pstjohn

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
nvmvle: We need to disable the checkpoint callback to ensure the pipeline does not fail due to the artifact size limit from JET Art.
FYI, here is the failed job without `disable-checkpointing` param: [188165756](https://gitlab-master.nvidia.com/dl/jet/ci/-/jobs/188165756), regarding this [Jenkins #621](https://prod.blsm.nvidia.com/bionemo-external-bionemo-fw/job/branch_pipeline_jet/621/)

jwilber (approved):

ci/benchmarks/perf/esm2_finetuning.yaml
jwilber: You'll also need to make `nodes: 2`

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
nvmvle: Yes, as I know, this `--early-stop-on-step` gives us the possibility to stop training early without modifying the training specification. Additionally, I follow the same logic used in other places in the project to handle this parameter. Maybe @dorotat-nv will have some ideas on this.

ci/benchmarks/partial-conv/esm2_finetuning.yaml
```diff
@@ -0,0 +1,76 @@
...
+  workspace: False
+  restore_from_checkpoint_path: False
+script_args:
+  # All arguments referenced in the script string must be specified here.
+  # Arguments not referenced in the script string must have the 'arg' field specified.
+  # See jet/core/configs.py for the specification of the configuration class
+  workspace: /workspace/bionemo2
+  data_base_path: /data/FLIP
+  restore_from_checkpoint_path: /data/esm2_650M_nemo2
+  nodes: [4]
```
yzhang123: why do you need so many nodes?

ci/benchmarks/perf/esm2_finetuning.yaml
```diff
@@ -0,0 +1,93 @@
...
+  limit_test_batches: 1
+  task: seq_classification
+  train_data_path: scl/train/x000.csv
+  valid_data_path: scl/val/x000.csv
+  task_type: classification
+  config_class: ESM2FineTuneSeqConfig
+  dataset_class: InMemorySingleValueDataset
+  max_steps: 30000
+  stop_steps: 300
+  experiment_name: seq-level-classification
```
yzhang123, line 41: @nvmvle are you also planning on adding seq-level regression  and token classification configurations?

ci/benchmarks/perf/esm2_finetuning.yaml
```diff
@@ -0,0 +1,93 @@
...
+    --lr=0.0005 \
+    --result-dir=${tensorboard_dir} \
+    --micro-batch-size=${batch_size} \
+    --limit-val-batches=${limit_val_batches} \
+    --limit-test-batches=${limit_test_batches} \
+    --precision=${precision} \
+    --label-column=scl_label \
+    --num-gpus=${gpus} \
+    --num-nodes=${nodes} \
+    --accumulate-grad-batches=2 \
```
yzhang123: since the global batch size in the fine-tuning case is 2*64=128, and you have 8 gpus, you don't need accumulate-grad, i.e. set --accumulate-grad-batches=1 and micro-batch-size=8, since accumulate grad > 1 makes it slower

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
```diff
@@ -166,6 +168,28 @@ def get_parser():
...
+        default=True,
+        dest="create_checkpoint_callback",
+        help="Disable creating a ModelCheckpoint callback.",
+    )
+
+    parser.add_argument(
+        "--early-stop-on-step",
+        type=int,
+        default=None,
+        help="Stop training on this step, if set. This may be useful for testing or debugging purposes.",
```
yzhang123, line 183: How is this used differently from max_steps?

nvmvle: --early-stop-on-step gives us the possibility to stop training early without modifying the training specification, like LR schedule...
`max_steps=num_steps if early_stop_on_step is None else early_stop_on_step`

ci/benchmarks/perf/esm2_finetuning.yaml
nvmvle: I set the --accumulate-grad-batches to 1

ci/benchmarks/perf/esm2_finetuning.yaml
nvmvle: I plan to add `seq-level regression` and `token classification` configurations, but I don't know should I add for both `partial-conv` and `perf`, or just for `perf` only?

ci/benchmarks/partial-conv/esm2_finetuning.yaml
nvmvle: updated to 1 node

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
```diff
@@ -127,7 +129,7 @@ def get_parser():
 
     # Checkpoint parameters
     parser.add_argument("--create-tensorboard-logger", action="store_true", help="Create tensorboard logger")
-    parser.add_argument("--restore-from-checkpoint-path", type=Path, default=None, help="Restore from checkpoint")
+    parser.add_argument("--restore-from-checkpoint-path", type=Path, required=True, help="Restore from checkpoint")
```
trvachov, line 132: Just so I understand, the correct behavior is when you finetune something, you should have a checkpoint you start from, correct? Just stating the obvious because it looks like `default=None` was very wrong before right?

trvachov (approved):

sub-packages/bionemo-esm2/src/bionemo/esm2/scripts/finetune_esm2.py
jwilber: Good question, so this may actually be `None`, now that I think about it.

To be clear, you should have a checkpoint when you fine-tune something, yes.. 

But here, it looks like we're forcing a `Path`, so the user has to be explicit about where the checkpoint is.

However, another option exists: if `--result-dir` is specified, and something is in there, pytorch-lightning will continue from the last checkpoint in that dir. (This was how we did things in the Evo 2 QA plan).

ci/benchmarks/perf/esm2_finetuning.yaml
yzhang123: partial-conv and perf

trvachov:
Bypassing rules to merge to unblock My -- we just released so low risk of screwing up any timelines.

PR #969, opened 2025-07-02
Code for general single cell benchmarking

polinabinder1:
### Description
This a simple, flexible framework for benchmarking any dataloader without requiring inheritance or modifications to your existing code. It works with any iterable dataloader. It measures dataset/ dataloader instantiation time and memory, along with iteration memory and speed. It can be easily run with multiple dataloaders to enable comparisons. The results are printed and exported into a csv. 

### Type of changes

- [x]  New feature (non-breaking change which adds functionality)
- [ ]  Refactor
- [ ]  Documentation update

### Usage

```python
import anndata as ad
from anndata.experimental import AnnCollection, AnnLoader
from bionemo.scspeedtest.benchmark import benchmark_single_dataloader, print_results
import numpy as np
filepath = "cellxgene_example_25k.h5ad"

#create a dataloader factory. This returns anndata in a dense format.
def anndata_factory(input_path, batch_size = 64):
    def factory():
        dataset = ad.read_h5ad(input_path)
        return AnnLoader(dataset, num_workers = 0,
        collate_fn = lambda batch: np.vstack([x.X for x in batch]))
    return factory

#benchmark the dataloader
result = benchmark_single_dataloader(
    dataloader_factory=anndata_factory(filepath),
    data_path=filepath,
    name="AnnLoader",
    max_time_seconds = 10
)

print_results(result)
```

### Pre-submit Checklist

 - [x] I have added/updated tests as needed
 - [x] All existing tests pass successfully


sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,324 @@
...
+    on disk. For directories, it recursively sums the sizes of all
+    files within the directory.
+
+    Args:
+        path: Path to file or directory to measure
+
+    Returns:
+        Size in megabytes (0.0 if path doesn't exist or error occurs)
+    """
+    result = subprocess.run(["du", "-sb", path], stdout=subprocess.PIPE, text=True)
```
skothenhill-nv: file_size_bytes = os.path.getsize(file_path)

i think is more likely to be OS agnostic

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Key Features
+
+- **Zero Inheritance Required**: Your dataloader doesn't need to inherit from anything
+- **Works with Any Iterable**: PyTorch DataLoaders, custom iterators, generators, lists, etc.
+- **Type Safety with Protocols**: Uses Python Protocols for better type checking
+- **Time-Based Benchmarking**: Set maximum runtime or warmup periods
+- **Modular Architecture**: Core benchmarking logic is reusable and extensible
+- **Comprehensive Metrics**: Disk space, memory usage, throughput, timing, **AND instantiation**
+- **Always Measures Instantiation**: Automatically measures dataloader creation time and memory
```
edawson: I suggest moving this to the end of the README and placing installation first, followed by a complete, standalone quickstart.

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+**Create a factory function and benchmark your dataloader!**
+
+```python
+from bionemo.scbenchmark import benchmark_dataloader
+
+# Create a factory function that returns your dataloader
+def create_my_dataloader():
+    dataset = MyDataset()
+    return DataLoader(dataset, batch_size=32)
```
edawson: Does this require a `from torch import DataLoader` import?

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,218 @@
...
+## Quick Start
+
+**Create a factory function and benchmark your dataloader!**
+
+```python
+from bionemo.scbenchmark import benchmark_dataloader
+
+# Create a factory function that returns your dataloader
+def create_my_dataloader():
+    dataset = MyDataset()
```
edawson: I think this can be any class that implements `__iter__` and `__len__`. Is that the case? If so, I would add a small example class definition before this so it's clear what MyDataset is.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: fixed

sub-packages/bionemo-scbenchmark/README.md
polinabinder1: fixed

sub-packages/bionemo-scbenchmark/README.md
polinabinder1: It doens't really matter what the dataset is = provided that it works with the dataloader.

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,177 @@
...
+### 1. Install Package
+
+```bash
+pip install -e .
+```
+
+**For baseline comparison** (optional):
+```bash
+pip install anndata bionemo-scdl scDataset
+```
```
skothenhill-nv: i'd suggest pinning versions and putting these in a requirements file. so it'd look like:

```bash
pip install -r baseline-requirements.txt
```

And baseline-requirements.txt:
```text
anndata=???
bionemo-scdl=???
scDataset=???
```

Keeps anything weird from breaking going forward. could probably test these too at some point.

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Quick Start
+
+Your dataloader just needs to be **iterable** (support `for batch in dataloader`).
+
+**Create a factory function**
```
skothenhill-nv: Is this still quick start? sounds like benchmarking a custom dataloader?

skothenhill-nv: as a reader I first think 'why would I run this, and what do I need to do to run this'

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,177 @@
...
+### Dataloader vs Dataset Factories
+
+The framework supports two distinct patterns for benchmarking, each optimized for different scenarios:
+
+**Dataloader Factory**: Creates both dataset and dataloader
+```python
+from torch import DataLoader
+
+def dataloader_factory():
+    dataset = load_dataset()  # Load data each time
```
skothenhill-nv: is `load_dataset` defined or imported somewhere?

sub-packages/bionemo-scbenchmark/README.md
```diff
@@ -0,0 +1,177 @@
...
+- Instantiation Time
+
+**Memory Usage:**
+- Peak memory (MB) during benchmarking and instantiation
+- Average memory (MB) throughout execution
+- Memory baseline tracking for accurate delta measurements
+
+**Storage & Resources:**
+- Disk usage of data files (MB)
+- Support for multiple files/directories
```
skothenhill-nv: These are features of `example/scdataset_script.py`?

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
```diff
@@ -0,0 +1,344 @@
...
+"""
+
+import os
+
+import anndata
+import torch
+from anndata.experimental import AnnCollection, AnnLoader
+from torch.utils.data import DataLoader
+
+# Import the benchmarking framework
```
skothenhill-nv:
```suggestion
```

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
```diff
@@ -0,0 +1,344 @@
...
+from bionemo.scbenchmark.common import export_benchmark_results
+from bionemo.scdl.io.single_cell_memmap_dataset import SingleCellMemMapDataset
+from bionemo.scdl.util.torch_dataloader_utils import collate_sparse_matrix_batch
+
+
+# Optional import for scDataset
+try:
+    from scdataset import scDataset
+except ImportError:
+    scDataset = None
```
skothenhill-nv: The readme made it seem like anndata and bionemo-scdl were optional too.

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
```diff
@@ -0,0 +1,344 @@
...
+# =============================================================================
+
+
+def anndata_collate(batch):
+    """Custom collate function for AnnData objects."""
+    # Convert AnnCollectionView to tensors
+    return torch.stack([item.X for item in batch])
+
+
+def create_annloader_config(batch_size=64, shuffle=True, num_workers=0, collate_fn=anndata_collate):
```
skothenhill-nv: method name is `create_annloader_config` but it looks like it returns a factory rather than a configuration.

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
```diff
@@ -0,0 +1,344 @@
...
+    def factory(dataset):
+        return AnnLoader(
+            dataset, batch_size=batch_size, shuffle=shuffle, num_workers=num_workers, collate_fn=collate_fn
+        )
+
+    return factory
+
+
+def create_scdl_traditional_factory(batch_size=64, shuffle=True, data_path=None, num_workers=0):
+    """Create a traditional SCDL dataloader factory that loads dataset each time.
```
skothenhill-nv: what does traditional mean?

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
```diff
@@ -0,0 +1,344 @@
...
+        dataset = SingleCellMemMapDataset(data_path)
+        print(f"✅ Dataset loaded: {len(dataset):,} samples")
+        return DataLoader(
+            dataset,
+            batch_size=batch_size,
+            shuffle=shuffle,
+            drop_last=False,
+            collate_fn=collate_sparse_matrix_batch,
+            num_workers=num_workers,
+        )
```
skothenhill-nv: suggestion for now or a soon update: for our dataloader factories we should probably pass around pre-defined kwargs for the DataLoaders. e.g. things like shuffle, drop last, num_workers, pin_memory... etc. this could be either a dict or something fancier like a dataclass.

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
```diff
@@ -0,0 +1,344 @@
...
+            "dataloader_factory": create_scdl_traditional_factory(
+                batch_size=64, shuffle=False, data_path=memmap_path, num_workers=0
+            ),
+            "num_epochs": num_epochs,
+            "max_time_seconds": 3.0,
+            "warmup_time_seconds": 0.5,
+            "data_path": memmap_path,
+            "num_runs": num_runs,
+        },
+    ]
```
skothenhill-nv: related to comments earlier (and applies to AnnData version of this)- might want to think about how to group dataloader kwargs within configurations. I usually prefer to have some kind of input validation, but the more I think about it the more I like the minimalist approach in this package (freeform dictionaries for configs)

sub-packages/bionemo-scbenchmark/examples/scddataset_test.py
```diff
@@ -0,0 +1,353 @@
...
+from bionemo.scdl.io.single_cell_memmap_dataset import SingleCellMemMapDataset
+from bionemo.scdl.util.torch_dataloader_utils import collate_sparse_matrix_batch
+
+
+# Optional import for scDataset
+try:
+    from scdataset import scDataset
+except ImportError:
+    scDataset = None
+
```
skothenhill-nv: same importing thing- readme implies anndata, bionemo-scdl, and scDataset are optional. but code only supports scDataset as optional.

skothenhill-nv: oh scDataset is also not optional in our tests- will fail if not installed in local environment.

sub-packages/bionemo-scbenchmark/examples/scddataset_test.py
```diff
@@ -0,0 +1,353 @@
...
+from datetime import datetime
+from typing import Sequence, Union
+
+import anndata
+import numpy as np
+import pandas as pd
+import torch
+from anndata.experimental import AnnCollection, AnnLoader
+from torch.utils.data import DataLoader
+
```
skothenhill-nv: What does this file (script? test?) do? It seems like there are duplicated methods in here, the file name indicates it might be a test- but there are no tests. is this stray dev code?

sub-packages/bionemo-scbenchmark/pyproject.toml
```diff
@@ -0,0 +1,71 @@
...
+    "isort>=5.0",
+    "flake8>=4.0",
+]
+
+[project.urls]
+Homepage = "https://github.com/NVIDIA/bionemo-framework"
+Repository = "https://github.com/NVIDIA/bionemo-framework"
+Documentation = "https://github.com/NVIDIA/bionemo-framework"
+Issues = "https://github.com/NVIDIA/bionemo-framework/issues"
+
```
skothenhill-nv: still good with these?

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+"""
+
+
+def _drop_caches():
+    """Helper function to drop system caches."""
+    try:
+        print("Dropping caches")
+        subprocess.run(["sh", "-c", "echo 3 > /proc/sys/vm/drop_caches"], check=True)
+    except subprocess.CalledProcessError:
+        print("⚠️ Warning: failed to drop caches — are you running with sudo?")
```
skothenhill-nv: document the operating systems this is tested with somewhere

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+    warmup_time_seconds: Optional[float] = None
+    print_progress: bool = True
+    data_path: Optional[Union[str, Path]] = None
+    madvise_interval: Optional[int] = None
+    shuffle: bool = False
+    num_workers: Optional[int] = None
+
+
+def run_benchmark(
+    dataloader: Any,
```
skothenhill-nv: dataloader: Any, should this be dataloader: DataLoader?

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+    config: BenchmarkConfig,
+    run_name: Optional[str] = None,
+    output_dir: Optional[str] = None,
+    instantiation_metrics: Optional[Dict[str, float]] = None,
+) -> BenchmarkResult:
+    """Run the actual benchmark and collect metrics.
+
```
skothenhill-nv: confused again, this looks like it runs a specific benchmark- I see scdl specific code below. But the module name is `benchmark.py` and the function is called `run_benchmark()`. What is being benchmarked? What should I expect from this file?

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+        total_samples=total_samples,
+        total_batches=total_batches,
+        setup_time=0,  # Will be set by caller
+        warmup_time=total_warmup_time,
+        elapsed_time=total_elapsed_time,  # Fixed: use total time
+        warmup_samples=total_warmup_samples,
+        warmup_batches=total_warmup_batches,
+        instantiation_metrics=instantiation_metrics,
+        epoch_results=epoch_results,
+    )
```
skothenhill-nv: if we are going the config route, i'd suggest using a set of dataclasses to pass into this. `from_raw_metrics` implies it only requires some 'raw metrics' to be instantiated. Are those `epoch_results?` what are the other things we pass in? I'd think about accepting multiple dataclasses (or dictionaries, but be consistent) if necessary here.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+
+    # Add epoch results to the result object for further analysis
+    result.epoch_results = epoch_results
```
skothenhill-nv: `epoch_results` is passed into `from_raw_metrics`, couldn't we assign epoch results as an attribute there?

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+
+    name: str
+    num_epochs: int = 1
+    max_batches: Optional[int] = None
+    max_time_seconds: Optional[float] = None
+    warmup_batches: Optional[int] = None
+    warmup_time_seconds: Optional[float] = None
+    print_progress: bool = True
+    data_path: Optional[Union[str, Path]] = None
+    madvise_interval: Optional[int] = None
```
skothenhill-nv: i think this is a scdl specific thing. we should put that in scdl rather than the benchmarking.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+                warmup_time_seconds=warmup_time_seconds,
+                print_progress=print_progress,
+                madvise_interval=madvise_interval,
+                shuffle=shuffle,
+                num_runs=num_runs,
+                output_dir=output_dir,
+                num_workers=num_workers,
+            )
+        else:
+            # Traditional mode: each config loads its own dataset
```
skothenhill-nv: can we delete traditional mode or is there a reason to keep it

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+                    # Save single result
+                    if output_dir:
+                        os.makedirs(output_dir, exist_ok=True)
+                        result.save_to_file(os.path.join(output_dir, f"{result.name}_results.json"))
+
+            # Print comparison
+            if len(results) > 1:
+                _print_comparison(results)
+
+            return results
```
skothenhill-nv: i'd suggest letting the consumer/caller of this method print the results.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+                csv_prefix = "consolidated_benchmark_results"
+            append_benchmark_result(
+                run_result, output_prefix=csv_prefix, create_headers=create_headers, output_dir=output_dir
+            )
+
+            del dataloader
+
+        # 🚫 Remove bulk export - we already appended each result individually
+        # export_benchmark_results(all_results)
+        return all_results
```
skothenhill-nv: two questions:

1) It is not obvious to me why the num_runs==1 case is special. Shouldn't the singular loop case look exactly the same as a multiple loop case?
2) In the multiple run case, setup time looks like its never reported.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+    warmup_batches: int = 5,
+    warmup_time_seconds: Optional[float] = None,
+    print_progress: bool = True,
+    madvise_interval: Optional[int] = None,
+    shuffle: bool = False,
+    num_runs: int = 1,
+    output_dir: Optional[str] = None,
+    num_workers: Optional[int] = None,
+) -> List[BenchmarkResult]:
+    """Benchmark multiple dataloader configs using a shared dataset (dataset reuse mode).
```
skothenhill-nv: similar to the comment above, I dont see any reason these cannot be collapsed into one. We should be able to infer if a dataset/dataloader is being reused. Our benchmark run and invocation should look the same in each case, the only difference being what is actually called when the dataset/dataloader is instantiated. Instead it looks like we duplicate the benchmarking code.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+            del dataloader
+
+    # 🆕 Step 9: Save results and print comparison
+    if output_dir:
+        os.makedirs(output_dir, exist_ok=True)
+        for result in all_results:
+            result.save_to_file(os.path.join(output_dir, f"{result.name}_results.json"))
+
+    if len(all_results) > 1:
+        _print_comparison(all_results)
```
skothenhill-nv: caller should print, (duplicated comment)

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+    warmup_time_seconds: float
+    total_iteration_time_seconds: float
+    average_batch_time_seconds: float
+    total_batches: int
+    total_samples: int
+    samples_per_second: float
+    batches_per_second: float
+    peak_memory_mb: float
+    average_memory_mb: float
+    errors: List[str] = None
```
skothenhill-nv: can do this with defaultfactory too
```suggestion
    errors: List[str] =  field(default_factory=list)
```

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+    warmup_samples: int = 0
+    warmup_batches: int = 0
+
+    # Speed metrics
+    total_speed_samples_per_second: float = 0.0  # Including warmup
+
+    def __post_init__(self):
+        """Initialize lists if not provided."""
+        if self.errors is None:
+            self.errors = []
```
skothenhill-nv: won't need this with field factory! :)

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+        total_batches: int = 0,
+        setup_time: float = 0.0,
+        warmup_time: float = 0.0,
+        elapsed_time: float = 0.0,
+        disk_size_mb: float = 0.0,
+        warmup_samples: int = 0,
+        warmup_batches: int = 0,
+        instantiation_metrics: Optional[Dict[str, float]] = None,
+        num_workers: Optional[int] = None,
+        epoch_results: Optional[List[Dict[str, Any]]] = None,
```
skothenhill-nv: I think I commented this previously- if we use configs, we should pack all configuration parameters into a few config objects instead of a big mix like this. it seems like total samplea, total batches, setup_time, warmup_time,elapsed_time, disk size, etc are all expected metrics to be generated. Maybe those should live in a `RawBenchmarkMetrics` dataclass or something.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+        Dictionary with derived metrics including:
+        - warmup_samples_per_sec
+        - num_epochs
+        - dataset_samples_per_epoch
+        - dataset_batches_per_epoch
+        - avg_batch_size
+        - dataset_size_k_samples
+        - inst_memory
+        - inst_time
+    """
```
skothenhill-nv: `from_raw_metrics` looks like it also creates some derived metris. perhaps they should be moved in here.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+    # Calculate warmup speed
+    warmup_samples_per_sec = (
+        (result.warmup_samples / result.warmup_time_seconds) if result.warmup_time_seconds > 0 else 0.0
+    )
+
+    # Get number of epochs (try multiple sources)
+    num_epochs = getattr(result, "num_epochs", 1) or 1
+    if hasattr(result, "epoch_results") and result.epoch_results:
+        num_epochs = len(result.epoch_results)
```
skothenhill-nv: this block seems really weird. I would think the experiment configuration would know the number of epochs though.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+    if detailed_rows:
+        pd.DataFrame(detailed_rows).to_csv(detailed_csv, mode=mode, header=header, index=False)
+        if should_create_headers:
+            print(f"📄 Created Detailed breakdown CSV: {os.path.abspath(detailed_csv)}")
+        else:
+            print(f"📄 Appended to Detailed breakdown CSV: {os.path.abspath(detailed_csv)}")
+
+
+def get_disk_size(path: Union[str, Path]) -> float:
+    """Get disk size of a file or directory in MB."""
```
skothenhill-nv: document which platforms this has been tested on.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+            size_in_blocks = int(result.stdout.split()[0])
+            size_in_bytes = size_in_blocks * 512  # macOS du uses 512-byte blocks by default
+        else:  # Linux and others
+            result = subprocess.run(["du", "-sb", str(path)], stdout=subprocess.PIPE, text=True, check=True)
+            size_in_bytes = int(result.stdout.split()[0])
+
+        return size_in_bytes / (1024 * 1024)
+    except (subprocess.CalledProcessError, ValueError, IndexError) as e:
+        print(f"Warning: Could not determine disk size for {path}: {e}")
+        return 0.0
```
skothenhill-nv: should raise here with the message. the caller should catch the exception and then assume the value is 0.0 if we want to recover.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,824 @@
...
+        pbar = tqdm(
+            desc=f"{config.name} - Epoch {epoch_num + 1}/{config.num_epochs}", disable=not config.print_progress
+        )
+        warm_up_start = time.perf_counter()
+        if not do_warmup:
+            config.warmup_time_seconds = 0
+        warm_up_end = warm_up_start + config.warmup_time_seconds
+        is_warming_up = True
+        for num, batch in enumerate(dataloader):
+            batch_size = get_batch_size(batch)
```
skothenhill-nv: isnt batch size an attribute on dataloader?

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,641 @@
...
+    This function attempts to determine the batch size from various
+    common batch formats including PyTorch tensors, lists, and
+    dictionaries with common keys.
+
+    Args:
+        batch: The batch object to measure
+
+    Returns:
+        Number of samples in the batch
+    """
```
skothenhill-nv: why do we need this?

skothenhill-nv:
> Lots to be done here, seems like there is deadcode and some half-baked interfaces. Let me know if you have questions about any of the specific comments. In general we should cleanup some of the redundant stuff and try to be consistent with how we pass around information.

Also, we need tests where possible! Examples should definitely be tested if we can pull it off, then the core features should get some as well, atleast the interfaces.

sub-packages/bionemo-scbenchmark/README.md
polinabinder1: did so!

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
polinabinder1: For the general dataloader, they are optional. In this example, they are not. Should I add them as required to make things easier?

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
polinabinder1: done

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
polinabinder1: This is changed going forward. This is when the dataset is reloaded every time.

sub-packages/bionemo-scbenchmark/examples/scddataset_test.py
polinabinder1: It is code for scdataset. I renamed it.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: This is for the case of a smaller batch at the end of the dataloader.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: Incomplete batch? Though I'm open to feedback there.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: done

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: This logic has been removed

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: Consolidated!

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: Consolidated things, let me know if it's enough.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: removed errors

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: removed errors

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: I refactored and merged these things.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: Refactored this, and it should avoid the issues.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: Fixed

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: done

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: With SCDL we would want to re-load every time potentially.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: OK, will do!

sub-packages/bionemo-scbenchmark/examples/comprehensive_benchmarking.py
polinabinder1: I added checking that the relevant kwargs (really only the dataloader factory and data_path) are there. Otherwise, default values are set.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,482 @@
...
+    Args:
+        dataloader: The dataloader to benchmark
+        config: Configuration for the benchmark run
+        run_name: Optional name for this run
+        **instantiation_kwargs: Instantiation metrics (dataset_instantiation_time_seconds,
+                               dataloader_instantiation_time_seconds, peak_memory_during_instantiation_mb,
+                               memory_before_instantiation_mb, memory_after_instantiation_mb)
+
+    Returns:
+        BenchmarkResult containing all collected data and calculated metrics
```
skothenhill-nv: Awesome- I think the data-flow is a lot more clear now.  i liked the name `instantiation_metrics` that you used when you called `run_benchmark`

or maybe more broadly, it seems like we could break this up into two steps, I think this will make testing a little cleaner too:
1) compute instantiation results/metrics
2) compute benchmark results/metrics
3) compile final benchmarking report (from results/metrics)

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,482 @@
...
+
+        dataloader, dl_baseline, dl_peak, _, _, dl_final, dl_time = measure_peak_memory_full(dataloader_from_dataset)
+
+        instantiation_metrics = {
+            "peak_memory_during_instantiation_mb": max(dl_peak, dataset_peak),
+            "memory_after_instantiation_mb": dl_final,
+            "memory_before_instantiation_mb": dataset_baseline_measured,
+            "dataset_instantiation_time_seconds": dataset_time,
+            "dataloader_instantiation_time_seconds": dl_time,
+        }
```
skothenhill-nv: I didnt look at these super closely- apologies if it doesnt make sense.

Could we have `measure_peak_memory_full(...)` reutrn the instantiation metrics? maybe as instantiation_results (or something).

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,482 @@
...
+            "peak_memory_during_instantiation_mb": peak,
+            "memory_after_instantiation_mb": final_mib,
+            "memory_before_instantiation_mb": dataset_baseline
+            if dataset_baseline is not None
+            else dataloader_baseline_measured,
+            "dataset_instantiation_time_seconds": dataset_instantiation_time
+            if dataset_instantiation_time is not None
+            else 0,  # Combined time when no separate dataset factory
+            "dataloader_instantiation_time_seconds": setup_time,
+        }
```
skothenhill-nv: same comment here about returning the metrics directly from `measure_peak_memory_full`

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,482 @@
...
+            name=run_name_str,
+            num_epochs=num_epochs,
+            max_batches=max_batches,
+            max_time_seconds=max_time_seconds,
+            warmup_batches=warmup_batches,
+            warmup_time_seconds=warmup_time_seconds,
+            data_path=data_path,
+            shuffle=shuffle,
+        )
+        run_result = run_benchmark(current_dataloader, run_config, run_name_str, **instantiation_metrics)
```
skothenhill-nv: love the direction this is going. it seems to me that we could break this up into de-coupled steps:

1) get instantiation metrics/results (whatever the output of `measure_peak_memory_full` is) as a 'Result' ( or whatever ).
2) get benchmarking results (throughput)? (produce BenchmarkResult, or whatever)
           maybe we should be more specific on what 'run_benchmark' is- since the memory check is also a benchmark, just a different kind of benchmark haha.
 3) combine all the 'Results' objects that you care about into a final report:
          BenchmarkingReport = ThroughputBenchmarkResult (w/ rename?) + MemoryBenchmarkResult + ComputeContext (I'm see disk_size below, im assuming there are other similar things youre collecting?)
          
 
 Let me know if you want to discuss this further, I didn't think about it very hard, I think it will it would pull apart the right pieces though.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
```diff
@@ -0,0 +1,430 @@
...
+                    else 0,
+                }
+            )
+
+    # Write detailed CSV (always append, never overwrite)
+    pd.DataFrame(detailed_rows).to_csv(detailed_csv, mode="a", header=header, index=False)
+    if not file_exists:
+        print(f"Created Detailed breakdown CSV: {os.path.abspath(detailed_csv)}")
+    else:
+        print(f"Appended to Detailed breakdown CSV: {os.path.abspath(detailed_csv)}")
```
skothenhill-nv: love it

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
```diff
@@ -0,0 +1,482 @@
...
+                    end_time = start_time + config.max_time_seconds if config.max_time_seconds is not None else None
+                    pbar.set_description(f"{config.name} - Epoch {epoch_num + 1} (warmup complete)")
+                else:
+                    if warmup_batches % update_interval == 0:
+                        elapsed_warmup = current_time - warm_up_start
+                        current_warmup_speed = warmup_samples / elapsed_warmup if elapsed_warmup > 0 else 0
+                        pbar.set_description(
+                            f"{config.name} - Warmup: {elapsed_warmup:.1f}/{config.warmup_time_seconds}s, {current_warmup_speed:.1f} samples/sec"
+                        )
+                        pbar.update(update_interval)
```
skothenhill-nv: I think we should carry over that warning/error if warmup consumes the entire dataset. probably a big ol' logger.warn

sub-packages/bionemo-scbenchmark/tests/bionemo/scbenchmark/test_benchmark.py
```diff
@@ -0,0 +1,407 @@
...
+    BenchmarkConfig,
+    benchmark_dataloaders_with_configs,
+    benchmark_single_dataloader,
+    run_benchmark,
+)
+from bionemo.scbenchmark.common import BenchmarkResult
+
+
+class MockDataloader:
+    """Mock dataloader for testing."""
```
skothenhill-nv: Quick pass but the unit tests look good. Maybe we should add a function/e2e test that runs a real benchmark? Benchmarking has to happen at runtime, i'm guessing unit tests (in most cases) won't be super helpful other thank making sure nothing silly is broken. e2e test that checks for edge conditions (like samples / s > 0, memory isnt nan, i dunno stuff like that) could give more confidence it works when everything is integrated.

skothenhill-nv: maybe what you did for speed-test (the configuration/autodownload) would make sense for this kind of thing ? Bonus is that it's something we are showing users.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: Will think about it.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: measure_peak_memory_full is called on the dataloader and/or dataset.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: measure_peak_memory_full is called on the dataloader and/or dataset. There is some nuance on whether there is a dataloader/ dataset and so on.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: Will think about it.

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/common.py
polinabinder1: ty

sub-packages/bionemo-scbenchmark/src/bionemo/scbenchmark/benchmark.py
polinabinder1: done

skothenhill-nv (approved):
thank you!

trvachov (approved):

PR #978, opened 2025-07-07
Fix database path in esm2_pretrain_recipes

pstjohn:
The database path still needs to be the sqlite file, currently these jobs are failing with
```
 2: [rank2]: sqlite3.DatabaseError: file is not a database
```


pstjohn:
force-merging while @dorotat-nv's out

PR #983, opened 2025-07-16
Reset parameters for the ESM-2 contact head on HF export

pstjohn:
Previously this weight was not initialized, and would be the result of whatever torch.empty() assigned to these values. If we want to use this layer in downstream tasks, we should at least randomly init these weights to reasonable values.


sub-packages/bionemo-esm2/tests/bionemo/esm2/model/test_convert.py
```diff
@@ -57,6 +57,9 @@ def test_nemo2_export_8m_weights_equivalent(tmp_path):
     hf_model_from_nemo = AutoModelForMaskedLM.from_pretrained(output_path)
     hf_model_from_hf = AutoModelForMaskedLM.from_pretrained("facebook/esm2_t6_8M_UR50D")
 
+    # We should not have any NaNs in the model, even in heads that we don't explicitly initialize.
```
jstjohn, line 60: empty shouldn't have NaN? it should just be junk without any particular distribution.

Any reason not to add initialization to something reasonable for this variable?

jstjohn (approved):

PR #989, opened 2025-07-18
change where dtype is found in checkpoint export

pstjohn:
For some reason the previous way we determined the bionemo model's dtype wasn't working correctly for nv-trained checkpoint exports; here we use model.dtype directly. Adds a test for exporting the nvidia-trained ESM checkpoints.

### DIFF
--- a/sub-packages/bionemo-esm2/tests/bionemo/esm2/model/test_convert.py
+++ b/sub-packages/bionemo-esm2/tests/bionemo/esm2/model/test_convert.py
@@ -104,6 +104,14 @@ def test_nemo2_conversion_equivalent_650m(tmp_path):
         assert_esm2_equivalence(tmp_path / "nemo_checkpoint", model_tag, atol=1e-4, rtol=1e-4)
 
 
+@pytest.mark.slow
+def test_nemo2_export_equivalent_650m(tmp_path):
+    ckpt_path = load("esm2/nv_650m:2.1")
+    output_path = io.export_ckpt(ckpt_path, "hf", tmp_path / "hf_checkpoint")
+    with megatron_parallel_state_utils.distributed_model_parallel_state():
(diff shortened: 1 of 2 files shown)


jstjohn (approved):

sub-packages/bionemo-esm2/tests/bionemo/esm2/model/test_convert.py
```diff
@@ -104,6 +104,14 @@ def test_nemo2_conversion_equivalent_650m(tmp_path):
         assert_esm2_equivalence(tmp_path / "nemo_checkpoint", model_tag, atol=1e-4, rtol=1e-4)
 
 
+@pytest.mark.slow
+def test_nemo2_export_equivalent_650m(tmp_path):
+    ckpt_path = load("esm2/nv_650m:2.1")
+    output_path = io.export_ckpt(ckpt_path, "hf", tmp_path / "hf_checkpoint")
+    with megatron_parallel_state_utils.distributed_model_parallel_state():
+        assert_esm2_equivalence(ckpt_path, output_path, precision="bf16")
```
jomitchellnv, line 112: How is the default for BF16 set, if we removed the default in https://github.com/NVIDIA/bionemo-framework/pull/989/files#diff-a47e5a3e6b4e6e6cbcf30a2254fe7a23488281aa61fa2ff72a400e4bd5d0a555R137 but are checking it here.

jomitchellnv (approved):
LGTM I have one question but otherwise ok

sub-packages/bionemo-esm2/tests/bionemo/esm2/model/test_convert.py
pstjohn: the precision here is what precision to use to eval both models. 
the precision in convert.py is just making sure the models are both in the same dtype, otherwise we get an error saying the dtype for the state_dicts don't match up

PR #1000, opened 2025-07-23
Hyena Inference Updates to support Flash Decode

jstjohn:
### Description
- Support for Flash Decode at inference time in evo2. The changes had been made upstream in NeMo previously ( https://github.com/NVIDIA/NeMo/pull/14315).
- Do not yet support cudagraph, which is another potential performance gain at inference time.
- Temporary work around for the issue of increased memory usage whenever passing an inference_context to hyena introduced by the flash decode API change: https://github.com/NVIDIA/bionemo-framework/issues/1013. This unblocks bumping nemo related PRs such as https://github.com/NVIDIA/bionemo-framework/pull/1012.
- All of this is related to supporting a future NIM backed by bionemo. Unfortunately this is demonstrating how much code overhead there is in adding this kind of support, and with the memory usage issues, the speed improvement may not be worth it.
- Training performance, as well as forward pass without inference context performance is not impacted.

### DIFF
--- a/sub-packages/bionemo-evo2/src/bionemo/evo2/run/infer.py
+++ b/sub-packages/bionemo-evo2/src/bionemo/evo2/run/infer.py
@@ -25,6 +25,7 @@ from typing import Literal, Optional
 import nemo.lightning as nl
 import torch
 from megatron.core.inference.common_inference_params import CommonInferenceParams
+from megatron.core.inference.inference_request import InferenceRequest
 from nemo.collections.llm import inference
 from nemo.utils import logging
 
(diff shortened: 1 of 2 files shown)


jstjohn:
TODO wait for https://github.com/NVIDIA/NeMo/pull/14315 to merge and upstream to Main NeMo which also has the change from https://github.com/NVIDIA/NeMo/pull/14359

Update: This is now done.

skothenhill-nv (approved):
I didn't see anything glaring. What would be helpful is more in the PR description. I'm missing a lot of context, why did we make these changes? how should we think about this if someone encounters a non-related bug in the future from this PR?

farhadrgh (approved):
LGTM

Issue #1007, opened 2025-07-29
[BUG] Evo2 dataset contains degenerate bases, and masking algorithm does not respect that.


jstjohn:
### BioNeMo Framework Version

c5b35a86465c3092121a826a3c0ba8c7d3ee36b2

### Bug Description

The following 8k sequence was found to be fully masked in the evo2 dataset at sample time, and is that way because of the existence of non-DNA degenerate tokens in some sequences in the evo2 training set.
```
[...]

PR #1008, opened 2025-07-29
Better masking in evo2, and default use of more robust megatron loss

jstjohn:
### Description
Switch to megatron's more robust `calculate_per_token_loss=True` by default, and better masking procedure in NeMo that will be less likely to mask entire sequences.

### Type of changes

- [ ]  Refactor
- [ ]  Documentation update

Relates to https://github.com/NVIDIA/bionemo-framework/issues/1007 and depends on https://github.com/NVIDIA/NeMo/pull/14359

### DIFF
--- a/sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
+++ b/sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
@@ -384,8 +384,9 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
     parser.add_argument(
         "--mamba-lowercase-loss-weight",
         type=float,
-        default=1.0,
-        help="Loss weight for the Mamba model for lowercase bases, if you are using a Mamba model.",
+        default=0.1,
+        help="Loss weight for the Mamba model for lowercase bases, if you are using a Mamba model. "
+        "Default is 0.1 like the Evo2 paper. Set to 1.0 to disable differential loss weighting.",
     )
     # rank as list of integers
     parser.add_argument(
--- a/sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
+++ b/sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
@@ -250,7 +250,7 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
         help="Restore optimizer state from initial checkpoint. Defaults to False.",
     )
     parser.add_argument(
-        "--no-average-in-collective",
+        "--average-in-collective",
         action="store_true",
         default=False,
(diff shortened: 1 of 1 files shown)


pstjohn (approved):

sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
```diff
@@ -381,6 +381,12 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
         default=False,
         help="Do not renormalize the loss weights.",
     )
+    parser.add_argument(
+        "--mamba-lowercase-loss-weight",
```
trvachov, line 385: maybe I should know -- there's a difference between lowercase and upper case base pairs?

cspades (approved):
Unblocking review, LGTM

sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
```diff
@@ -250,7 +250,7 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
         help="Restore optimizer state from initial checkpoint. Defaults to False.",
     )
     parser.add_argument(
-        "--no-average-in-collective",
+        "--average-in-collective",
```
trvachov, line 253: Is this change (and corresponding function param changes) just to make the boolean logic simpler? note this also changes the default (default=False is unchanged, so default behavior is different now)

sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
cspades: I recall case-sensitivity was definitely an issue and/or a feature.

jstjohn: Yes, lowercase basepairs represent "repeat" elements. The default in the original evo2 publication was to assign 0.1 weight to them.

sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
jstjohn: Correct, `average_in_collective` is not supported with `calculate_per_token_loss=True` in megatron. Since we want to use `calculate_per_token_loss` by default (more robust, handles 0 unmasked tokens in a microbatch properly) we also need `average_in_collective=False` by default now where before it was default True as you said. It should be equivalent, but theoretically slightly faster.

sub-packages/bionemo-evo2/src/bionemo/evo2/run/train.py
jstjohn: I can change this to 0.1, I like that better I think.

PR #1038, opened 2025-08-13
add changed files to github workflow

pstjohn:
Uses changed_files action to detect whether any framework-related code has changed, and only triggers the bionemo image build and tests in that event.

We should continue to run these tests on the schedule event, but skip them if no framework files have changed in the PR and merge_group CI.


.github/workflows/unit-tests-framework.yml
```diff
@@ -20,18 +20,51 @@ concurrency:
...
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          submodules: "recursive"
       - uses: actions/setup-python@v5
         with:
           python-version: "3.13"
-          cache: "pip"
-      - run: pip install -r requirements-dev.txt
```
jstjohn, line 34: Are we not doing dev requirements anymore? Maybe the idea is because this happens in the context of unit tests we do not want them there, but in general these get included in build?

jstjohn (approved):

.github/workflows/unit-tests-framework.yml
pstjohn: yeah that's right. we don't need to install them just to run pre-commit 🤷 . pre-commit handles that for us

PR #1052, opened 2025-08-19
Add bionemo-recipes

pstjohn:
Adds BioNeMo-recipes, examples for training high-performance biology models outside of NeMo using TransformerEngine and nvFSDP.

Most of the new work in this PR is around getting CI running on the nv_gha_runner backend. In summary, I've renamed the previous CI to "unit-tests-framework.yml", and use step_security/changed_files to detect when any code changes outside of recipes/ and models/

The new recipes CI lives in "unit-tests-recipes.yml", and also uses step_security/changed_files to build a list of directories to test. Since the models and recipes are independent folders, we only run CI for each if a file has changed in them. We then use a matrix strategy job to kick off a job for each changed models/ or recipes/ folder.


.devcontainer/framework/devcontainer.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "BioNeMo2 Development Container",
+	"name": "BioNeMo2 Framework Dev Container",
```
jstjohn, line 2: So before you would open VSCode or Cursor and the devcontainer button would just pop up and ask you if you want to launch the code in a devcontainer. How does this structure change things? How does a user pick/switch between the framework and recipe dev container?

Maybe we should add a section to the readme talking about any special things that would need to be done to switch back and forth or whatever you think the development user experience is like now?

pstjohn: Yeah that's right, after running "> Dev Containers: Rebuild or Reopen in Container", you get a dropdown list to choose from:

Screenshot 2025-08-22 at 10 37 06 AM

Screenshot 2025-08-22 at 10 37 12 AM

cspades (approved):
Reviewed a few use cases of Megatron-FSDP, those LGTM assuming we have some good unit tests on our side.

.devcontainer/recipes/requirements.txt
```diff
@@ -0,0 +1,13 @@
+--extra-index-url https://test.pypi.org/simple/ # TODO: Remove this once megatron-fsdp is public
+accelerate
+datasets
+deepspeed
+hydra-core
+torch
+megatron-fsdp==0.1.0rc0
+torchmetrics
+tqdm
+# transformer_engine  # Add when we're not deriving from pytorch base container
```
jomitchellnv: what's the plan for using TE versions that aren't available inside the pytorch base container

jomitchellnv (approved):
Approved

.devcontainer/recipes/requirements.txt
pstjohn: I'm currently getting this error trying to install TE with uv:
```
0.453   × Failed to resolve dependencies for `transformer-engine` (v2.4.0+3cd6870)                                                                                                     
0.453   ╰─▶ Package `nvdlfw-inspect` was included as a URL dependency.                                                                                                                 
0.453       URL dependencies must be expressed as direct requirements
0.453       or constraints. Consider adding `nvdlfw-inspect @
0.453       git+https://github.com/NVIDIA/nvidia-dlfw-inspect.git@v0.1#egg=nvdlfw-inspect`
0.453       to your dependencies or constraints file.
```

pstjohn: OK, i've reverted to using `pip` inside the git workflow rather than UV, and keeping TE in the dependencies list. It still may be tricky to use TE versions that are outside the base container until their pip installation story is easier, just because we (currently) don't cache these pip installs

pstjohn:
force-merging with CI passing and dorota OOO

PR #1062, opened 2025-08-21
update license check and ignore .gitignore in framework ci

pstjohn:
Should allow the bionemo-recipes PR to avoid triggering the framework CI


pstjohn:
test failure looks like a huggingface hub timeout; pre-commit passes which was the only thing this touches. Merging to unblock bionemo-recipes

PR #1075, opened 2025-08-26
add continue-on-error to get-pr-info

pstjohn:
Looks like get-pr-info can fail on submodule updates. Let's just continue on error here, or maybe we can revert to the PR labels action we use in bionemo-framework.


pstjohn:
closing in favor of #1079

PR #1076, opened 2025-08-26
Remove index.md markdown to prevent 3.13 requirement

jwilber:
The cards in the docs `index.md` require a specific format.
However, `mdformat` changes this format. To workaround this, we had to add excludes functionalty for `mdformat` that prevented it from formatting `index.md`. For some reason, the excludes functionality in `mdformat` which requires python 3.13. This is broke our nightly build.

To fix this, I've converted the `index.md` card logic from markdown to html, so that it no longer gets formatted. This removes the need for the excludes functionality and  the 3.13 dependency it brings.

### DIFF
--- a/.mdformat.toml
+++ b/.mdformat.toml
@@ -1,4 +1,1 @@
 number = true        # options: {false, true}
-exclude = [
-    "docs/docs/index.md",	# Exclude highly formatted index page
-]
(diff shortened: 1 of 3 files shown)


cspades (approved):
Had to approve, the code was a bit hard to look at, approving minimizes those boxes.

(Also, how do you get that... byte code (?) from an SVG, can we always do this with any SVG we find?)

pstjohn (approved):

pstjohn:
all relevant CI has passed, merging to save some time

PR #1077, opened 2025-08-26
Add ESM-2 model gradient tests

pstjohn:
Adds gradient tests to ensure the gradients we get from ddp, fsdp2, and nvfsdp are consistent

## Summary by CodeRabbit

- Chores
  - Simplified image builds by switching to standard pip for editable installs.
- Dependencies
  - Updated core ML dependencies and enabled/unpinned select components for better compatibility.
- Tests
  - Added comprehensive distributed-training validation tests comparing strategies across single- and multi-GPU setups with automatic GPU gating and detailed checks.
- Refactor
  - Extracted test data generation into a reusable helper to streamline test setup.

### DIFF
--- a/models/esm2/tests/test_distributed_strategies.py
+++ b/models/esm2/tests/test_distributed_strategies.py
@@ -40,7 +40,7 @@ requires_multi_gpu = pytest.mark.skipif(
     ],
 )
 @pytest.mark.parametrize("backend", ["te", "eager"])
-def test_ddp_vs_fsdp2_single_gpu(strategy, backend):
+def test_ddp_vs_fsdp_single_gpu(strategy, backend):
     cmd = [
         "torchrun",
         "--nproc_per_node=1",
@@ -68,7 +68,13 @@ def test_ddp_vs_fsdp2_single_gpu(strategy, backend):
 @requires_multi_gpu
 @pytest.mark.parametrize("strategy", ["fsdp2", "mfsdp"])
 @pytest.mark.parametrize("backend", ["te", "eager"])
-def test_ddp_vs_fsdp2_multi_gpu(strategy, backend):
+def test_ddp_vs_fsdp_multi_gpu(strategy, backend):
+    if strategy == "mfsdp":
+        pytest.skip(
+            "MFSDP multi-gpu tests are currently failing because tensors are not always evenly sharded, leaving p.grad "
+            "to be None on some ranks (BIONEMO-2726)"
+        )
+
     cmd = [
         "torchrun",
         "--nproc_per_node=2",
--- a/models/esm2/tests/test_distributed_strategies.py
+++ b/models/esm2/tests/test_distributed_strategies.py
@@ -1,0 +1,245 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-Apache2
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import logging
+import os
+import subprocess
+
+import pytest
+import torch
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+requires_multi_gpu = pytest.mark.skipif(
+    not torch.cuda.is_available() or torch.cuda.device_count() < 2,
+    reason="Test requires at least 2 GPUs",
+)
+
+
+@pytest.mark.parametrize(
+    "strategy",
+    [
+        "fsdp2",
+        "mfsdp",
+    ],
+)
+@pytest.mark.parametrize("backend", ["te", "eager"])
+def test_ddp_vs_fsdp2_single_gpu(strategy, backend):
+    cmd = [
+        "torchrun",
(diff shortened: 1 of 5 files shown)


models/esm2/tests/test_distributed_strategies.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # get gradients
+        if strategy is Strategy.FSDP2:
+            grads = {name: p.grad.full_tensor() for name, p in model.named_parameters() if p.grad is not None}
+
+        elif strategy is Strategy.DDP:
+            grads = {name: p.grad for name, p in model.module.named_parameters() if p.grad is not None}
+
+        elif strategy is Strategy.MFSDP:
+            grads = {name: p.grad.full_tensor() for name, p in model.module.named_parameters() if p.grad is not None}
```
pstjohn: @cspades do you see anything here that would be causing multi-device tests to hang?

models/esm2/tests/test_distributed_strategies.py
```diff
@@ -0,0 +1,245 @@
...
+        stderr=subprocess.PIPE,
+        timeout=240,
+    )
+    if result.returncode != 0:
+        print(f"STDOUT:\n{result.stdout}")
+        print(f"STDERR:\n{result.stderr}")
+        pytest.fail(f"Command failed with exit code {result.returncode}")
+
+
+@requires_multi_gpu
```
jomitchellnv, line 68: do we need a more precise decorator that says "requires 2 gpus"?

models/esm2/tests/test_distributed_strategies.py
```diff
@@ -0,0 +1,245 @@
...
+
+if __name__ == "__main__":
+    import argparse
+    import enum
+    from dataclasses import dataclass, field
+
+    import torch.distributed as dist
+    import transformer_engine.pytorch
+    import transformers
+    from megatron_fsdp.fully_shard import fully_shard as megatron_fsdp_fully_shard
```
jomitchellnv, line 104: so we have to import mcore just to get nvFSDP? yikes

models/esm2/tests/test_distributed_strategies.py
```diff
@@ -0,0 +1,245 @@
...
+
+        rank: int = field(default_factory=dist.get_rank)
+        local_rank: int = field(default_factory=lambda: int(os.environ["LOCAL_RANK"]))
+        world_size: int = field(default_factory=dist.get_world_size)
+
+        def is_main_process(self) -> bool:
+            """This is the global rank 0 process, to be used for wandb logging, etc."""
+            return self.rank == 0
+
+    def run_forward_backward(use_te: bool, strategy: Strategy, input_data: dict, dist_config: DistributedConfig):
```
jomitchellnv, line 136: so much if else if stuff, maybe just split it out for separate functions at the top of the file?

models/esm2/tests/test_distributed_strategies.py
```diff
@@ -0,0 +1,245 @@
...
+        timeout=240,
+    )
+    if result.returncode != 0:
+        print(f"STDOUT:\n{result.stdout}")
+        print(f"STDERR:\n{result.stderr}")
+        pytest.fail(f"Command failed with exit code {result.returncode}")
+
+
+@requires_multi_gpu
+@pytest.mark.parametrize("strategy", ["fsdp2", "mfsdp"])
```
jomitchellnv: I always prefer to have more tests than to use parametrize here, that way its clear. It's easier to edit down the line.

models/esm2/tests/test_distributed_strategies.py
```diff
@@ -0,0 +1,245 @@
...
+@pytest.mark.parametrize("backend", ["te", "eager"])
+def test_ddp_vs_fsdp2_single_gpu(strategy, backend):
+    cmd = [
+        "torchrun",
+        "--nproc_per_node=1",
+        os.path.relpath(__file__),
+        "--strategy",
+        strategy,
+    ]
+    if backend == "te":
```
jomitchellnv, line 51: nit: you should just make another test that doesn't use TE.

models/esm2/tests/test_distributed_strategies.py
pstjohn: no, this is `megatron-fsdp`: [pypi.org/project/megatron-fsdp](pypi.org/project/megatron-fsdp)

we're not calling it nvFSDP anymore. It's megatron-fsdp.

models/esm2/tests/test_distributed_strategies.py
pstjohn: fair; id just need to have multiple test files probably then. These tests are slightly odd in that there's a `__main__` entrypoint we call from the actual pytest tests, since we need to invoke these as subprocesses to ensure they're run with multiple processes

models/esm2/tests/test_distributed_strategies.py
cspades: Can we change it to this:
```
        elif strategy is Strategy.MFSDP:
            # Because of uneven sharding, we need to manually gather the gradients.
            sharded_grads = [(name, p.grad) for name, p in model.module.named_parameters()]
            grads = {}
            for name, grad in sharded_grads:
                grad_shards = [None] * device_mesh["dp"].size()
                # For FSDP, we are not strided sharding, so gathering across dp_shard_cp is sufficient.
                # For HSDP, we need to first gather across dp_shard_cp, then gather across dp_inter,
                # not the other way around or you'll get wrong zig-zags.
                torch.distributed.all_gather_object(grad_shards, grad, group=device_mesh["dp"].get_group())
                all_valid_shards = [shard for shard in grad_shards if shard is not None]
                # Megatron-FSDP is always sharded across dim=0.
                grads[name] = torch.cat([s.to_local().to(device) for s in all_valid_shards], dim=0)
```
The test will still fail as-is due to our gradient x2 issue, but this code will be useful for the future.

models/esm2/tests/test_distributed_strategies.py
pstjohn: right, as explained above i agree but this test suite is a little different, since we need to call `torchrun` on a file with different argparse commands. idk. nice to keep them together in the same file

PR #1113, opened 2025-09-04
uses image with zstd compression

pstjohn:
Uses an image verified to use zstd compression, which will hopefully give faster builds

## Summary by CodeRabbit

- Chores
  - Updated the default container image tag used by CI/local test runs.
- Tests
  - No changes to test behavior or execution flow; workflows continue to run as before.
  - Fallback behavior for selecting a container remains unchanged.


pstjohn:
merged elsewhere

PR #1127, opened 2025-09-07
Add wandb logic to recipes esm2_accelerate

jwilber:
Add wandb logic to esm2_accelerate

## Summary by CodeRabbit

- New Features
  - Integrated Weights & Biases experiment tracking with configurable run settings.
  - Added automatic logging of training step duration (train/step_time) for performance monitoring.

- Chores
  - Updated the L0_sanity configuration to include default Weights & Biases project, run name, and mode.
  - Ensured clean shutdown of tracking and distributed processes after training to avoid lingering runs or sessions.

### DIFF
--- a/recipes/esm2_accelerate/hydra_config/L0_sanity.yaml
+++ b/recipes/esm2_accelerate/hydra_config/L0_sanity.yaml
@@ -12,3 +12,8 @@ trainer:
   logging_steps: 1
   report_to: "none"
   dataloader_num_workers: 0
+
+wandb_init_args: # These arguments are for managing the weights and biases experiment.
+  name: "L0_sanity" # this setting defines the name of the experiment
+  project: "bionemo-recipes-esm2_accelerate" # this setting defines the project name
+  mode: "online"
--- a/recipes/esm2_accelerate/callbacks.py
+++ b/recipes/esm2_accelerate/callbacks.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+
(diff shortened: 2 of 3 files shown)


pstjohn (approved):

recipes/esm2_accelerate/hydra_config/L0_sanity.yaml
```diff
@@ -12,3 +12,8 @@ trainer:
   logging_steps: 1
   report_to: "none"
   dataloader_num_workers: 0
+
+wandb_init_args: # These arguments are for managing the weights and biases experiment.
```
dorotat-nv, line 16: Imho  the grouping should be different to give nice comparison between runs and keep history undestrubed 

ie 
* project: bionemo-recipe-L0-sanity-tests, ideally if done by CI with suffix --<branch_name> since they will be run on different branches

* group=esm2  if done by CI with suffix --<GPU_TYPE> where GPU type is A100 H200 or similar since these configs would be run on a different GPU types
* job_type = accelerate
* name - for CI work, ideally <DATE>_<COMMIT_SHA>_<PIPELINE_ID>

please add 
https://github.com/NVIDIA/bionemo-framework/blob/223b2111d93f8af5da0aa882fc6a2ef3f6f51262/ci/benchmarks/partial-conv/evo2_pretrain.yaml#L85

You can have in base.yaml ie which are required

git:
  sha: "${GIT_SHA}"
  branch: "${GIT_BRANCH}"
  
 and defined as 
 export GIT_SHA=$(git rev-parse --short HEAD)
export GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)

PR #1135, opened 2025-09-08
Update amplify model, add loss tests

pstjohn:
Not sure we support changing configs like this, but this would fail without handling the non-swiglu variant

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixes feed‑forward sizing for non‑SwiGLU activations and prevents initialization/runtime errors, improving stability across activation options.
  - Note: forward method signatures were simplified (removed variable kwargs), which changes the public forward API.

- **Tests**
  - Adds deterministic masking seed and four loss/regression tests validating model behavior for pretrained and reinitialized variants.

- **Chores**
  - Updates development container setup and install/run configuration for the project.

### DIFF
--- a/models/amplify/tests/test_amplify_model.py
+++ b/models/amplify/tests/test_amplify_model.py
@@ -168,3 +168,52 @@ def test_convert_state_dict():
     te_state_dict_keys.remove("decoder.bias")
 
     assert len(te_state_dict_keys) == 0
+
+
+def test_hf_trained_model_loss(input_data):
+    model = amp_hf.AMPLIFY.from_pretrained("chandar-lab/AMPLIFY_120M")
+    model.to("cuda", dtype=torch.bfloat16)
+    input_data = {k: v.to("cuda") for k, v in input_data.items()}
+    model.eval()
+    with torch.amp.autocast(device_type="cuda", dtype=torch.bfloat16):
+        output = model(**input_data)
+
+    torch.testing.assert_close(output.loss.detach().cpu(), torch.tensor(2.4), atol=1e-1, rtol=1e-2)
+
+
+def test_te_trained_model_loss(input_data):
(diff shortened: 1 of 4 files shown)


models/amplify/tests/test_amplify_model.py
```diff
@@ -168,3 +168,52 @@ def test_convert_state_dict():
     te_state_dict_keys.remove("decoder.bias")
 
     assert len(te_state_dict_keys) == 0
+
+
```
jomitchellnv, line 172: needs gpu pytest mark parametrize

models/amplify/tests/test_amplify_model.py
```diff
@@ -168,3 +168,52 @@ def test_convert_state_dict():
...
+    model.to("cuda", dtype=torch.bfloat16)
+    input_data = {k: v.to("cuda") for k, v in input_data.items()}
+    model.eval()
+    with torch.amp.autocast(device_type="cuda", dtype=torch.bfloat16):
+        output = model(**input_data)
+
+    torch.testing.assert_close(output.loss.detach().cpu(), torch.tensor(2.4), atol=1e-1, rtol=1e-2)
+
+
+def test_te_trained_model_loss(input_data):
```
jomitchellnv, line 184: needs gpu

jomitchellnv (approved):
Coderabbit is awesome. What was the error with KWargs and Amplify?

models/amplify/tests/test_amplify_model.py
pstjohn: Everything needs a GPU currently, let's go back and annotate if we end up having a split CI where we can test some of this on CPU only workers

pstjohn:
> Coderabbit is awesome. What was the error with KWargs and Amplify? 

If you had kwargs in your models forward pass, accelerate will scale your loss by the number of processes per node.

PR #1149, opened 2025-09-16
Add THD input format support to ESM-2 model

pstjohn:
Allows our ESM-2 model to natively support THD inputs via

```python
AutoModel.from_pretrained(
  "nvidia/esm2_t6_8M_UR50D",
  trust_remote_code=True, 
  attn_input_format="thd"
)
```

also adds golden value tests to ensure the outputs from BSHD and THD attention formats are equivalent


models/esm2/src/esm/collator.py
```diff
@@ -0,0 +1,181 @@
...
+    def __init__(
+        self,
+        # DataCollatorForLanguageModeling
+        tokenizer: PreTrainedTokenizerBase,
+        mlm: bool = True,
+        mlm_probability: float | None = 0.15,
+        mask_replace_prob: float = 0.8,
+        random_replace_prob: float = 0.1,
+        pad_to_multiple_of: int | None = None,
+        tf_experimental_compile: bool = False,
```
jomitchellnv, line 92: whats the TF line?

models/esm2/tests/test_thd_inputs.py
```diff
@@ -0,0 +1,95 @@
...
+
+    input_data_bhsd = {k: v.to("cuda") for k, v in input_data_bhsd.items()}
+    input_data_thd = {k: v.to("cuda") if isinstance(v, torch.Tensor) else v for k, v in input_data_thd.items()}
+
+    bshd_outputs = model_bshd(**input_data_bhsd)
+    thd_outputs = model_thd(**input_data_thd)
+
+    print("bshd_outputs.loss", bshd_outputs.loss)
+    print("thd_outputs.loss", thd_outputs.loss)
+    torch.testing.assert_close(bshd_outputs.loss, thd_outputs.loss, atol=1e-2, rtol=1e-2)
```
jomitchellnv: how close are they? Does the seed actually work in the collator? You could potentially split up this test and remove the data part and check that they are a near perfect match right?

pstjohn: good catch, this flagged an issue with the built-in DataCollatorWithFlattening. 

This line: https://github.com/huggingface/transformers/blob/v4.56.1/src/transformers/data/data_collator.py#L2028
```
                batch["labels"] += [separator_id] + sample["labels"][1:]
```
means that if we masked the first label entry, that value wouldn't participate in the loss calculation.

but i actually don't do any masking here; i just make the model do the identity task of predicting each label based on it's input value. The seed for the collators _does_ work, but it behaves different for the flattened vs non-flattened collator. So this is close enough without introducing randomness from the masking

pstjohn:
@jomitchellnv i'm force-merging to unblock the folder migration in #1151, let's re-open any issues you find for a later PR

PR #1150, opened 2025-09-16
nit: fix amplify config

pstjohn:
Update amplify L0 test config to use the amplify LR schedule

## Summary by CodeRabbit

- Chores
  - Updated the default configuration for the ESM2 Accelerate recipe to use the amplify profile, so runs without overrides will now pick up the amplify defaults.
  - Other parameters (training steps, dataset, trainer) remain unchanged aside from the selected defaults profile.
  - This change is limited to the ESM2 Accelerate recipe and does not affect other recipes.


jomitchellnv (approved):
yessir

PR #1160, opened 2025-09-17
add mxfp8 padding and tests for esm-2

pstjohn:
Adds padding to the vocab dimension of ESM-2 to make it compatible with FP8 and MXFP8 kernels

## Summary by CodeRabbit

- **New Features**
  - FP8/MXFP8 execution support with parity checks; configurable padded vocabulary (padded_vocab_size, default 64) enabling FP8-friendly embeddings/LM outputs while truncating outputs back to the original vocab size.
- **Refactor**
  - Simplified, PyTorch-centric data-collation pipeline; collators enforce PyTorch tensors and support padding to a specified multiple.
- **Documentation**
  - Config docs updated to describe padded_vocab_size and its constraints.
- **Tests**
  - Added FP8/MXFP8 forward-pass tests, collator/unit tests, and state-dict padding validation; test fixtures reorganized for a shared converted checkpoint.

### DIFF
--- a/bionemo-recipes/models/esm2/src/esm/collator.py
+++ b/bionemo-recipes/models/esm2/src/esm/collator.py
@@ -19,8 +19,9 @@ This should eventually get moved to a separate package, or possibly upstreamed i
 """
 
 from dataclasses import dataclass
+from typing import Any
 
-import numpy as np
+import torch
 from transformers import DataCollatorForLanguageModeling, DefaultDataCollator, PreTrainedTokenizerBase
 
 
@@ -59,7 +60,6 @@ class MLMDataCollatorWithFlattening:
         >>> collator = MLMDataCollatorWithFlattening(
         ...     tokenizer=tokenizer,
         ...     mlm_probability=0.15,
-        ...     return_flash_attn_kwargs=True,
         ... )
         >>>
         >>> # Process batch
@@ -89,12 +89,10 @@ class MLMDataCollatorWithFlattening:
         mask_replace_prob: float = 0.8,
         random_replace_prob: float = 0.1,
         pad_to_multiple_of: int | None = None,
+        label_pad: int = -100,
         tf_experimental_compile: bool = False,
         return_tensors: str = "pt",
         seed: int | None = None,
-        # DataCollatorWithFlattening
-        return_flash_attn_kwargs=True,
-        return_seq_idx=False,
     ):
         """Initialize the MLMDataCollatorWithFlattening."""
         self.mlm_collator = DataCollatorForLanguageModeling(
@@ -108,12 +106,9 @@ class MLMDataCollatorWithFlattening:
             return_tensors=return_tensors,
             seed=seed,
         )
-        self.flattening_collator = DataCollatorWithFlattening(
-            return_flash_attn_kwargs=return_flash_attn_kwargs,
-            return_seq_idx=return_seq_idx,
-            return_tensors=return_tensors,
-        )
         self.return_tensors = return_tensors
+        self.pad_to_multiple_of = pad_to_multiple_of
+        self.label_pad = label_pad
 
     def __call__(self, features, return_tensors=None):
         """Process a batch of variable-length sequences for Flash Attention with MLM.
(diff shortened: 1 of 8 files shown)


bionemo-recipes/models/esm2/src/esm/collator.py
```diff
@@ -130,205 +153,197 @@ def __call__(self, features, return_tensors=None):
...
+            [
+                batch["cu_seq_lens_q"],
+                torch.tensor([batch["cu_seq_lens_q"][-1] + remainder], dtype=batch["cu_seq_lens_q"].dtype),
+            ],
+            dim=0,
+        )
+        batch["cu_seq_lens_k"] = batch["cu_seq_lens_q"]
+
+    if "max_length_q" in batch:
+        batch["max_length_q"] = max(batch["max_length_q"], remainder)
```
jomitchellnv, line 341: So the `remainder` is a separate entry in the input_ids right? So that's why we check to see if the len(remainder) is > len(current_max_length) here right?

bionemo-recipes/models/esm2/src/esm/collator.py
```diff
@@ -130,205 +153,197 @@ def __call__(self, features, return_tensors=None):
...
+        batch["attention_mask"] = torch.tensor(
+            [[v for sample in features for v in sample["attention_mask"]]], dtype=torch.int64
+        )
+    return batch
+
+
+def _pt_pad_to_multiple_of(batch: dict[str, Any], pad_to_multiple_of: int, token_pad: int, label_pad: int):
+    """Pad a batch to a multiple of pad_to_multiple_of.
+
+    Appends a mock sequence to the end of the batch with the given token_pad and label_pad to make the total number of
```
jomitchellnv, line 303: It adds multiple mock sequences right? One to the `input_ids`, another to the `labels`. 

Also, what about the positional_ids? We need to add them as well, since now we have no positional_ids for the new sequence right?

bionemo-recipes/models/esm2/src/esm/collator.py
```diff
@@ -130,205 +153,197 @@ def __call__(self, features, return_tensors=None):
...
+    cu_seq_lens[1:] = torch.cumsum(torch.tensor(sample_lengths), dim=0, dtype=torch.int32)
+    batch["cu_seq_lens_q"] = batch["cu_seq_lens_k"] = cu_seq_lens
+    if "attention_mask" in features[0]:
+        batch["attention_mask"] = torch.tensor(
+            [[v for sample in features for v in sample["attention_mask"]]], dtype=torch.int64
+        )
+    return batch
+
+
+def _pt_pad_to_multiple_of(batch: dict[str, Any], pad_to_multiple_of: int, token_pad: int, label_pad: int):
```
jomitchellnv, line 300: It's probably cleaner if we pass the actual arguments such as `input_ids`, `labels`, etc instead of the `batch` dictionary. I can move this into TransformerEngine utilities later if we want also.

bionemo-recipes/models/esm2/src/esm/convert.py
```diff
@@ -114,3 +111,40 @@ def _pack_qkv_bias(ctx: io.TransformCTX, query, key, value):
     concat_biases = concat_biases.transpose(0, 1).contiguous()
     concat_biases = concat_biases.view(*input_shape)
     return concat_biases
+
+
+def _pad_weights(ctx: io.TransformCTX, source_embed):
```
jomitchellnv, line 116: missing docstring about what source_embed is. It looks like we're padding the vocab size here, i'm not sure why though. Isn't the vocab size just for looking up values but doesn't directly affect the intermediate shape?

bionemo-recipes/models/esm2/src/esm/convert.py
```diff
@@ -114,3 +111,40 @@ def _pack_qkv_bias(ctx: io.TransformCTX, query, key, value):
...
+    source_key="lm_head.decoder.weight",
+    target_key="lm_head.decoder.weight",
+)(_pad_weights)
+
+
+@io.state_transform(
+    source_key="lm_head.bias",
+    target_key="lm_head.decoder.bias",
+)
+def _pad_bias(ctx: io.TransformCTX, source_bias):
```
jomitchellnv, line 142: not sure what's going on here or why this is necessary

pstjohn: that final linear layer dimension needs to be divisible by 64 for mxfp8; but the original bias is only 33. So we need to pad it out with zeros

bionemo-recipes/models/esm2/src/esm/collator.py
pstjohn: we add a single mock sequence (e.g., protein sequence), that has both input_ids and labels. 
we don't currently return positional_ids. Do we need them?

bionemo-recipes/models/esm2/src/esm/collator.py
pstjohn: "remainder" is the length of the pseudo-sequence we're adding

bionemo-recipes/models/esm2/src/esm/convert.py
pstjohn: it's tied to the decoder weights

PR #1165, opened 2025-09-19
update devcontainer for recipes

pstjohn:
Addresses some of @savitha-eng's onboarding issues with the recipe devcontainer

## Summary by CodeRabbit

- New Features
  - Dev container now includes the uv tool and uses it to install pre-commit (with guard to only install hooks inside a git repo).
  - Automatic initialization creates caches, SSH/GPG directories, and history/netrc files if missing.
  - Adds host-mounted read-only SSH keys and GnuPG data for secure in-container use.

- Chores
  - Establishes PRE_COMMIT_HOME for isolated caching.
  - Keeps existing bash-history mount behavior with minor formatting updates.

### DIFF
--- a/.devcontainer/recipes/devcontainer.json
+++ b/.devcontainer/recipes/devcontainer.json
@@ -8,14 +8,20 @@
     "mounts": [
         "source=${localEnv:HOME}/.cache,target=/home/ubuntu/.cache,type=bind,consistency=cached",
         "source=${localEnv:HOME}/.netrc,target=/home/ubuntu/.netrc,type=bind,consistency=cached",
-        "source=${localEnv:HOME}/.bash_history_devcontainer,target=/home/ubuntu/.bash_history,type=bind,consistency=cached"
+        "source=${localEnv:HOME}/.bash_history_devcontainer,target=/home/ubuntu/.bash_history,type=bind,consistency=cached",
+        "source=${localEnv:HOME}/.ssh,target=/home/ubuntu/.ssh,readonly,type=bind,consistency=cached",
+        "source=${localEnv:HOME}/.gnupg,target=/home/ubuntu/.gnupg,type=bind,consistency=cached"
     ],
     "postCreateCommand": ".devcontainer/recipes/postCreateCommand.sh",
+    "initializeCommand": ".devcontainer/recipes/initializeCommand.sh",
     "remoteUser": "ubuntu",
     "runArgs": [
         "--gpus=all",
         "--shm-size=4g"
     ],
+    "containerEnv": {
+        "PRE_COMMIT_HOME": "/home/ubuntu/.cache/pre-commit-devcontainer"
+    },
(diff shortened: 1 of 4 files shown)


.devcontainer/recipes/devcontainer.json
```diff
@@ -8,14 +8,20 @@
     "mounts": [
         "source=${localEnv:HOME}/.cache,target=/home/ubuntu/.cache,type=bind,consistency=cached",
         "source=${localEnv:HOME}/.netrc,target=/home/ubuntu/.netrc,type=bind,consistency=cached",
-        "source=${localEnv:HOME}/.bash_history_devcontainer,target=/home/ubuntu/.bash_history,type=bind,consistency=cached"
+        "source=${localEnv:HOME}/.bash_history_devcontainer,target=/home/ubuntu/.bash_history,type=bind,consistency=cached",
```
jomitchellnv, line 11: what does the bashhistory do here?

.devcontainer/recipes/devcontainer.json
```diff
@@ -8,14 +8,20 @@
...
+        "source=${localEnv:HOME}/.ssh,target=/home/ubuntu/.ssh,readonly,type=bind,consistency=cached",
```
jomitchellnv, line 12: are we doing this so we can use git inside the containers? I don't think we need this do we? The $USER of those git commits will be `ubuntu`

.devcontainer/recipes/devcontainer.json
pstjohn: It means it preserves your command history across devcontainer sessions

.devcontainer/recipes/devcontainer.json
jomitchellnv: Since we don't ship dev containers its prob not a security issue, but I don't think we need this....

.devcontainer/recipes/devcontainer.json
```diff
@@ -8,14 +8,20 @@
...
     ],
     "postCreateCommand": ".devcontainer/recipes/postCreateCommand.sh",
+    "initializeCommand": ".devcontainer/recipes/initializeCommand.sh",
     "remoteUser": "ubuntu",
     "runArgs": [
         "--gpus=all",
         "--shm-size=4g"
     ],
+    "containerEnv": {
+        "PRE_COMMIT_HOME": "/home/ubuntu/.cache/pre-commit-devcontainer"
```
jomitchellnv, line 23: this adds pre-commit commands into the container? That's helpful for sure 👍

jomitchellnv (approved):

.devcontainer/recipes/devcontainer.json
pstjohn: yeah i should test whether it brings over your global git config. might be possible to do that without the user being ubuntu

PR #1184, opened 2025-09-23
Consolidate perf logging and add final averaged metrics output

pstjohn:
Updates the esm2_native_te recipe to use a centralized perf logger class, moves wandb handling to this logger, and prints out final tokens/sec, non-padded tokens/sec, and mean step time at the end of training.

## Summary by CodeRabbit

- New Features
  - Integrated a performance logger with per-step metrics and progress.
  - Added configurable checkpoint saving (save_checkpoints, save_final_model).
  - Introduced logger tuning for step-time window.

- Changes
  - Restructured configs: checkpoint settings consolidated under checkpoint; updated defaults (resume enabled, save interval reduced).
  - Final model now saved under the checkpoint path.

- Refactor
  - Unified checkpointing and resume across DDP/mFSDP/FSDP2, preserving scheduler state and simplifying flows.

- Tests
  - Added extensive distributed checkpointing coverage and automatic cleanup between tests.


bionemo-recipes/recipes/esm2_native_te/checkpoint.py
```diff
@@ -127,52 +153,90 @@ def save_final_model_ddp(
...
+        "scheduler": scheduler.state_dict(),
+        "metadata": {},
+    }
+    torch.distributed.checkpoint.load(state_dict=ckpt_state_dict, checkpoint_id=checkpoint_path)
 
-        ckpt_state_dict = {"model": model.state_dict(), "optimizer": optimizer.state_dict()}
-        torch.distributed.checkpoint.load(state_dict=ckpt_state_dict, checkpoint_id=str(checkpoint_path))
+    model.load_state_dict(ckpt_state_dict["model"])
+    optimizer.load_state_dict(ckpt_state_dict["optimizer"])
+    scheduler.load_state_dict(ckpt_state_dict["scheduler"])
```
jomitchellnv, line 185: so it just loads the scheduler and we don't need to set `scheduler.step` anywhere not right?

pstjohn: yeah that's right, this should be saved in the scheduler's state_dict.

ultimately we probably want "stop and go"-style tests like we have in bionemo2 here. Make sure training curves and logits are identical

PR #1189, opened 2025-09-24
add position ids to esm2 collator

pstjohn:
Adds back the position_ids to the esm2 collator with tests

## Summary by CodeRabbit

- New Features
  - Added an optional return_position_ids flag to data collators to include position_ids in output batches.
  - Position IDs reflect original token positions across flattening and are preserved/aligned when padding to a multiple of a value.
  - Supported in both standard and MLM collator flows.

- Tests
  - Expanded test coverage to validate position_ids generation and alignment across basic, padded, and MLM scenarios.

### DIFF
--- a/bionemo-recipes/models/esm2/src/esm/collator.py
+++ b/bionemo-recipes/models/esm2/src/esm/collator.py
@@ -107,6 +107,7 @@ class MLMDataCollatorWithFlattening:
         return_tensors: str = "pt",
         seed: int | None = None,
         pad_to_multiple_of: int | None = None,
+        return_position_ids: bool = False,
     ):
         """Initialize the MLMDataCollatorWithFlattening.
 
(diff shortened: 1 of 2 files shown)


bionemo-recipes/models/esm2/src/esm/modeling_esm_te.py
```diff
@@ -77,6 +78,7 @@ def __init__(
                 `b` batch size, `h` the number of heads, `d` head size. Note that these
                 formats are very closely related to the `qkv_format` in the
                 `MultiHeadAttention` and `DotProductAttention` modules.
+            attn_mask_type: Type of attention mask passed into softmax operation for encoder.
```
jomitchellnv: missing
```
self.attn_mask_type=attn_mask_type
```
below

jomitchellnv (approved):

PR #1219, opened 2025-10-03
Evo2 Gradient Tests

gagank1:
### Description

Adds gradient comparison tests to Evo2 to accurately recognize regressions. It runs several equivalent variants of the same model, and checks to see if the resulting gradients are the same as expected.

All tested configurations match the 1 GPU base case.

```
1: DP=1, TP=1, CP=1, PP=1 (baseline on 1 GPU that everything else is compared to)
2: DP=2, TP=1, CP=1, PP=1
3: DP=1, TP=2, CP=1, PP=1
4: DP=1, TP=1, CP=2, PP=1
5: DP=1, TP=1, CP=1, PP=2
```

Rather than a standard `approx_equal` on the two gradient tensors, this implements the recommended similarity formula mentioned in the ttrace paper https://www.arxiv.org/pdf/2506.09280

#### Usage

```bash
pytest --dataset-dir=<training data dir> --training-config=<training data config yaml> sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py::test_distributed_training_gradient_equivalence
```

### Type of changes

- [ ] Refactor
- [ ] Documentation update
- [x] Other (please describe): New test type


sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py
```diff
@@ -17,18 +17,21 @@
 # limitations under the License.
```
jstjohn, line 17: Can you move these into a new test file? Let's call it `test_distributed_gradients.py` maybe?

sub-packages/bionemo-testing/src/bionemo/testing/assert_optimizer_grads_match.py
```diff
@@ -0,0 +1,348 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jstjohn, line 1: Since for now this is only used inside of the new tests, maybe just add this all to the same test file `test_distributed_gradients.py` if that's what you call it? You can keep the:

```
def main():
    parser = ArgumentParser(
        description="Given checkpoints saved with adam b1,b2=0 trained for one step, "
        "we can check that the gradients match under different training configurations. "
        "Currently this test script has some hard-coded assumptions for GPT style models, "
        "namely which layers are RowParallel and require different unsharding logic."
    )
    parser.add_argument("checkpoints", nargs="+", type=Path, help="Path to the checkpoints to compare")
    args = parser.parse_args()
    _test_optimizer_states_match(args.checkpoints)


if __name__ == "__main__":
    main()
```

At the bottom incase the user wants to run it the way you are able to on a set of checkpoints. If we decide we want to make it more accessible we can move this out into an `assert_evo2_optimizer_grads_match.py` and maybe put it in the `run` directory or something. For now this test is the only user so I don't think it matters. You can also leave this file in the testing directory but name it `assert_evo2_optimizer_grads_match.py` so that pytest doesn't look for tests inside of the file? Any of these options seems reasonable to me.

sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py
jstjohn: Actually my recommendation breaks our testing recommendation. I think you can leave it where it is, just rename the other file to not have "test" in the prefix, like `assert_...` or something.

sub-packages/bionemo-testing/src/bionemo/testing/assert_optimizer_grads_match.py
jstjohn: Actually I didn't notice that you already did what I was suggesting with putting it into it's own spot. Just rename to `assert_` rather than `test_` and I'm happy.

gagank1: done

sub-packages/bionemo-evo2/tests/bionemo/evo2/run/test_train.py
gagank1: done

sub-packages/bionemo-testing/src/bionemo/testing/assert_optimizer_grads_match.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+def _fro_norm(x: TensorLike) -> torch.Tensor:
+    """Frobenius norm; supports sharded tensors (sum of shard ||·||_F^2)."""
+    it = _as_iter(x)
+    s = torch.tensor(0.0, device=next(iter(it)).device if it else "cpu")
+    for t in it:
+        s = s + t.float().pow(2).sum()
+    return torch.sqrt(s)
+
```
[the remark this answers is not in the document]
gagank1: fixed

PR #1221, opened 2025-10-03
Updates to ESM-2 for better THD input support

pstjohn:
Fixes a number of small issues with ESM-2 identified after THD convergence testing at scale:

1. fixes the "token dropout" logic to correctly handle THD batches and numbers of tokens per batch
2. ensures that we don't use torch.amp.autocast when calculating rope embeddings during the forward pass

## Summary by CodeRabbit

- New Features
  - Embeddings now support high-throughput input formats and rotary position embeddings.
  - Token dropout (masking/scale), optional layer normalization, and attention masking added.
  - Position-embedding computation moved to a runtime path with correct device/dtype and autocast handling.

- Tests
  - Expanded THD vs non-THD coverage for outputs, hidden states, logits, and gradients across backends and autocast.
  - Added backend-switching fixture and new embedding tests; test padding changed to multiples of 32.

### DIFF
--- a/bionemo-recipes/models/esm2/src/esm/modeling_esm_te.py
+++ b/bionemo-recipes/models/esm2/src/esm/modeling_esm_te.py
@@ -39,7 +39,7 @@ from transformers.modeling_outputs import (
 )
 from transformers.modeling_utils import PreTrainedModel
 from transformers.models.esm.configuration_esm import EsmConfig
-from transformers.models.esm.modeling_esm import EsmEmbeddings, EsmPooler
+from transformers.models.esm.modeling_esm import EsmPooler
 from transformers.utils import logging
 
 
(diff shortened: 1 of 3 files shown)


bionemo-recipes/models/esm2/src/esm/modeling_esm_te.py
```diff
@@ -195,35 +185,22 @@ def forward(
...
-            elif self.config.attn_input_format == "thd":
-                assert cu_seq_lens_q is not None
-                te_rope_emb = self.rotary_embeddings(max_seq_len=cu_seq_lens_q[-1]).to(
-                    device=hidden_states.device, dtype=hidden_states.dtype, non_blocking=True
-                )
-
-            else:
-                raise ValueError(f"Unsupported attention input format: {self.config.attn_input_format}")
+        # Ensure that rotary embeddings are computed with at a higher precision outside the torch autocast context.
+        with torch.autocast(device_type="cuda", enabled=False):
```
pstjohn, line 197: this was the key line. we should honestly upstream this inside the forward pass of `RotaryPositionEmbedding`

jomitchellnv, line 197: Whats the point of having autocast if enabled=false?

pstjohn: This explicitly turns it off -- if this forward pass is called in a parent autocast loop, these steps will be executed without it

jomitchellnv: got it

PR #1231, opened 2025-10-07
use torch.compile in esm2_native_te

pstjohn:
Adds support for (and enables by default for ddp and fsdp2) `torch.compile`, ensuring this works correctly with checkpointing and distributed training.

## Summary by CodeRabbit

* **New Features**
  * Added a use_torch_compile configuration flag to optionally enable torch.compile.
  * When enabled, models are compiled prior to resuming training for DDP and FSDP2 runs.

* **Documentation**
  * Warns users that torch.compile is not supported with M-FSDP; a runtime warning is shown if enabled with that strategy.

### DIFF
--- a/bionemo-recipes/recipes/esm2_native_te/train_mfsdp.py
+++ b/bionemo-recipes/recipes/esm2_native_te/train_mfsdp.py
@@ -117,7 +117,10 @@ def main(args: DictConfig) -> float | None:
         fp8_recipe = None
 
     if args.use_torch_compile:
-        logger.warning("BIONEMO-2977: Using torch.compile with mfsdp is currently not supported.")
+        logger.warning(
+            "BIONEMO-2977: Using torch.compile with mfsdp is currently not supported. `use_torch_compile` was set to "
+            "true, but will be ignored."
+        )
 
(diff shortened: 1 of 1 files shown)


bionemo-recipes/recipes/esm2_native_te/train_mfsdp.py
```diff
@@ -116,6 +116,9 @@ def main(args: DictConfig) -> float | None:
     else:
         fp8_recipe = None
 
+    if args.use_torch_compile:
+        logger.warning("BIONEMO-2977: Using torch.compile with mfsdp is currently not supported.")
```
trvachov: Is it gonna error out? Or...what's gonna happen, "warning" then run with bad numerics?

pstjohn: it just doesn't do anything, we don't call into `torch.compile` in this script. Mainly just leaving ourselves a breadcrumb here to investigate and fix.

trvachov: Ah ok -- let's just add that to the warning, "You turned on torch.compile, but it's not supported and wont be used".

trvachov (approved):
approved pending slightly better warning for mFSDP + torch.compile

PR #1257, opened 2025-10-15
add token packing dataset

pstjohn:
Adds the capability to dynamically change the batch size in THD model training, packing batches up to a desired number of total tokens (rather than individual sequences).

Pads up to a total number of tokens for consistent array dimensions rather than splitting partial sequences.

## Summary by CodeRabbit

* **New Features**
  * Optional token-packing mode for higher token-per-batch efficiency with a dedicated token-packed data path.

* **Behavior**
  * Data loading supports both standard and token-packed flows and returns tokenized data plus appropriate iterables/samplers.
  * Collator now pads packed sequences to the configured max tokens per batch.
  * Improved handling for distributed vs. streaming datasets.

* **Configuration**
  * Introduced top-level use_sequence_packing flag; CLI/examples updated.

* **Tests**
  * Added unit and sanity tests for token-packing and end-to-end scenarios.

### DIFF
--- a/bionemo-recipes/recipes/esm2_native_te/dataset.py
+++ b/bionemo-recipes/recipes/esm2_native_te/dataset.py
@@ -22,47 +22,22 @@ from torchdata.stateful_dataloader import StatefulDataLoader
 from transformers import AutoTokenizer
 from transformers.data.data_collator import DataCollatorForLanguageModeling
 
-from collator import MLMDataCollatorWithFlattening
+from collator import MLMDataCollatorWithFlattening, TokenPackingDataset
 from distributed_config import DistributedConfig
 
 
 logger = logging.getLogger(__name__)
 
 
-def create_dataloader(
+def create_tokenized_dataset(
     distributed_config: DistributedConfig,
     tokenizer_name: str,
     load_dataset_kwargs: dict,
-    micro_batch_size: int,
-    num_workers: int,
     max_seq_length: int = 1024,
-    seed: int = 42,
-    use_sequence_packing: bool = False,
-    sequence_packing_pad_to_multiple_of: int | None = None,
     buffer_size: int = 10_000,
     use_lazy_tokenization: bool = True,
-    mlm_probability: float = 0.15,
 ):
-    """Create a dataloader for the dataset.
-
-    Args:
-        distributed_config: The distributed configuration.
-        tokenizer_name: The name of the tokenizer to pull from the HuggingFace Hub.
-        load_dataset_kwargs: Keyword arguments to pass to `load_dataset` for the train dataset.
-        micro_batch_size: The batch size per device.
-        num_workers: The number of workers to use for the dataloader.
-        max_seq_length: The maximum length of the protein sequences.
-        seed: The seed to use for the distributed sampler and data collator.
-        use_sequence_packing: Whether to use sequence packing.
-        sequence_packing_pad_to_multiple_of: The padding to use for the sequence packing collator, for fp8 support.
-        buffer_size: The buffer size to use for the distributed sampler.
-        use_lazy_tokenization: Whether to use datasets.set_transform for tokenization if the dataset is a
-            non-streaming datasets.Dataset. Defaults to True.
-        mlm_probability: The probability of masking tokens for MLM (default 0.15). Set to 0 for no masking.
-
-    Returns:
-        A dataloader that can be used for training.
-    """
+    """Create a tokenized dataset."""
     logger.info(f"Loading dataset with kwargs: {load_dataset_kwargs}")
     dataset = datasets.load_dataset(**load_dataset_kwargs)
     logger.info(f"Loaded dataset: {dataset}")
@@ -74,14 +49,6 @@ def create_dataloader(
             world_size=distributed_config.world_size,
         )
         dataset = dataset.shuffle(seed=42, buffer_size=buffer_size)
-        sampler = None
-    else:
-        sampler = DistributedSampler(
-            dataset,
-            rank=distributed_config.rank,
-            num_replicas=distributed_config.world_size,
-            seed=seed,
-        )
 
     tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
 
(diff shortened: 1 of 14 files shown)


bionemo-recipes/recipes/esm2_native_te/dataset.py
```diff
@@ -42,23 +44,28 @@ def create_dataloader(
...
+    use_token_packing: bool = False,
 ):
     """Create a dataloader for the dataset.
 
     Args:
         distributed_config: The distributed configuration.
         tokenizer_name: The name of the tokenizer to pull from the HuggingFace Hub.
         load_dataset_kwargs: Keyword arguments to pass to `load_dataset` for the train dataset.
-        micro_batch_size: The batch size per device.
+        micro_batch_size: The batch size per device. If use_token_packing is True, this is the maximum number of tokens
```
jomitchellnv: we should use a different variable name for token_packing. it's conflicting

bionemo-recipes/recipes/esm2_native_te/dataset.py
```diff
@@ -109,7 +116,8 @@ def tokenize_function(examples):
         data_collator = MLMDataCollatorWithFlattening(
             tokenizer=tokenizer,
             mlm_probability=mlm_probability,
-            pad_to_multiple_of=sequence_packing_pad_to_multiple_of,
+            # When we're using token packing, we want to pad to the maximum number of tokens per batch.
+            pad_to_multiple_of=sequence_packing_pad_to_multiple_of if not use_token_packing else micro_batch_size,
```
jomitchellnv: with token packing, we still need to pad to the multiple of, if the token packing number is not that divisible by the required divisiblity factory

bionemo-recipes/recipes/esm2_native_te/dataset.py
```diff
@@ -120,14 +128,49 @@ def tokenize_function(examples):
...
+
+@dataclass
+class TokenPackingDataset(torch.utils.data.IterableDataset):
+    """Dataset that uses sequence packing to construct batches with variable length up to a maximum number of tokens."""
+
+    dataset: datasets.IterableDataset
+    max_length: int
+
+    def __iter__(self):
+        """Yield batches of samples, each with a variable number of tokens up to the maximum length.
```
jomitchellnv: If, let's say you want `token_micro_batch_size=100` and you have already packed 90 tokens from 9 different samples of length 10 each, now the next sample in line is length 15, which would 90+15=105 > 100. Would you simply avoid packing the next sample in line because it will exceed the token_micro_batch_size, or would you partially pack the first 10 tokens, or would you find another sample of length 10 to pack instead?

pstjohn: exactly -- we avoid packing the next sample, yield the batch of 9 sequences, and start packing the next batch with that sample (we now have 15 tokens in it).

That length-9 list will then get passed into the DataCollatorWithFlattening, which will have `pad_to_multiple_of` set to 100. So we'll create some mock sequence of length 10 that's all mask tokens to ensure the T dimension is always 100.

bionemo-recipes/recipes/esm2_native_te/dataset.py
pstjohn: i answer this in the next question but this is just to ensure that array dimensions stay consistent, even when we're not packing partial sequences.

bionemo-recipes/recipes/esm2_native_te/dataset.py
jomitchellnv: Perfect that's what I wanted -- thanks for the clarity

jomitchellnv: Do we have a unit test to demonstrate this behavior btw?

bionemo-recipes/recipes/esm2_native_te/dataset.py
jomitchellnv: ack - resolving

bionemo-recipes/recipes/esm2_native_te/hydra_config/defaults.yaml
```diff
@@ -16,6 +16,7 @@ dataset:
   num_workers: 1
   max_seq_length: 1024
   use_sequence_packing: false
+  use_token_packing: false
```
jstjohn: why not make this true? It seems like what a user would _usually_ want unless they're doing some kind of test.

jstjohn: oh this is because you are doing BSHD for this example?

bionemo-recipes/recipes/esm2_native_te/dataset.py
```diff
@@ -120,14 +128,49 @@ def tokenize_function(examples):
...
 
     return train_dataloader, dataset if sampler is None else sampler
+
+
+@dataclass
+class TokenPackingDataset(torch.utils.data.IterableDataset):
+    """Dataset that uses sequence packing to construct batches with variable length up to a maximum number of tokens."""
+
+    dataset: datasets.IterableDataset
+    max_length: int
```
jstjohn: so max length here is the maximum number of tokens per batch? Eg if you were thinking about `BSHD` with no padding it would be `B * S`?

jstjohn: If so, `max_tokens_per_batch` is probably overly verbose but more clear. `max_tokens` maybe?

jstjohn: Oh yeah... so this is microbatch size too right?

bionemo-recipes/recipes/esm2_native_te/hydra_config/defaults.yaml
pstjohn: Yeah, I think we can probably deprecate the BSHD-compatible use case outside of our test suite, and just have 2 dataloaders:
1. BSHD for compatibility with the HF checkpoints
2. THD where we pack to some maximum token number

bionemo-recipes/recipes/esm2_native_te/dataset.py
pstjohn: We still have a max sequence length even in THD, but i suppose we could play around with that and see if allowing longer sequences within THD batches blows up our memory. But we need to figure out the right way of having users specify the max tokens per batch

bionemo-recipes/recipes/esm2_native_te/dataset.py
```diff
@@ -130,4 +142,70 @@ def tokenize_function(examples):
...
+        tokenizer_name=tokenizer_name,
+        load_dataset_kwargs=load_dataset_kwargs,
+        max_seq_length=max_seq_length,
+        buffer_size=buffer_size,
+    )
+
+    assert isinstance(tokenized_dataset, datasets.IterableDataset), "THD token packing requires a streaming dataset."
+    if token_micro_batch_size is None:
+        assert micro_batch_size is not None, "Only one of micro_batch_size or token_micro_batch_size can be provided."
+        token_micro_batch_size = micro_batch_size * max_seq_length
```
jomitchellnv, line 191: im ok with this behavior. works nicely because its a THD dataloader

bionemo-recipes/recipes/esm2_native_te/hydra_config/L0_sanity.yaml
```diff
@@ -9,13 +9,12 @@ num_train_steps: 250
...
 use_torch_compile: true
 
+use_sequence_packing: false
 dataset:
   tokenizer_name: ${model_tag}
   micro_batch_size: 2
   num_workers: 1
   max_seq_length: 1024
-  use_sequence_packing: false
-  sequence_packing_pad_to_multiple_of: null
```
jomitchellnv, line 18: are we no longer able to specify what we want it to pad to a multiple of?

bionemo-recipes/recipes/esm2_native_te/dataset.py
jstjohn: > You can do micro_batch_size and it does the B * S multiplication, or you can provide T directly with tokens_micro_batch_size

From the slack conversation around this from Peter. So if you want to think about batch * seq_len you can, and THD internally turns that into tokens by multiplying the two, or you can set T directly through another arg.

bionemo-recipes/recipes/esm2_native_te/tests/test_dataset.py
```diff
@@ -311,12 +308,11 @@ def test_iterable_stateful_dataloader_resumption_multi_process(tmp_path):
     )
 
     # Load rank1 dataloader state at step 4
-    rank1_dataloader_reloaded, _ = create_dataloader(
+    rank1_dataloader_reloaded, _ = create_thd_dataloader(
```
jomitchellnv, line 311: so you just mixed up the tests to confirm that they both work?

bionemo-recipes/recipes/esm2_native_te/tests/test_dataset.py
```diff
@@ -833,3 +829,32 @@ def test_stateful_dataloader_load_fails_if_num_ranks_mismatch(tmp_path, caplog):
         "Dataloader num_workers mismatch: 1 != 1 or num_ranks mismatch: 2 != 1, starting dataloader from scratch."
         in caplog.text
     )
+
+
+def test_token_packing_dataloader():
```
jomitchellnv, line 834: Is there a test that checks if 
1. If the last sequence to be added will push you over the token_batch_size
2. if the last seuqence to be added will make you equal to the token batch size
3. If the last sequence to be added will give you less than the token batch size (like the last sequence in the dataset or something)

bionemo-recipes/recipes/esm2_native_te/dataset.py
jstjohn: My guess is that they are not equivalent since some ops are not linear in memory usage with sequence length. Eg linear layers would scale this way but attention layers would not, and you'd prob OOM in many cases where the same number of tokens is passed through depending on how small the individual sequences are.

jstjohn: So... here is `max_length` like `max_sequence_length`? I think I was just confused thinking about the naming and the T parameter.

bionemo-recipes/recipes/esm2_native_te/tests/test_dataset.py
pstjohn: added these, but to `test_collator.py` in models/esm2

bionemo-recipes/recipes/esm2_native_te/tests/test_dataset.py
pstjohn: yeah pretty much 🤷

bionemo-recipes/recipes/esm2_native_te/hydra_config/L0_sanity.yaml
pstjohn: for sequence packing, no, since we always pad out to the maximum token length.

bionemo-recipes/recipes/esm2_native_te/dataset.py
pstjohn: i'm happy to call it `max_tokens_per_batch`, done

bionemo-recipes/recipes/esm2_native_te/dataset.py
pstjohn: yup, added

bionemo-recipes/recipes/esm2_native_te/dataset.py
pstjohn: So now we say 
```python
micro_batch_size: The batch size (number of sequences) per device. This will set the token_micro_batch_size to
    micro_batch_size * max_seq_length. Defaults to None.
token_micro_batch_size: The maximum number of tokens per batch. If None, the micro_batch_size * max_seq_length
    will be used. Defaults to None.
```

jomitchellnv (approved):
LGTM

pstjohn:
@coderabbitai pause

PR #1263, opened 2025-10-20
Add memory to esm2 perf_logger

jwilber:
Adding some memory usage indicators to perf_logger (avg and peak memory usage).

Originally, we were planning to add some measure of spread here (e.g. variance or std) across memory usage, since different GPUs having different memory profiles was/is an indicator if issue. After thinking about it, I didn't do that for 2 reasons:
1. We can get the same info from just avg and peak: if peak and avg are very close, memory is balanced; if peak is much higher than avg, there's likely an imbalance (or at least cause for investigation). So the two metrics together give enoug sense of whether memory is evenly distributed.
2. Directly calculating var/std will require an all_gather.

## Summary by CodeRabbit

* **New Features**
  * GPU memory allocation metrics (maximum and average) are now collected as part of the regular metrics flow and logged periodically during training.

* **Chores**
  * ESM2 training configuration updated: num_train_steps increased to 20,000 and num_warmup_steps increased to 2,000.


pstjohn (approved):
LGTM, don't gate behind a `track_memory`, just do it automatically

PR #1278, opened 2025-10-27
reduce batch size to fix oom issue

jwilber:
Reduce batch size from `64` to `32` to fix oom issue.

https://dashboard.dgxc-lepton.nvidia.com/workspace/vfco61g2/compute/jobs/list?search=esm2-native-650m-fsdp2-thd&creators=-&status=-&labels=-


trvachov:
Need @pstjohn 's check this wont change convergence results too much.

pstjohn:
We'd need to double the nodes if we want this to be the 2048 GBS baseline

PR #1290, opened 2025-10-28
feat: benchmark `annbatch` in `scspeedtest`

ilan-gold:
### Description

Great suite here!

I added an example for `annbatch` although I'm not exactly sure that's where it should live.

I also tinker with a couple of things, some of which are in the TODOs:

1. ~`annbatch.ZarrSparseDataset` is not based on `torch.utils.data.DataLoader` so `num_workers` doesn't really apply - the loader is threaded, but I think that is different (I added a comment about this).  So I'm not sure how to handle that.  That being said, there are cases we might want to use `DataLoader` (small `block_size`) but we wouldn't use all workers, only maybe a fourth of total available threads or something on this order of magnitude.~  I just let the benchmark does whatever.  I guess we'll do a grid search anyway with different numbers of `num_workers` so it shouldn't matter
2. ~~I think our data loader is faster when you have `chunk_size=1` i.e., perfect randomness, when wrapped in `torch.utils.data.DataLoader`.  But this isn't a requirement - what is interesting is that the `create_collection` function does a lot of unnecessary work in that case.  It shuffles the data so in theory, we could special-case dataset creation for this by just writing the data to disk as zarr v3 anndata~~ I noticed in the other scripts that `SingleCellMemMapDataset` takes in a pre-computed format, not creating one on disk.  Should we do that as well? I am a little confused what the `factory` function does with `dataset` argument.
3. ~~I have noticed you can't vstack a `torch.Tensor`.  I wrote something to get around this using `cupy` but maybe y'all have some suggestions? It seems like a universal limitation but maybe not?~~ I'll just take this as a 2.0 problem for us and use `cupy` when needed.  What I committed appears to work.  It would be great [if torch solved this](https://github.com/pytorch/pytorch/issues/129843#issuecomment-3461842626).  I also noticed you can't pin memory for a sparse tensor? That seems like it would also be good.
4. Speaking of `cupy`, I added it as a dep, but I'm not sure that makes sense.  It gives `ZarrSparseDataset` a performance boost without relying on `torch` (i.e., so the loader could be used with `jax`) but also requires a GPU on the installing machine.  Is this a safe assumption? Probably not, ~~but~~ I added ~~it anyway to make this clear and can change it~~ as an extra
5. Is there interest in "fake memory pressure?" This would allow us to "fake" big data by pre-allocating a buffer to take up some predefined percentage of memory.  This would make the small dataset case itself a bit more informative but also give nice metrics like `available_memory / dataset size` ratio as another axis.  The 25K dataset is probably too small for this to make sense since it is about 245MB in memory and I could see allocating `amount_of_ram - 245MB` bytes and then trying to do anything as flaky but we could do this on a ~1GB or ~2GB size dataset and be able to mock this "big data" behavior.  I've in general had trouble reasoning about physical RAM available accurately using `psutil` so this might be something to hardcode i.e., allcoate N bytes because we know there are physically N bytes available.  I would be interested in this because we support O_DIRECT reading as we've noticed thrashing on some linux machines when the page cache is full: https://annbatch.readthedocs.io/en/latest/zarr-configuration.html#zarrs-performance
6. ~~Where to dump the data? I think I understood right that the on-disk dataset creator i.e., `shared_dataset_factory` should write to disk, but where to? I went with `Path(input).parent` but maybe that was not right? I couldn't quite be sure from the scdataset example since it include a pre-computed `scdl` file, no?~~ I think this is a rehash of point 2.  In general I'm not exactly clear how to create an on-disk dataset.  A new script is needed? Are benchmarking how long it takes to make a dataset?

~~Marking as a draft since (a) `annbatch` is not `0.0.1` yet (so don't want to assume performance of anything quite yet since things might change a bit) and (b)~~ I am not sure what else if anything is needed (tests? I see it says that in the PR checklist but it's not applicable here, I think, although I'd be happy to add tests).

In general ready to go!

#### Usage

This should be in the example, according to the API

### Type of changes

- [ ] Refactor
- [ ] Documentation update
- [x] Other (please describe): Benchmarking

### CI Pipeline Configuration

- [ciflow:skip](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:skip) - Skip all CI tests for this PR
- [ciflow:notebooks](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:notebooks) - Run Jupyter notebooks execution tests for bionemo2
- [ciflow:slow](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:slow) - Run slow single GPU integration tests marked as @pytest.mark.slow for bionemo2
- [ciflow:all](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:all) - Run all tests (unit tests, slow tests, and notebooks) for bionemo2. This label can be used to enforce running tests for all bionemo2.

> [!NOTE]

#### Authorizing CI Runs

### Pre-submit Checklist

### DIFF
--- a/sub-packages/bionemo-scspeedtest/examples/annbatch_script.py
+++ b/sub-packages/bionemo-scspeedtest/examples/annbatch_script.py
@@ -18,6 +18,7 @@
 
 import argparse
 import os
+from collections.abc import Callable, Iterable
 from datetime import datetime
 from pathlib import Path
 
@@ -32,11 +33,13 @@ from bionemo.scspeedtest import benchmark_dataloaders_with_configs
 zarr.config.set({"codec_pipeline.path": "zarrs.ZarrsCodecPipeline"})
 
 
-def create_on_disk_collection(adata_path: Path | str, output_path: Path | None = None):
-    """Creates a collection at either the output_path / "collection" or at the adata_path.parent / "collection" if output_path is None."""
-    output_zarr_collection = (
-        Path(adata_path).parent / "collection" if output_path is None else output_path / "collection"
-    )
+def _get_default_collection_path(adata_path: Path) -> Path:
+    return Path(adata_path).parent / "collection"
+
+
+def create_on_disk_collection(adata_path: Path, *, output_path: Path | None = None) -> Path:
+    """Creates a collection at either the output_path or at the adata_path.parent / "collection" if output_path is None."""
+    output_zarr_collection = _get_default_collection_path(adata_path) if output_path is None else output_path
 
     def load_adata(adata_path):
         adata = ad.experimental.read_lazy(adata_path)
@@ -53,30 +56,30 @@ def create_on_disk_collection(adata_path: Path | str, output_path: Path | None =
         load_adata=load_adata,
         zarr_compressor=None,
     )
-    return output_zarr_collection
+    return Path(output_zarr_collection)
+
+
+def create_adata(collection_path: Path, num_workers: int, *, use_direct_io: bool = False) -> list[ad.AnnData]:
+    """Create a list of AnnData objects loaded from collection_path with the specified parameters, loading only X."""
+    # Allocate each worker an even number of threads plus a little
+    # TODO: How big is the data? Presumably if it is not big enough to fit in memory, we would want O_DIRECT reading to be on.
+    # TODO: There are probably faster ways to get a directory size? If you're using this data loader, presumably your data doesn't fit in memory?
+    # TODO: Does X always contain the genes of interest? Layers? Raw?
+    # TODO: Should num_workers control threading? If scdataset were to wrap zarr, it would still be multithreaded under the processes, so my inclination is "no".
+    with zarr.config.set(
+        {
+            "threading.max_workers": (os.cpu_count() // max(num_workers, 1)) + 1,
+            "codec_pipeline.direct_io": use_direct_io,
+        }
+    ):
+        return [ad.AnnData(X=ad.io.sparse_dataset(zarr.open(p)["X"])) for p in collection_path.iterdir()]
 
 
 def create_dataset_factory(
-    collection_path: Path | str, num_workers: int, *, use_direct_io: bool = False
-) -> ad.AnnData:
+    collection_path: Path, num_workers: int, *, use_direct_io: bool = False
+) -> Callable[[], list[ad.AnnData]]:
     """Generate an `anndata.AnnData` object for use in `annbatch` i.e., zarr v3 on disk, loaded using `anndata.io.sparse_dataset`."""
-    collection_path = Path(collection_path)
-
-    def dataset_factory():
-        # Allocate each worker an even number of threads plus a little
-        # TODO: How big is the data? Presumably if it is not big enough to fit in memory, we would want O_DIRECT reading to be on.
-        # TODO: There are probably faster ways to get a directory size? If you're using this data loader, presumably your data doesn't fit in memory?
-        # TODO: Does X always contain the genes of interest? Layers? Raw?
-        # TODO: Should num_workers control threading? If scdataset were to wrap zarr, it would still be multithreaded under the processes, so my inclination is "no".
-        with zarr.config.set(
-            {
-                "threading.max_workers": (os.cpu_count() // max(num_workers, 1)) + 1,
-                "codec_pipeline.direct_io": use_direct_io,
-            }
-        ):
-            return [ad.AnnData(X=ad.io.sparse_dataset(zarr.open(p)["X"])) for p in collection_path.iterdir()]
-
-    return dataset_factory
+    return lambda: create_adata(collection_path=collection_path, num_workers=num_workers, use_direct_io=use_direct_io)
 
 
 def to_annbatch(
@@ -87,7 +90,7 @@ def to_annbatch(
     block_size: int = 1,
     fetch_factor: int = 2,
     num_workers: int = 0,
-) -> ZarrSparseDataset:
+) -> Iterable:
     """Generate an `annbatch.ZarrSparseDataset` based on configuration and a list of input `anndata.AnnData` objects backed by zarr v3 on disk."""
     if num_workers > 0:
         ds = ZarrSparseDataset(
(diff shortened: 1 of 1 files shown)


ilan-gold:
Ok we released 0.0.1 and I pinned it.  I noticed that `scDataset` has released a `0.2.0` which breaks the tests here -  maybe best to pin that as well while it's unstable?

ilan-gold:
> There is the option of converting it each time vs. just using a created dataset. So for the creation of a SCDL dataset, we have a create_scdl_dataset_and_loader_factory. But we could also split it up into create_scdl_from_anndata (this would create a scdl dataset on disk from adnndata) and create_scdl_dataloader_factory (which would create a dataloader from that dataset)

I looked into this and was a little confused - it looks like there is no "conversion" in the other examples, but rather a `scdl-path` passed in and then it benchmarks the time needed to instantiate the class but not do conversion i.e., `dataset = SingleCellMemMapDataset(data_path)` does no actual on-disk writing, no? So I separated that but maybe I'm wrong.

> The issue here is that we don’t benchmark on GPU currently. It doesn’t look like ZarrSparseDataset explicitly depends on there should be a fallback option that works on CPU.

Ok, and my colleague just noticed we assume you have `cupy` installed by default (i..e, the default settings in `ZarrSparseDataset` use it). So we'll fix that but it's explicitly set to `False` no.  Ideally, when run on a GPU, we would have it installed.

> Point 5. That’s a good idea - I have experimented with adding RAM pressure + running benchmarking. It could be interesting to allow for that pressure within the framework.

Yea, this is the only way for me to develop since I don't have the time/disk space on my linux machine to get the full dataset, but I'm going to change that soon!


Here are the results - I added cache pressure because, like I said, I've only really looked at it with cache pressure.  As you can see, the dataset is 18GB on disk, so about ~100GB uncompressed and I only gave myself 10GB of free RAM. I didn't turn on `direct_io`, but I also don't think it's such a big deal with such a small chunk size (in fact, probably harmful) - my hunch is that it only gets to be useful once the `block_size` goes really big like 512 (which is probably not a good idea unless your data is shuffled on-disk :wink:). Also, very interested in the `num_workers` grid search.  So grid searching over both of those would be super cool i.e., `use_direct_io` on and off, and then `num_workers. 

Here are the results, which broadly make sense to me:
[annbatch_benchmark_20251104_121014_detailed_breakdown.csv](https://github.com/user-attachments/files/23337145/annbatch_benchmark_20251104_121014_detailed_breakdown.csv)

polinabinder1:
Here's an example of data conversion: https://github.com/NVIDIA/bionemo-framework/blob/pbinder/benchmark_conversion_example/sub-packages/bionemo-scspeedtest/examples/scdl_conversion_example.py

Could you add the specs of your machine/ share a comparison to SCDL? Also, could you share the dataset that this is on? We have seen a lot of variability in our benchmarking work based on the machine, so I would be excited to play with this.

ilan-gold:
```bash
lscpu
```
gives
```
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             48 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      16
  On-line CPU(s) list:       0-15
Vendor ID:                   AuthenticAMD
  Model name:                AMD EPYC-Rome Processor
    CPU family:              23
    Model:                   49
    Thread(s) per core:      1
    Core(s) per socket:      1
    Socket(s):               16
    Stepping:                0
    BogoMIPS:                5988.74
    Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm rep_good nopl
                              cpuid extd_apicid tsc_known_freq pni pclmulqdq ssse3 fma cx16 sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lah
                             f_lm cmp_legacy cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw topoext perfctr_core ssbd ibrs ibpb stibp vmmcall fsgsbase tsc_adjust bmi1 avx2 smep bmi
                             2 rdseed adx smap clflushopt clwb sha_ni xsaveopt xsavec xgetbv1 clzero xsaveerptr wbnoinvd arat umip rdpid arch_capabilities
Virtualization features:     
  Hypervisor vendor:         KVM
  Virtualization type:       full
Caches (sum of all):         
  L1d:                       512 KiB (16 instances)
  L1i:                       512 KiB (16 instances)
  L2:                        8 MiB (16 instances)
  L3:                        256 MiB (16 instances)
NUMA:                        
  NUMA node(s):              1
  NUMA node0 CPU(s):         0-15
Vulnerabilities:             
  Gather data sampling:      Not affected
  Indirect target selection: Not affected
  Itlb multihit:             Not affected
  L1tf:                      Not affected
  Mds:                       Not affected
  Meltdown:                  Not affected
  Mmio stale data:           Not affected
  Reg file data sampling:    Not affected
  Retbleed:                  Mitigation; untrained return thunk; SMT disabled
  Spec rstack overflow:      Mitigation; SMT disabled
  Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl and seccomp
  Spectre v1:                Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:                Mitigation; Retpolines; IBPB conditional; STIBP disabled; RSB filling; PBRSB-eIBRS Not affected; BHI Not affected
  Srbds:                     Not affected
  Tsa:                       Not affected
  Tsx async abort:           Not affected
```

I don't know about the SSD this is run on. I could enquire, but it's just a provisioned machine from https://cloud.denbi.de/wiki/ so shouldn't be anything crazy.  That being said, it appears faster than the average sagemaker instance.  So I definitely understand you about hardware differences, but we haven't seen any relative change in performance i.e., one method is 10X faster than another on machine, but only 5X on another.  Maybe that could change, would be interested to find out!

The data is a 6 million cell subset of tahoe.  I could try generating an scdl dataset, but I don't have a script.  If you were to move that script into `main` I would be happy to run it.  

I added a `measure-collection-creation-time` option to the CLI instead of making a separate script.  Hope that's ok.  I get what seem to be reasonable numbers for dataset creation time when it is enabled.

PR #1292, opened 2025-10-28
Remove nemo dependency for state dict conversion

pstjohn:
Moves to a modifed single-file state.py solution for state dict conversion


pstjohn:
force-merging after cory's review and CI passing, blocked by ci/ changes

PR #1310, opened 2025-11-05
Fix TE test for model generation

pstjohn:
Quick fix to at least show how generate can work, although we need to make sure we support 
* batched generation
* kv caching through InferenceParams


pstjohn:
force-merging since this test doesn't run in CI anyways

PR #1311, opened 2025-11-05
Add cursor rule to run pre-commit, update devcontainer for recipes to install cursor-agent

pstjohn:


jstjohn (approved):
Looks good, can we also add a ticket for adding something like this to the framework devcontainer?

PR #1322, opened 2025-11-14
always use THD inputs for llama3 model

pstjohn:
Because left-side padding isn't supported by TE, we can always ensure that our TE inputs are THD-packed to support the standard HF generation pipeline.

### DIFF
--- a/bionemo-recipes/models/llama3/modeling_llama_te.py
+++ b/bionemo-recipes/models/llama3/modeling_llama_te.py
@@ -21,6 +21,7 @@ import torch.nn as nn
 import transformer_engine.pytorch
 import transformers
 from transformer_engine.pytorch.attention import InferenceParams
+from transformer_engine.pytorch.attention.inference import PagedKVCacheManager
 from transformer_engine.pytorch.attention.rope import RotaryPositionEmbedding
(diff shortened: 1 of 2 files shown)


bionemo-recipes/models/llama3/modeling_llama_te.py
```diff
@@ -123,71 +120,74 @@ def forward(
...
-            attention_mask = None
+            cu_seq_lens_q = kwargs["cu_seq_lens_q"]
+            cu_seq_lens_k = kwargs["cu_seq_lens_k"]
+            max_length_q = kwargs["max_length_q"]
+            max_length_k = kwargs["max_length_k"]
+
+        # If we're using kv-caching, we can't trust the max_length_q value as the true max length for rotary
+        # embeddings, since this will be 1 in generation. Instead we can take the max sequence length from the past
+        # key values object.
+        te_rope_emb = self.rotary_emb(
```
jomitchellnv, line 154: so you want the rope embeddings to be from BSHD still right?

pstjohn: the rope embeddings don't really change one way or the other; we just need to know the maximum sequence length.

PR #1318, opened 2025-11-12
Dataset class and tests for LLAMA3; support for streaming, parquet

savitha-eng:
### Description

This PR adds a genomic dataset module for training Llama3 models on nucleotide sequences, following the ESM2 native TE pattern (much of the code is similar to the ESM2 dataset). 

**Key features:**
- **Streaming Parquet datasets**: Efficient loading of large genomic datasets using HuggingFace `datasets` library with `streaming=True` to avoid loading entire datasets into memory
- **Windowing/Strided sampling**: Automatic creation of overlapping windows from long genomic sequences using the tokenizer's built-in `return_overflowing_tokens=True` parameter
- **Shuffle buffer**: Large shuffle buffer (500K samples by default) for better randomization during streaming
- **Distributed training support**: Built-in support for multi-GPU training with `split_dataset_by_node`
- **Causal LM collation**: Uses `DataCollatorForLanguageModeling` with `mlm=False` for next-token prediction

**Implementation details:**
- `create_tokenized_dataset()`: Loads Parquet data, handles dataset splits, applies tokenization with windowing
- `create_bshd_dataloader()`: Creates PyTorch DataLoader with appropriate sampler and collator
- Supports both streaming and non-streaming modes
- Supports both lazy and eager tokenization

**Testing:**
- 8 dataset tests covering windowing, streaming, lazy tokenization, and batch structure
- Mock data fixtures in `conftest.py` for CI/CD compatibility
- Note that the tests focus on testing the single node behavior/functionality of the dataloader, distributed dataset tests following the esm2 pattern are a TODO

#### Usage

```python
from dataset import create_bshd_dataloader
from distributed_config import DistributedConfig

# Configure distributed training (defaults to single GPU if env vars not set)
distributed_config = DistributedConfig()

# Create dataloader for genomic sequences
dataloader, sampler = create_bshd_dataloader(
    distributed_config=distributed_config,
    tokenizer_path="/path/to/nucleotide_fast_tokenizer",
    load_dataset_kwargs={
        "path": "parquet",
        "data_files": "/path/to/genomic_sequences.parquet",
        "split": "train",
        "streaming": True,  # Memory-efficient streaming
    },
    micro_batch_size=4,
    max_seq_length=8192,    # Window size
    stride=200,              # Overlap between windows (200 tokens)
    buffer_size=500_000,     # Shuffle buffer size
)

# Train
for batch in dataloader:
    # batch contains: input_ids, attention_mask, labels
[...]

PR #1314, opened 2025-11-07
Add nucleotide tokenizer for llama3 model to recipes

savitha-eng:
### Description

This PR adds a nucleotide tokenizer for the llama3 model, following HuggingFace's PreTrainedTokenizerFast pattern with Rust backend for performance. It follow's NeMo special token conventions (i.e., for EOS, BOS, PAD etc.). 

**Key features:**
- ASCII-based tokenizer that maps nucleotides (A, T, C, G, N, R, Y) to their ASCII values
- NeMo-style special tokens: EOS=0, PAD=1, BOS=2, UNK=3
- Pre-built tokenizer files in `nucleotide_fast_tokenizer/` directory
- `create_tokenizer.py` script for reproducibility
- Comprehensive unit tests covering tokenization, padding, and attention masks

This tokenizer will be used in genomic data loading pipelines for training llama3 models on nucleotide sequences.

#### Usage

```python
from transformers import AutoTokenizer

# Load the nucleotide tokenizer
tokenizer = AutoTokenizer.from_pretrained("bionemo-recipes/models/llama3/nucleotide_fast_tokenizer")

# Tokenize a nucleotide sequence
sequence = "ATCGATCG"
encoded = tokenizer.encode(sequence, add_special_tokens=True)
# Returns: [2, 65, 84, 67, 71, 65, 84, 67, 71, 0]
# [BOS, A, T, C, G, A, T, C, G, EOS]

# Batch tokenization with padding
batch = tokenizer(["AAAA", "TTTTTTTT"], padding=True, return_tensors="pt")
# Returns: {'input_ids': tensor([[2, 65, 65, 65, 65, 0, 1, 1, 1, 1],
#                                  [2, 84, 84, 84, 84, 84, 84, 84, 84, 0]]),
#           'attention_mask': tensor([[1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
#                                      [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]])}
```

### Type of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactor
- [ ] Documentation update
- [ ] Other (please describe):

### CI Pipeline Configuration

Configure CI behavior by applying the relevant labels. By default, only basic unit tests are run.

- [ciflow:skip](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:skip) - Skip all CI tests for this PR
[...]

PR #1325, opened 2025-11-17
Pstjohn/savitha/llama3 dataset pr

pstjohn:
merges #1318 and #1314 to main to start the llama3 recipe, fixes a few pre-commit lints


bionemo-recipes/models/llama3/tests/test_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def test_tokenizer_mixed_nucleotides(tokenizer):
+    """Test all standard nucleotides encode correctly."""
+    sequence = "ATCGGTC"
+    encoded = tokenizer.encode(sequence, add_special_tokens=False)
+
+    # A=65, T=84, C=67, G=71
+    # ATCGGTC = A, T, C, G, G, T, C
+    expected = [65, 84, 67, 71, 71, 84, 67]
```
trvachov, line 138: Should this file just have a simple struct that defines this ATCG  = 65, 84, 67, 71 mapping? So if it changes, you just have to change it in one place?

pstjohn: it might make the test more readable, but those are the unicode code point for 'A', etc, which are unlikely to change unless we change how we do the vocab.

trvachov: ok fine.

trvachov (approved):

bionemo-recipes/models/llama3/tests/test_tokenizer.py
pstjohn: @savitha-eng could also just do
```python
sequence = "ATCGGTC"
encoded = tokenizer.encode(sequence, add_special_tokens=False)
expected = [ord(c) for c in sequence]
```

PR #1334, opened 2025-11-19
starting work for swqa

pstjohn:
We should be more explicit about our non-PyTorch install path, and ensure all our tests pass.
We should also make sure we have a reasonable training command for SWQA to run on assorted hardware SKUs to report performance and functionality.

### DIFF
--- a/.devcontainer/recipes/devcontainer.json
+++ b/.devcontainer/recipes/devcontainer.json
@@ -33,7 +33,8 @@
                 "ms-toolsai.jupyter",
                 "eamodio.gitlens",
                 "tamasfe.even-better-toml",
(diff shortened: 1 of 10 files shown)


trvachov (approved):

pstjohn:
confirmed working locally with

```bash
docker build --progress=plain -t esm2_cuda_test -f Dockerfile.cuda .
 docker run -it --gpus all --network host --ipc=host --rm -v ${PWD}:/workspace/bionemo esm2_cuda_test /bin/bash -c "pytest -v ."
```

but building flash-attention from source is not fun:
```
#8 DONE 7196.1s                                                                                                                
```

PR #1341, opened 2025-11-21
Add checkpoint export script to llama3 model, ensure recipes/llama3 is using TE layers

pstjohn:


bionemo-recipes/models/llama3/modeling_llama_te.py
```diff
@@ -148,6 +161,12 @@ def forward(
             max_length_q = kwargs["max_length_q"]
             max_length_k = kwargs["max_length_k"]
 
+        else:
```
savitha-eng, line 164: is this mainly for backwards compatibility? so generation won't work here right?

pstjohn:
force-merging to skip the framework CI

bionemo-recipes/models/llama3/modeling_llama_te.py
pstjohn: not with BSHD inputs, but you can always set the model to use THD attention and generate

PR #1343, opened 2025-11-24
add config overrides for llama recipe

pstjohn:
Updates the hydra config to let us pass model size information via the hydra configs


jomitchellnv (approved):
LGTM

PR #1345, opened 2025-11-25
Adding sequence packing / THD dataloading to llama3

pstjohn:
Adds sequence packing to llama3 recipe


bionemo-recipes/recipes/llama3_native_te/dataset.py
```diff
@@ -120,7 +123,7 @@ def create_bshd_dataloader(
     tokenizer_path: str,
     load_dataset_kwargs: dict,
     micro_batch_size: int,
-    num_workers: int = 0,
+    num_workers: int = 1,
```
jomitchellnv, line 126: Was the `0` set previously for debugging?

bionemo-recipes/recipes/llama3_native_te/genomic_dataset.py
```diff
@@ -13,25 +13,48 @@
...
-"""Data collator for genomic sequence training with custom masking.
+"""Genomic sequence masking functions for data preprocessing.
 
-This module provides a collator that wraps any base collator and adds genomic masking:
-- Uppercase labels (while keeping inputs mixed case)
-- Mask degenerate bases (non-ACGT characters)
-- Mask control characters (@, #)
+Core functions for genomic data preprocessing during training:
+- make_upper_case: Convert lowercase tokens to uppercase
+- Evo2MaskingConstants: Standard DNA tokens and control characters
```
jomitchellnv, line 20: Evo2?

bionemo-recipes/recipes/llama3_native_te/hydra_config/defaults.yaml
```diff
@@ -22,6 +22,8 @@ use_torch_compile: false
 # Whether to use gradient checkpointing to trade compute for memory
 use_gradient_checkpointing: false
 
+use_sequence_packing: false
```
jomitchellnv, line 25: why not set to true?

bionemo-recipes/recipes/llama3_native_te/dataset.py
pstjohn: I think so, yeah. But we can always set it back to debug; good to make sure we're not blocking on dataloading by default though

bionemo-recipes/recipes/llama3_native_te/genomic_dataset.py
pstjohn: huh, i didn't make those changes 😆 . I'll revert, wonder if it was cursor when i was trying to tab-fix something

bionemo-recipes/recipes/llama3_native_te/hydra_config/defaults.yaml
pstjohn: 🤷  would change the current test suite that opts-in to sequence packing. It's a new feature, I think it's fine to leave off by default for now

bionemo-recipes/recipes/llama3_native_te/genomic_dataset.py
pstjohn: oh, i see -- this came from the original `genomic_masking_functions.py` file. i combined these together

PR #1368, opened 2025-12-09
Ensure parameters are initialized correctly on the meta device

pstjohn:
It looks like `reset_parameters` isn't enough to ensure that newly-initialized models using the torch.device("meta") semantics are identical to those created without it. This PR adds tests to ensure that these parameter mean and stdevs are identical, and uses 
```python
    if args.use_meta_device:
        model.to_empty(device=device)
        model.apply(model._init_weights)
```
to ensure the weights are correctly initialized following `to_empty()`.

### DIFF
--- a/bionemo-recipes/models/llama3/tests/test_meta_device_init.py
+++ b/bionemo-recipes/models/llama3/tests/test_meta_device_init.py
@@ -13,6 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+Test that parameter distributions are identical with and without meta device initialization.
+
+These tests verify that when using meta device initialization (creating the model on meta device, then calling
+`to_empty` and `_init_weights`), the resulting parameter distributions (mean and std) match those from normal
+initialization. This is important because we previously observed differences in convergence between meta-device-init and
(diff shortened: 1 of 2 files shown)


bionemo-recipes/models/llama3/tests/test_meta_device_init.py
```diff
@@ -0,0 +1,189 @@
...
+from modeling_llama_te import NVLlamaConfig, NVLlamaForCausalLM
+
+
+requires_multi_gpu = pytest.mark.skipif(
+    not torch.cuda.is_available() or torch.cuda.device_count() < 2,
+    reason="Test requires at least 2 GPUs",
+)
+
+
+def test_meta_device_init():
```
jomitchellnv, line 49: can you add a bit of info about how this test works? What is model normal vs model meta and why do we expect them to be the same

pstjohn: ok, added. Yeah, these tests pass now, but didn't before. I suspect that's the reason we were seeing differences in the convergence between using meta-device init and not using it. Especially the rope weights, which aren't trained, weren't being set.

PR #1371, opened 2025-12-10
adding lm-eval test harness

pstjohn:
Adds explicit testing for lm-eval, although these don't trigger in CI (currently). We should find a faster way to test this


savitha-eng (approved):
lgtm

PR #1374, opened 2025-12-12
Add llama3 recipe readme

pstjohn:
Adds an initial README for llama3, with the initial lingua-1b benchmark results.


bionemo-recipes/recipes/llama3_native_te/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+### How to deploy this recipe on cloud providers
+
+🚧 Under development
+
+## Supported Models and Training Features
+
+| Model                                    | BF16 | FP8<sup>[1]</sup> | THD Input Format | FP8 with THD Input Format | MXFP8<sup>[2]</sup> | Context Parallelism |
+| ---------------------------------------- | ---- | ----------------- | ---------------- | ------------------------- | ------------------- | ------------------- |
+| [Llama 3](../../models/llama3/README.md) | ✅   | ✅                | ✅               | ✅                        | ✅                  | 🚧                  |
```
trvachov: Can this be added to the other READMEs tables as well? Toplevel: https://github.com/NVIDIA/bionemo-framework and https://github.com/NVIDIA/bionemo-framework/tree/main/bionemo-recipes ?

bionemo-recipes/recipes/llama3_native_te/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+```bash
+docker run -it --gpus all --network host --ipc=host --rm -v ${PWD}:/workspace/bionemo llama3_native_te /bin/bash
+```
+
+Alternatively, the dependencies can be installed manually in an environment with CUDA support. See `requirements.txt` for the list of dependencies.
+
+### Performance Benchmarks
+
+![Llama 3 Lingua 1B Validation](../../../docs/docs/assets/images/recipes/llama3_lingua_1b_validation.png)
```
trvachov: I think we need to make a note about the spikes...do we know why? they seem to recover always?

trvachov:
Shouldn't we make the comparison against the original Lingua, not NeMo?

bionemo-recipes/recipes/llama3_native_te/README.md
pstjohn: this run was relatively smooth, idk, it's not monotonically decreasing but we've seen much worse. Could be the data; the DDP and FSDP2 runs had a similar noise profile. But we could also try John's spike-no-more initialization and see if that helps

trvachov: My bad, I think I must have commented on the wrong picture -- I am talking about the step-timing spikes.

pstjohn: ah gotcha. those are saving checkpoints 😢  well, some of them are... probably also just when `gc` happens to run

bionemo-recipes/recipes/llama3_native_te/README.md
pstjohn: yup added

trvachov:
Approving, because this is close enough to done, but there's two more things that need changes.
* Adding llama3 to the top-level readme (look under the drop down: bionemo-recipes support matrix
* your description of the step-timing spikes is incomplete: "The spike in training step time every 10,000 steps are due" ... I am just looking for something high level to let the user know "this is normal", e.g. "step timing spikes are due to intermittent checkpoint saving and/or garbage collection".

bionemo-recipes/recipes/llama3_native_te/README.md
trvachov: One more missing, toplevel README (root/README)

bionemo-recipes/recipes/llama3_native_te/README.md
trvachov: I think the step-timing spikes description sentence isn't quite complete...looks like it got cut off mid sentence. "The spike in training step time every 10,000 steps are due". ..

trvachov (approved):

PR #1375, opened 2025-12-12
fix fsdp2 checkpoint load

pstjohn:
We need to pass the dataloader even if we're not loading it from a checkpoint, since it gets passed through as none otherwise.

### DIFF
--- a/bionemo-recipes/recipes/llama3_native_te/train_fsdp2.py
+++ b/bionemo-recipes/recipes/llama3_native_te/train_fsdp2.py
@@ -116,7 +116,7 @@ def main(args: DictConfig) -> float | None:
             scheduler=scheduler,
             ckpt_path=ckpt_path,
             dist_config=dist_config,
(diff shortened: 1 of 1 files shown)


bionemo-recipes/recipes/llama3_native_te/train_fsdp2.py
```diff
@@ -116,7 +116,7 @@ def main(args: DictConfig) -> float | None:
             scheduler=scheduler,
             ckpt_path=ckpt_path,
             dist_config=dist_config,
-            dataloader=train_dataloader if args.dataset.use_stateful_dataloader else None,
+            dataloader=train_dataloader,
```
savitha-eng, line 119: i think esm2's train_fsdp2 also has this code - not sure if it needs to be fixed there too.

trvachov: great catch if true

PR #1382, opened 2025-12-16
Refactor MLMDataCollatorWithFlattening

pstjohn:
Updates this class to be more general for llama3 usage by making it accept a base collator and perform flattening after-the-fact. Essentially we're now always doing the bshd-compatable forward call.

This is going to make it easier to implement CP in the llama3 recipe


jomitchellnv (approved):
LGTM but let's build a container and test out the CP tests to make sure this passes before submission

PR #1397, opened 2025-12-24
Make sure we pass device="meta" to TE layers during meta device init

pstjohn:
TransformerEngine requires that we pass `device="meta"` (and importantly _not_ `device=torch.device("meta")`) to layer constructors to initialize parameters on the meta device. 

This makes sure we pass the right device to the layer constructor and adds tests to ensure the parameters are actually being placed on the right devices.

For TransformerEngine layers, we want to be moving parameters from the meta device to cuda with
```python
for module in model.modules():
    if hasattr(module, "reset_parameters"):
        module.reset_parameters()
```

while for HF layers (e.g., nn.Embedding), we want to be doing
```python
model.to_empty("cuda")
model.apply(model._init_weights)
```
to ensure that we pick up the config.initializer_range initialization correctly.

The issue is that we can't do `to_empty("cuda")` or `_init_weights` on TE layers, nor can we do `reset_parameters()` on the HF layers without the preceeding `to_empty`, and this doesn't use the HF config when creating initial values.

Closes BIO-1

### DIFF
--- a/bionemo-recipes/recipes/esm2_native_te/train_fsdp2.py
+++ b/bionemo-recipes/recipes/esm2_native_te/train_fsdp2.py
@@ -93,7 +93,12 @@ def main(args: DictConfig) -> float | None:
     # If we're using meta device, we need to move sharded weights to the cuda device and initialize the parameters.
     # Note, this should happen before we create the optimizer.
     if args.use_meta_device:
-        model.init_empty_weights()
+        if hasattr(model, "init_empty_weights"):
+            # TE layers require special handling to initialize the weights from the meta device.
+            model.init_empty_weights()
+        else:
+            model.to_empty(device=device)
+            model.apply(model._init_weights)
 
     # Create optimizer. Convert OmegaConf to regular dict to avoid serialization issues (BIONEMO-2873).
     optimizer = AdamW(model.parameters(), **OmegaConf.to_container(args.adamw_kwargs, resolve=True))  # type: ignore


bionemo-recipes/models/llama3/tests/test_meta_device_init.py
```diff
@@ -61,9 +61,17 @@ def test_meta_device_init():
...
         model_meta_init = NVLlamaForCausalLM(config)
 
+    # Assert parameters are actually on the meta device
+    for name, parameter in model_meta_init.named_parameters():
+        assert parameter.device == torch.device("meta"), f"Parameter {name} is not on the meta device"
+
     model_meta_init.to_empty(device="cuda")
     model_meta_init.apply(model_meta_init._init_weights)
 
+    # Assert parameters are actually on the cuda device after to_empty
```
trvachov: So IIUC these tests now make sure we do not make the same mistake with llama3 and esm2 models, but to avoid making this mistake on other models, we'd need to add those tests as well, right? E.g. for geneformer or whatnot. 

What I'm trying to understand is how can we prevent a user who is copy-pasting/using the recipes as a template from making this mistake?

pstjohn: Right, so i think the solution here is some sort of importable testing framework that would run through a bunch of these standard tests; and do things like
* check parameter initialization
* check checkpointing
* FP8 support and accuracy
* check golden values against some reference model
* check downstream library compatibility (PEFT / lm-eval / etc)
e.g., all the things we currently do in our `models/**/tests` directory. But this probably isn't the PR to figure out the best way of setting that up.

cspades (approved):

PR #1401, opened 2026-01-05
BIO-3: fix multi-gpu mfsdp tests

pstjohn:
turn off clip grad norm until we can fix BIO-3


cspades (approved):
Thanks, I think in general now it means we need MFSDP to support gradient modification utilities that support uneven numbers of sharded gradients on each rank.

@shjwudp for viz, more work to maintain sharded gradient support in the PyTorch ecosystem, similar to clip_grad_norm for FSDP1.

trvachov (approved):

PR #1407, opened 2026-01-09
Add CP support to modeling_llama_te.py

pstjohn:
* Makes the attention_mask type more easily controllable through a config kwarg
* passes the additional `_padded` kwargs through to the TE layers
* simplifies some of the pre-processing logic and adds comments as to why they're needed

Fixes BIO-12

### DIFF
--- a/bionemo-recipes/recipes/llama3_native_te/modeling_llama_te.py
+++ b/bionemo-recipes/recipes/llama3_native_te/modeling_llama_te.py
@@ -43,6 +43,7 @@ class NVLlamaConfig(LlamaConfig):
     """NVLlama configuration."""
 
     attn_input_format: str = "thd"
+    self_attn_mask_type: str = "padding_causal"
 
 
 class NVLlamaPreTrainedModel(PreTrainedModel):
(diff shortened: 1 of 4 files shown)


bionemo-recipes/recipes/llama3_native_te/modeling_llama_te.py
```diff
@@ -233,6 +217,8 @@ def forward(
             )
             past_key_values.pre_step(OrderedDict(zip(list(range(len(lengths))), lengths)))
 
+        te_rope_emb = self.rotary_emb(max_seq_len=self.config.max_position_embeddings)
```
jomitchellnv, line 220: isn't this length based on max(cu_seq_lens[i] - cu_seq_lens[i-1])

jomitchellnv (approved):
one question

savitha-eng (approved):

bionemo-recipes/recipes/llama3_native_te/modeling_llama_te.py
pstjohn: when you're in generation mode, the length coming in is only 1, but you still need to have enough rope embeddings. I think it's easier just to use this config parameter, and increase it if necessary

PR #1409, opened 2026-01-09
set weights_only=True to resolve security concern

pstjohn:
Closes BIO-15


cspades (approved):
weights_only=True is the default so we can also just delete it

PR #1434, opened 2026-01-24
add fp8 to config

jwilber:
Add fp8 to esm2 15b config.
Also:
- reduce non-15b steps from 20k to 10k.
- remove an esm2 650m job
- remove esm2 3b from automated tracking

### DIFF
--- a/.github/workflows/convergence-tests.yml
+++ b/.github/workflows/convergence-tests.yml
@@ -39,7 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


ci/lepton/model_convergence/configs/recipes/esm2_native_te_3b.yaml
```diff
@@ -58,15 +58,15 @@ wandb_init_args:
 # these should match the keys in the recipe's config file
 model_tag: nvidia/esm2_t36_3B_UR50D
 # task_cmd: train_fsdp2 # mfsdp
-num_train_steps: 20_000
+num_train_steps: 10_000
```
jomitchellnv, line 61: I thought we were going to do 20k steps?

jwilber: that's the 3B model, not 15B

PR #1437, opened 2026-01-27
add inference.nvidia.com setup

pstjohn:
Adds inference.nvidia.com setup for the recipe devcontainer


.devcontainer/recipes/setup_claude_env.sh
```diff
@@ -0,0 +1,589 @@
...
+
+# ============================================================================
+# NVIDIA API Key Setup
+# ============================================================================
+echo ""
+echo "🔑 NVIDIA API Key"
+echo "   ─────────────────────────────────────────────────────────────────"
+
+# Define the API key file
+API_KEY_FILE="$HOME/.nvidia-api-key"
```
jomitchellnv, line 245: We ignore dotfiles right? So this won't make its way into a container

PR #1440, opened 2026-01-28
pin transformers==4.57.6

jwilber:
pin transformers==4.57.6


jwilber:
Petter resolved in #1439

PR #1441, opened 2026-02-02
pin transformers, increase tolerance in gradient check

pstjohn:
Fix a few issues in recipes ci


pstjohn:
force-merging to unblock ci

PR #1445, opened 2026-02-04
Extract modeling tests into a common extendable test suite

pstjohn:
We currently have a lot of duplication in model tests, which makes adding new models harder than it should be (and means that we may be missing some test coverage between models). We can copy `transformers` and have a TestClass base class that we extend for each model that runs each through a common battery of tests.

## Summary by CodeRabbit

* **New Features**
  * Added Claude development environment setup script with NVIDIA API key management and local proxy integration
  * Introduced unified testing framework for BioNeMo models with reusable test infrastructure
  * Added LiteLLM proxy configuration for model routing with context-window fallback strategies
  * Introduced separator token parameter for sequence packing in data collators

* **Improvements**
  * Enhanced FP8 quantization handling for model inference
  * Consolidated test fixtures and utilities across model implementations
  * Improved test coverage for model initialization and conversion workflows

### DIFF
--- a/bionemo-recipes/models/llama3/tests/common/test_modeling_common.py
+++ b/bionemo-recipes/models/llama3/tests/common/test_modeling_common.py
@@ -25,6 +25,7 @@ import torch
 import transformer_engine.pytorch
 from torch import nn
 from transformer_engine.pytorch import QuantizedTensor
+from transformer_engine.pytorch.quantization import FP8GlobalStateManager
 from transformers import AutoConfig, PretrainedConfig, PreTrainedModel, PreTrainedTokenizer, set_seed
 
 
@@ -411,6 +412,11 @@ class BaseModelTest(ABC):
     @pytest.fixture(autouse=True, scope="function")
     def set_seed(self):
         set_seed(42)
+
+    @pytest.fixture(autouse=True, scope="function")
+    def reset_fp8_context(self):
+        """Make sure we clean up the FP8 context after each test."""
+        FP8GlobalStateManager.reset()
 
     # ==================== Forward and Backward Smoke Tests ====================
 
--- a/bionemo-recipes/models/esm2/tests/common/test_modeling_common.py
+++ b/bionemo-recipes/models/esm2/tests/common/test_modeling_common.py
@@ -25,6 +25,7 @@ import torch
 import transformer_engine.pytorch
 from torch import nn
 from transformer_engine.pytorch import QuantizedTensor
+from transformer_engine.pytorch.quantization import FP8GlobalStateManager
 from transformers import AutoConfig, PretrainedConfig, PreTrainedModel, PreTrainedTokenizer, set_seed
 
 
@@ -411,6 +412,11 @@ class BaseModelTest(ABC):
     @pytest.fixture(autouse=True, scope="function")
     def set_seed(self):
         set_seed(42)
+
+    @pytest.fixture(autouse=True, scope="function")
+    def reset_fp8_context(self):
+        """Make sure we clean up the FP8 context after each test."""
+        FP8GlobalStateManager.reset()
 
     # ==================== Forward and Backward Smoke Tests ====================
 


bionemo-recipes/models/esm2/src/esm/collator.py
```diff
@@ -213,6 +218,7 @@ def _pad_sequences_to_be_divisible_by(self, batch):
         batch["labels"] = labels_padded.unsqueeze(0)
         batch["cu_seq_lens_q_padded"] = cu_seqlens_padded.to(torch.int32)
         batch["cu_seq_lens_k_padded"] = cu_seqlens_padded.to(torch.int32)
+        batch["pad_between_seqs"] = True
```
jomitchellnv, line 221: Looks like this way just moved from 321 to here yea? I wonder why we had it down there

bionemo-recipes/models/esm2/src/esm/modeling_esm_te.py
```diff
@@ -597,14 +599,18 @@ def forward(
...
+                    src_lengths_padded = src_lengths
                 # We need to find the number of masked tokens in each sequence in the padded batch.
                 is_masked = (input_ids == self.mask_token_id).squeeze(0)
                 n_masked_per_seq = torch.nested.nested_tensor_from_jagged(
                     is_masked, offsets=kwargs["cu_seq_lens_q"]
                 ).sum(1)
                 mask_ratio_observed = n_masked_per_seq.float() / src_lengths
                 scale_factor = (1 - mask_ratio_train) / (1 - mask_ratio_observed)
-                reshaped_scale_factor = torch.repeat_interleave(scale_factor, src_lengths, dim=0)
+                reshaped_scale_factor = torch.repeat_interleave(scale_factor, src_lengths_padded, dim=0)
```
jomitchellnv, line 613: whats this aboot [british accent]

bionemo-recipes/models/esm2/tests/test_thd.py
```diff
@@ -326,4 +326,5 @@ def test_thd_vs_padded_thd_equivalence(
     # Now index select the padded logits to get the real logits.
     logits_unpadded = outputs_thd_padded.logits.index_select(0, real_idx.cuda())
 
+    breakpoint()
```
jomitchellnv: deleteme

bionemo-recipes/models/esm2/src/esm/collator.py
pstjohn: yeah this was a bug 😄 . we only set this for CP, not padded THD (which we use in FP8), and the outputs are just silently wrong

bionemo-recipes/models/llama3/tests/common/test_modeling_common.py
```diff
@@ -0,0 +1,924 @@
...
+            For ESM2: model.esm.encoder.layers
+            For LLaMA3: model.model.layers
+        """
+        pass
+
+    @abstractmethod
+    def get_test_input_data(
+        self,
+        format: Literal["bshd", "thd"] = "bshd",
+        pad_to_multiple_of: int | None = None,
```
jomitchellnv, line 170: Should the padding functions here be extended to include `pad_between_seqs` etc? OR should there be an optional `Callable` function we can call to do any kinda custom padding?

bionemo-recipes/models/llama3/tests/common/test_modeling_common.py
```diff
@@ -0,0 +1,924 @@
...
+
+        Returns:
+            List of format strings (e.g., ["bshd", "thd"]).
+        """
+        return ["bshd"]
+
+    def verify_model_parameters_initialized_correctly(
+        self,
+        model: PreTrainedModel,
+        atol: Optional[float] = None,
```
jomitchellnv: Didn't you like using `atol : float | None` instead of `Optional`

bionemo-recipes/models/esm2/src/esm/modeling_esm_te.py
pstjohn: yeah apparently this didn't work for ESM-2 with padded THD. not sure how we were running those benchmarks unless we turned this token_dropoff stuff off. Anyways, it makes sure we correctly broadcast these scale factors to the padded embedding matrix.

bionemo-recipes/models/esm2/tests/test_thd.py
pstjohn: we should delete the whole file actually, we have this covered in the baseline test now

bionemo-recipes/models/llama3/tests/common/test_modeling_common.py
pstjohn: i just use that single argument to do both.  if it's THD and `pad_to_multiple_of`; it's pad_between_seqs

bionemo-recipes/models/llama3/tests/common/test_modeling_common.py
pstjohn: this was claude lol, i'll fix

PR #1449, opened 2026-02-05
Fix CP Collator with causal LM

pstjohn:
### Fix for DataCollatorForContextParallel
when we pass labels to the HF loss function in causal LM, it shifts the labels for us; e.g.
`input_ids = labels = [0, 1, 2, 3, 4, 5, 6, 7]`
becomes
`shift_labels = [1, 2, 3, 4, 5, 6, 7, -100]`
so that the loss function correctly applies the casual cross-entropy.

But when we split tokens for CP, this gets jumbed. E.g., if we have CP=2, the inputs on rank0 will be `[0, 1, 6, 7]` and on rank 1 will be `[2, 3, 4, 5]`. if we let HF do the shift for us, we'll get
`[1, 6, 7, -100]`, [3, 4, 5, -100]`, when what we actually want is
`[1, 2, 7, -100]` and `[3, 4, 5, 6]`.

The solution is to shift the tokens _before_ splitting and passing the shifted labels through the CP split algorithm and to the model.

Luckily HF supports this through the `shift_labels` input, https://github.com/huggingface/transformers/pull/36607.

### Fix for DataCollatorWithFlattening
for causal modeling with THD inputs, we need to make sure that labels are separated with a -100 masking token between inputs, since the huggingface loss function doesn't take in `cu_seq_lens` to know when it hits a sequence boundary.

E.g. for inputs `[0, 1, 2, A, B, C]`, where 012, ABC are different sequences, we want the corresponding labels (before shifting) to be `[0, 1, 2, -100, B, C]`, so that the input `2` doesn't try to predict `A` from the next sequence.

## Summary by CodeRabbit

* **New Features**
  * Added causal language model support with label shifting capability across model implementations
  * Introduced separator labels for optimized sequence packing in flattening operations

* **Tests**
  * Expanded test coverage for causal LM batch preparation and label handling
  * Added context-parallel data collation tests for causal LM scenarios

* **Chores**
  * Updated development container configuration

### DIFF
--- a/bionemo-recipes/models/esm2/src/esm/collator.py
+++ b/bionemo-recipes/models/esm2/src/esm/collator.py
@@ -56,6 +56,7 @@ class DataCollatorWithFlattening:
         return_position_ids (bool): Whether to return position ids (default False).
         pad_to_multiple_of (int, optional): If set, pads the total sequence length to be divisible by this number.
         pad_sequences_to_be_divisible_by (int, optional): If set, each individual sequence is padded to this value.
+    separator_label (int, optional): A label to insert between sequences, typically should be -100 for causal LM.
 
     Example:
         >>> from transformers import AutoTokenizer, DataCollatorForLanguageModeling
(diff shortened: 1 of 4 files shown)


pstjohn:
Squashed into #1445

PR #1454, opened 2026-02-06
Bug fixes for evo2 inference.  CP at least doesn't error.  TP working…

kjaniknvidia:
… on 1 node.

### Description

**EVO2 Dockerfile Change** - Added ENVs to force casual_conv1d to be built

**CP Partial Fixes**
Running Evo2 Megatron inference with --context-parallel-size 2 caused tensor shape mismatches because the inference path doesn't split sequences across CP ranks (each rank gets the full sequence), but the code was still applying CP-related AllToAll operations and CP-sharded parameter slicing.

Tested that it no longer errors and returns the generation, but enabling full useful CP is a much heavier lift.

**TP FIxes**
With TP > 1, sequence_parallel was automatically set to True. Megatron's inference engine doesn't support sequence parallelism for non-MoE models, raising NotImplementedError. Sequence parallelism is a training-only activation memory optimization and isn't needed for inference.

In ParallelHyenaOperator.forward_long(), the call to engine.parallel_iir() passed self.hidden_size (the full model hidden size, e.g., 4096). With TP=8, each rank only has width_per_tp_group = 4096 / 8 = 512 channels, so the concatenated tensor u has 3 * 512 = 1536 channels. The parallel_iir function tried to split it into three chunks of 4096, causing a RuntimeError.

**Infer.py Test Enabling**
I didn't know this, but apparently unix command lines have a character limit.  Even though I was generating the prompt it evaluated to something longer than 128K characters.  Added a --prompt-file switch that reads the prompt... from a file.

#### Usage

Tested TP=8 on one node.
16K prompt fits on one GPU. 32K OOMs

Confirmed that 128K prompt fits on 8 GPUs and 256K prompt OOMs.

Generate the prompt:
python3 -c "
import random
random.seed(42)
seq = ''.join(random.choices('ACGT', k=131072))
prompt = '|d_Bacteria;p_Pseudomonadota;c_Gammaproteobacteria;o_Enterobacterales|' + seq
with open('/tmp/prompt128k.txt', 'w') as f:
   f.write(prompt)
print(f'Wrote prompt of length {len(prompt)}')
"

Execute:
torchrun --nproc-per-node 8 ./src/bionemo/evo2/run/infer.py \
 --tensor-parallel-size 8 \
 --ckpt-dir $CKPT_OUT_DIR \
 --max-new-tokens 100 \
 --max-seq-length 131300 \
 --prompt-file /tmp/prompt128k.txt

### Type of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor
- [ ] Documentation update

### CI Pipeline Configuration

It's my first real change.  Run everything?

> [!NOTE]

#### Authorizing CI Runs

#### Triggering Code Rabbit AI Review

To trigger a code review from code rabbit, comment on a pull request with one of these commands:

- @coderabbitai review - Triggers a standard review
- @coderabbitai full review - Triggers a comprehensive review

### Pre-submit Checklist

### DIFF
--- a/bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/hyena_utils.py
+++ b/bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/hyena_utils.py
@@ -1152,9 +1152,7 @@ class ParallelHyenaOperator(nn.Module):
                         x1=x1, x2=x2, v=v, h=h, bias=conv_bias, inference_context=inference_context
                     )
                 elif self.operator_type == "hyena":
-                    z = self.forward_long(
-                        x1=x1, x2=x2, v=v, h=h, bias=conv_bias, inference_context=inference_context
-                    )
+                    z = self.forward_long(x1=x1, x2=x2, v=v, h=h, bias=conv_bias, inference_context=inference_context)
                 else:
                     raise ValueError(f"Unsupported operator_type for inference: {self.operator_type}")
 
--- a/bionemo-recipes/recipes/evo2_megatron/tests/bionemo/evo2/run/test_predict.py
+++ b/bionemo-recipes/recipes/evo2_megatron/tests/bionemo/evo2/run/test_predict.py
@@ -297,9 +297,7 @@ def baseline_predictions_7b_1m_results(
         ),  # Cover case where FP8 was not supported with TP=2
         pytest.param(1, 1, 1, 2, False, "epoch", id="ddp=1,cp=1,pp=1,tp=2,fp8=False,wi=epoch"),
         pytest.param(1, 1, 1, 8, False, "epoch", id="ddp=1,cp=1,pp=1,tp=8,fp8=False,wi=epoch"),
-        pytest.param(
-            1, 1, 1, 8, True, "epoch", id="ddp=1,cp=1,pp=1,tp=8,fp8=True,wi=epoch"
-        ),  # Cover TP=8 with FP8
+        pytest.param(1, 1, 1, 8, True, "epoch", id="ddp=1,cp=1,pp=1,tp=8,fp8=True,wi=epoch"),  # Cover TP=8 with FP8
     ],
 )
 @pytest.mark.slow
@@ -418,7 +416,7 @@ def test_predict_evo2_equivalent_with_log_probs(
             #  This should be investigated. TP=2 on some GPUs needs even more tolerance.
             rel = 2e-3
         elif fp8:
-            # FP8 + TP can have ~1–2% log-prob drift vs baseline; use 2% relative tolerance.
+            # FP8 + TP can have 1 to 2% log-prob drift vs baseline; use 2% relative tolerance.
             rel = 2e-2
         else:
             rel = 1e-6


trvachov:
@balvisio could you help review?

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/hyena_mixer.py
```diff
@@ -288,15 +288,16 @@ def forward(self, x, layer_past=None, inference_context=None, _hyena_use_cp=True
...
+        # The AllToAll operations in Hyena operators assume sequence-split input which
+        # only happens during training.
+        if inference_context is not None:
+            _proj_use_cp = False
+        elif _hyena_use_cp:
             cp_group = self.pg_collection.cp
             cp_size = cp_group.size()
-        else:
-            cp_group = None
-            cp_size = 1
```
jstjohn, line 297: Should we set this to 1/None in the case when we do not use CP? Forgot why we had these defaults here but there may be weird cases that fail where we divide by cp_size somewhere...

kjaniknvidia:
I ran py.test tests/bionemo/evo2/

Were these all passing previously?

=========================== short test summary info ============================
FAILED tests/bionemo/evo2/run/test_predict.py::test_predict_evo2_equivalent_with_log_probs[ddp=1,cp=1,pp=1,tp=2,fp8=True,wi=epoch]
FAILED tests/bionemo/evo2/test_stop_and_go.py::test_stop_and_go[1-1-1-False-bf16_with_nvfp4_mixed]
FAILED tests/bionemo/evo2/test_stop_and_go.py::test_stop_and_go[1-1-1-False-bf16_with_mxfp8_mixed]
==== 3 failed, 232 passed, 9 skipped, 161107 warnings in 2885.77s (0:48:05) ====

First one missed, just outside the range.

E             comparison failed
E             Obtained: -76.17060852050781
E             Expected: -77.04071044921875 ± 0.770407

The other two look like they are related to TE not supporting (NVFP4/MXFP8) precisions for hyena.....so "shouldn't" have anything to do with the inference changes?

jstjohn:
@kjaniknvidia interesting, yeah we should mark `bf16_with_mxfp8_mixed` and `bf16_with_nvfp4_mixed` as skip. I don't think those were passing before. We may want to get them passing at some point though, so maybe mark them xfail or something.

tests/bionemo/evo2/run/test_predict.py::test_predict_evo2_equivalent_with_log_probs[ddp=1,cp=1,pp=1,tp=2,fp8=True,wi=epoch]
This one is interesting. It's likely a hardware difference. Were you running this on an H100 or something else?

The thing I'm more interested in is adding test coverage for the code you added. Specifically:

Add a test to https://github.com/NVIDIA/bionemo-framework/blob/main/bionemo-recipes/recipes/evo2_megatron/tests/bionemo/evo2/run/test_infer.py that uses the same gold-standard values and test method of https://github.com/NVIDIA/bionemo-framework/blob/main/bionemo-recipes/recipes/evo2_megatron/tests/bionemo/evo2/test_evo2.py#L735 but with torchrun CLI calls and use of TP=[1,2], PP=[1,2] and look into whether megatron inference supports CP like is being added to vLLM https://docs.vllm.ai/en/latest/serving/context_parallel_deployment/ if CP is supported also add CP=[1,2] coverage. Add skips for CI but verify that this works manually.

The test works by taking four well known sequences and first splitting them in half. These are the same four sequences that Arc used for internal testing of their model. Then 500 tokens (verify this number) are generated for each sequence and compared to the next 500 known tokens for each sequence. We have expected sequence identity for each of the 4 sequences given the deterministic topk=1 sampling strategy for generation.

kjaniknvidia:
> @kjaniknvidia interesting, yeah we should mark `bf16_with_mxfp8_mixed` and `bf16_with_nvfp4_mixed` as skip. I don't think those were passing before. We may want to get them passing at some point though, so maybe mark them xfail or something.

@jstjohn 
I added the test but I'm looking at the other two tests.  They have a comment saying to XFAIL.
I don't know if that's doing anything.  The XFAIL examples at https://docs.pytest.org/en/stable/how-to/skipping.html all put the @pytest.mark.xfail before the definition so I'm not sure how to do it when it's been parameterized like this....

        pytest.param(
            1,
            1,
            1,
            False,
            "bf16_with_nvfp4_mixed",
            marks=pytest.mark.skipif(bool(os.environ.get("CI")), reason="Skip in CI"),
        ),  # XFAIL other than blackwell+
        pytest.param(
            1,
            1,
            1,
            False,
            "bf16_with_mxfp8_mixed",
            marks=pytest.mark.skipif(bool(os.environ.get("CI")), reason="Skip in CI"),
        ),  # XFAIL other than blackwell+
        
        
 Okay I probably should have asked Claude in the first place.  Here's what it did:
        
        pytest.param(
            1,
            1,
            1,
            False,
            "bf16_with_nvfp4_mixed",
            marks=(
                pytest.mark.skipif(bool(os.environ.get("CI")), reason="Skip in CI"),
                pytest.mark.xfail(reason="NVFP4: TE checkpoint/backward issues; known on non-Blackwell", strict=False),
            ),
        ),
        pytest.param(
            1,
            1,
            1,
            False,
            "bf16_with_mxfp8_mixed",
            marks=(
                pytest.mark.skipif(bool(os.environ.get("CI")), reason="Skip in CI"),
                pytest.mark.xfail(reason="MXFP8: TE checkpoint/backward issues; known on non-Blackwell", strict=False),
            ),
        ),

bionemo-recipes/recipes/evo2_megatron/tests/bionemo/evo2/run/test_infer.py
```diff
@@ -399,6 +518,175 @@ def test_different_prompts_produce_different_outputs(mbridge_checkpoint_path, tm
...
+        # (15 attention heads and 128-width HyenaMixer). TP>1 requires the 7b model.
+        pytest.param(1, 1, id="tp=1,cp=1"),
+        pytest.param(
+            1,
+            2,
+            id="tp=1,cp=2",
+            marks=pytest.mark.xfail(reason="CP>1 is known broken for inference", strict=False),
+        ),
+    ],
+)
```
jstjohn, line 545: Can you skip these 7b/1m test cases in CI?
```
pytest.mark.skipif(bool(os.environ.get("CI")), reason="Skip in CI"),
```

bionemo-recipes/recipes/evo2_megatron/tests/bionemo/evo2/run/test_infer.py
```diff
@@ -399,6 +518,175 @@ def test_different_prompts_produce_different_outputs(mbridge_checkpoint_path, tm
...
+        pytest.param(8, 1, id="tp=8,cp=1"),
+        pytest.param(
+            1,
+            2,
+            id="tp=1,cp=2",
+            marks=pytest.mark.xfail(reason="CP>1 is known broken for inference", strict=False),
+        ),
+    ],
+)
+def test_parallel_inference_accuracy_7b(mbridge_checkpoint_7b_1m_path, tmp_path, dna_sequences, tp, cp):
```
jstjohn, line 644: Can you add a skip for this test in CI? just add a decorator I think: 
```
@pytest.mark.skipif(bool(os.environ.get("CI")), reason="Skip 7b-1m tests in in CI")
def test_...
```

PR #1548, opened 2026-04-04
Generalize MFU/FLOPs module across recipes with log_mfu training hook

gagank1:
### Description

Adds an architecture-independent FLOPs counting and MFU tracking module (flops.py) that can be used by any native_te recipe. The module auto-detects model architecture (GQA vs MHA, SwiGLU vs standard FFN) from HuggingFace config dicts and supports analytical, simplified, and Hyena (Evo2) FLOPs formulas.                                                   

MFU tracking is integrated into all 11 training scripts across 4 recipes, gated behind a log_mfu: false Hydra flag

Some MFU results: https://linear.app/nvidia/issue/BIO-406/mfu-calculation-experiment#comment-50296297

#### Usage

During training:
```
# Any training script — just add log_mfu=true
  torchrun --nproc_per_node=2 train_fsdp2_cp.py --config-name L2_lingua_1b log_mfu=true                                   
                                                                                                                          
  Output during training:                                                                                                 
  [__main__][INFO] - MFU: 60.1% (93.19 TFLOPS/GPU) 
```

CLI tool:
```
  # Show GPU and peak bf16 TFLOPS                                                                                         
  python flops.py gpu-info                                                                                                
                                                                  
  # Compute FLOPs for any model config (auto-detects architecture)                                                        
  python flops.py flops --config-path ./model_configs/lingua-1B
  python flops.py flops --config-path nvidia/esm2_t36_3B_UR50D   # from HF Hub                                            
                                                                  
  # Estimate CP ring attention communication overhead                                                                     
  python flops.py flops --config-path ./model_configs/lingua-1B --num-gpus 2 --cp-size 2 --seq-len 16384
                                                                                                                          
  # Measure P2P GPU bandwidth (matches CP ring attention pattern)                                                         
  torchrun --nproc_per_node=2 flops.py bandwidth                                                                          
```

### Type of changes

- [ ] Refactor
- [ ] Documentation update

### CI Pipeline Configuration

- [ciflow:skip](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:skip) - Skip all CI tests for this PR
- [ciflow:notebooks](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:notebooks) - Run Jupyter notebooks execution tests
- [ciflow:slow](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:slow) - Run slow single GPU integration tests marked as @pytest.mark.slow
- [ciflow:all](https://github.com/NVIDIA/bionemo-framework/blob/main/docs/docs/main/contributing/contributing.md#ciflow:all) - Run all tests (unit tests, slow tests, and notebooks). This label can be used to enforce running all framework tests.

> [!NOTE]

#### Authorizing CI Runs

#### Triggering Code Rabbit AI Review

To trigger a code review from code rabbit, comment on a pull request with one of these commands:

- @coderabbitai review - Triggers a standard review
- @coderabbitai full review - Triggers a comprehensive review

### Pre-submit Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Model FLOPs Utilization (MFU) tracking to monitor per-step compute efficiency during training via `log_mfu=true` parameter.
  * Added FLOPs computation utilities with CLI tools for GPU peak performance detection, bandwidth measurement, and context-parallel communication overhead estimation.

* **Documentation**
  * Added "MFU Tracking" guides with usage examples for all recipe READMEs.
  * Documented `flops.py` CLI commands for performance profiling and GPU analysis.

* **Tests**
  * Added comprehensive test coverage for FLOPs calculations and MFU tracking functionality.


bionemo-recipes/recipes/llama3_native_te/compare_mfu_multigpu.py
```diff
@@ -0,0 +1,292 @@
...
+    # =========================================================================
+    total_flops_readme = compute_flops_readme(b, s, h, num_layers, vocab_size)
+    total_flops_fp, breakdown, lm_head_fwd = compute_flops_first_principles(
+        b, s, h, num_layers, n_kv_heads, head_dim, ffn_hidden_size, vocab_size
+    )
+
+    if rank == 0:
+        print("Counting FLOPs with HF model (meta device)...")
+    hf_config_meta = LlamaConfig.from_pretrained(args.config_path)
+    hf_config_meta._attn_implementation = "eager"
```
pstjohn: This is going to be the slowest possible attention, can you use fa2 or sdpa?

gagank1:
@coderabbitai full review

bionemo-recipes/recipes/llama3_native_te/compare_mfu_multigpu.py
gagank1: This was the old version of the PR, no longer includes HF comparison

ci/scripts/check_copied_files.py
```diff
@@ -199,6 +199,18 @@ def _compare_file_contents(source_file: Path, dest_file: Path, source_display: s
     "bionemo-recipes/models/codonfm/modeling_codonfm_te.py": [
         "bionemo-recipes/recipes/codonfm_native_te/modeling_codonfm_te.py",
     ],
+    # FLOPs / MFU module - synced across recipes
+    "bionemo-recipes/recipes/llama3_native_te/flops.py": [
```
pstjohn: do we have to copy these to each recipe? it's just to calculate flops? maybe we just put it in one and point folks at it. 

This should maybe be in the models/ definition though? is this just printing out a total flops for a given model size?

gagank1: flops.py has 2 usage paths:
1. flops counting for a given model config and comms bandwidth/overhead calculation using CLI
2. can be used in the recipes' train scripts with config `log_mfu=true` to log MFU with each step

so it doesn't go purely with the model defs.

bionemo-recipes/recipes/esm2_native_te/flops.py
```diff
@@ -0,0 +1,714 @@
...
+            self.breakdown = None
+            self.lm_head_flops = 0
+        elif formula == "hyena":
+            self.total_flops = compute_flops_hyena(config, batch_size, seq_len, hyena_layer_counts)
+            self.breakdown = None
+            self.lm_head_flops = 0
+        else:
+            raise ValueError(f"Unknown formula: {formula!r}. Use 'analytical', 'simplified', or 'hyena'.")
+
+        self.per_gpu_flops = self.total_flops // max(num_gpus, 1)
```
trvachov: Why divide by num_gpus? You mentioned you tested this with CP, num_gpus>2 and got reasonable numbers?

bionemo-recipes/recipes/llama3_native_te/flops.py
```diff
@@ -0,0 +1,708 @@
...
+    """Simplified formula assuming standard MHA + standard FFN with I=4H.
+
+    This is the formula from the Llama3 README:
+    (24*B*S*H^2 + 4*B*S^2*H) * 3*L + 6*B*S*H*V
+
+    The 24*H^2 coefficient assumes: 8*H^2 for 4 attention projections (MHA) +
+    16*H^2 for 2 MLP projections with I=4H.
+    """
+    b, s, h = batch_size, seq_len, hidden_size
+    return (24 * b * s * h * h + 4 * b * s * s * h) * (3 * num_layers) + (6 * b * s * h * vocab_size)
```
trvachov: Does this work with gradient accumulation?

bionemo-recipes/recipes/esm2_native_te/train_ddp.py
```diff
@@ -187,6 +200,17 @@ def main(args: DictConfig) -> float | None:
                 lr=optimizer.param_groups[0]["lr"],
             )
 
+            if mfu_tracker is not None and perf_logger.last_step_time is not None:
+                mfu_info = mfu_tracker.compute_mfu(perf_logger.last_step_time)
+                if dist_config.is_main_process():
+                    logger.info(
```
trvachov: Does this take into account there is a logging_frequency variable? It seems like this would spam this every step.

bionemo-recipes/recipes/llama3_native_te/flops.py
```diff
@@ -0,0 +1,708 @@
...
+                self.config.num_hidden_layers,
+                self.config.num_kv_heads,
+                self.config.head_dim,
+                cp_size,
+            )
+
+        if dp_size > 1:
+            # FSDP reduce-scatter estimate: ~2 * model_params * dtype_bytes * (dp-1)/dp
+            model_params = _estimate_model_params(self.config)
+            total += 2 * model_params * 2 * (dp_size - 1) // dp_size
```
trvachov: Ok...a number of concerns with this bit of code:
 
1. The second "2" --> is that supposed to be dtype_bytes = 2? That works for bf16/fp16 but we sometimes have weights in FP32.
2. The comment says there is an FSDP2 reduce-scatter comm estimate, but isn't step 1 of FSDP2 an all gather for each shared module? I am just not sure I understand the distributed comm steps "~2 * model_params * dtype_bytes * (dp-1)/dp" is trying to model.
3. Is the assumption here that you can sum up all the per-module FSDP2 comms into "model_params"? FSDP2 does per module sharding so I'm trying to understand how you get an estimate the comms without doing a sum over the modules that are being sharded.

bionemo-recipes/recipes/llama3_native_te/flops.py
```diff
@@ -0,0 +1,708 @@
...
+        """Estimate communication overhead as a fraction of step time.
+
+        Args:
+            step_time: Measured step time in seconds.
+            measured_bw_gbps: Measured P2P bandwidth in GB/s. If None, uses a default estimate.
+
+        Returns:
+            Dict with comm_bytes, estimated_comm_time, comm_pct.
+        """
+        bw = measured_bw_gbps or 6.0  # Default ~PCIe Gen3 x8
```
trvachov: Why PCIe gen3? Isn't that >15 years old?

Also -- have you done the measurement on PCIe Gen3? I ask because, if this 6.0 is a "datasheet number", it might be very hard to measure especially through pytorch because of launch overhead. You might need a very specialized environment to measure max-PCIe speeds.

bionemo-recipes/recipes/llama3_native_te/flops.py
```diff
@@ -0,0 +1,708 @@
...
+    print(f"  {'Model params':<20} {model_params / 1e9:.2f}B")
+
+
+# =============================================================================
+# CLI
+# =============================================================================
+
+
+def _cli_bandwidth():
+    """Measure P2P bandwidth. Launch with: torchrun --nproc_per_node=2 flops.py bandwidth."""
```
trvachov: Does this CLI command run for you? It just hangs indefinitely for me...

trvachov (changes requested):
Across all the added code, I think you shouldn't use `batch_size` in isolation as a variable name, or at least if you do make it explicit that in that function it doesn't matter whether it's micro or global batch. Otherwise, use `micro_batch_size` or `global_batch_size`, because it gets very confusing if the distributed flops math is right without this.

The comm overhead calculations here have a number of issues including assumptions about model precision and also about FSDP2 communication patterns. Adding comm overhead calculations is a reasonable idea but I think it should be a separate PR. Each PR should be one-idea -> one PR.

There is a flops.py CLI, but I ran `torchrun --nproc_per_node=2 flops.py bandwidth` as written in the docstring it and it just hangs. This was on 8xA100 80GB PCIe, pytorch:26.02-py3 container. Does this work for you?

Regarding overall code structure --> indeed there is a lot of duplicated code, but that is also to some degree the philosophy of recipes. @pstjohn Can you help with some guidance here? Should flops.py really be in perf_logger.py to reduce duplicate files? Also I think the FLOPS metrics should likely be recorded in WANDB just like all of the other metrics we have.

pstjohn:
A couple general thoughts:

1. The simplicity of our training scripts is incredibly important; even as it stands today i'm not happy with how much has crept in there. These need to be accessible and readable to folks coming to the repository for the first time, and be basic scaffolds agents can build from. I really don't think a flops calculator like this needs to touch every single training script in our recipes. At most, it would go in the perf logger, but even there, I'd rather see this be something like https://huggingface.co/docs/accelerate/usage_guides/model_size_estimator where this lives as a javascript app in our docs, not necessarily something that runs alongside model training.
2. Logging Flops/MFU directly is notoriously a footgun when it comes to performance, especially if we end up re-calculating model flops every iteration. Are we 100% sure we're not doing this?

bionemo-recipes/recipes/llama3_native_te/flops.py
gagank1: yes it runs in around 3 seconds for me. are you having the same issue on another machine?

trvachov: Yep, infinite hang. Log:

```
root@2eafaa0982e9:/workspace/bionemo-framework/bionemo-recipes/recipes/llama3_native_te# torchrun --nproc_per_node=2 flops.py bandwidth
W0415 13:48:45.772000 309 torch/distributed/run.py:851]
W0415 13:48:45.772000 309 torch/distributed/run.py:851] *****************************************
W0415 13:48:45.772000 309 torch/distributed/run.py:851] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed.
W0415 13:48:45.772000 309 torch/distributed/run.py:851] *****************************************

Measuring P2P bandwidth between 2 GPUs...
  GPU 0: NVIDIA A100 80GB PCIe
  GPU 1: NVIDIA A100 80GB PCIe
[rank0]:[W415 13:48:48.106866174 ProcessGroupNCCL.cpp:4084] Warning: An unbatched P2P op (send/recv) was called on this ProcessGroup with size 2.  In lazy initialization mode, this will result in a new 2-rank NCCL communicator to be created. (function operator())
[rank1]:[W415 13:48:48.112559519 ProcessGroupNCCL.cpp:4084] Warning: An unbatched P2P op (send/recv) was called on this ProcessGroup with size 2.  In lazy initialization mode, this will result in a new 2-rank NCCL communicator to be created. (function operator())
[rank0]:[W415 13:48:48.352708490 ProcessGroupNCCL.cpp:4084] Warning: An unbatched P2P op (send/recv) was called on this ProcessGroup with size 2.  In lazy initialization mode, this will result in a new 2-rank NCCL communicator to be created. (function operator())
[rank1]:[W415 13:48:48.353136693 ProcessGroupNCCL.cpp:4084] Warning: An unbatched P2P op (send/recv) was called on this ProcessGroup with size 2.  In lazy initialization mode, this will result in a new 2-rank NCCL communicator to be created. (function operator())
^CW0415 13:57:20.219000 309 torch/distributed/elastic/agent/server/api.py:739] Received 2 death signal, shutting down workers
W0415 13:57:20.220000 309 torch/distributed/elastic/multiprocessing/api.py:1010] Sending process 376 closing signal SIGINT
W0415 13:57:20.221000 309 torch/distributed/elastic/multiprocessing/api.py:1010] Sending process 377 closing signal SIGINT
^Z
[1]+  Stopped                 torchrun --nproc_per_node=2 flops.py bandwidth
```

bionemo-recipes/recipes/llama3_native_te/flops.py
gagank1: 1. yes this assumes bf16, fixing that to be generalized
2. the comment is wrong, fixing that. 
3. yes, although it can be a slight overestimation since it assumes every param is sharded.

With module i with p_i parameters, total bytes across all modules for both directions:                                                                    
                                                                                                                         
  sum_over_modules(2 * p_i * dtype_bytes * (dp-1)/dp)                                                                    
  = 2 * dtype_bytes * (dp-1)/dp * sum(p_i)                        
  = 2 * dtype_bytes * (dp-1)/dp * total_params

ci/scripts/check_copied_files.py
gagank1: Addressed. `flops.py` is gone entirely. The FLOPs math is now a ~30-line helper (`_compute_per_token_flops`) inlined at the top of each recipe's existing `perf_logger.py`, so there's no new copied-files target. MFU metrics flow through the existing `torchmetrics` → WANDB path (`train/mfu_pct`, `train/tflops_per_gpu`).

bionemo-recipes/recipes/esm2_native_te/flops.py
gagank1: Good catch — the `num_gpus` divide was wrong in the general case (double-counted for pure DP, only coincidentally right for single-GPU/CP-only setups). The new formula drops it entirely:

```
per_gpu_flops_per_step = per_token_flops(seq_len) * num_unpadded_tokens_on_this_rank
```

`num_unpadded_tokens_on_this_rank` is sourced from the batch the rank actually sees — which is already sharded by both DP (distinct micro-batches per rank) and CP (sequence chunks per rank). So no `num_gpus`, `dp_size`, or `cp_size` factor is needed: each rank's FLOPs naturally reflect its share of work. This is what lets the same code path serve DDP, FSDP2, FSDP2+CP, DDP+CP, and mFSDP without per-strategy divisors.

bionemo-recipes/recipes/llama3_native_te/flops.py
gagank1: Fixed. The old code took a static `batch_size` and multiplied once per optimizer step, which under-reported MFU by `grad_acc_steps`. The new code uses the `num_unpadded_tokens` counter already accumulated across all grad-acc micro-batches in `PerfLogger.log_micro_step()` (same counter that drives `tokens_per_second_per_gpu`), so per-step FLOPs scales with the full optimizer step's token volume regardless of grad_acc setting. For esm2 (no grad-acc in this recipe) the current batch's unpadded tokens are used directly.

bionemo-recipes/recipes/esm2_native_te/train_ddp.py
gagank1: Right — the old path did log every step and spammed stdout. Fixed: MFU is now computed *inside* the `step % logging_frequency == 0` branch of `log_step` and emitted through the same `torchmetrics` → WANDB path as loss/grad_norm/step_time. Nothing is logged between logging steps.

bionemo-recipes/recipes/llama3_native_te/flops.py
gagank1: Taking your "one idea per PR" comment — I've removed the comm-overhead code entirely from this PR (including `_estimate_comm`, `estimate_cp_comm_bytes`, `_estimate_model_params`, the dtype_bytes / FSDP2 pattern logic, and the PCIe bandwidth measurement). This PR is now scoped to MFU tracking only. Comm overhead estimation will be a follow-up PR where we can address the precision and FSDP2 sharding assumptions properly.

bionemo-recipes/recipes/llama3_native_te/flops.py
gagank1: Removed along with the comm-overhead code. Will revisit in the follow-up comm-overhead PR with a more realistic default (or with a measurement-based approach).

bionemo-recipes/recipes/llama3_native_te/flops.py
gagank1: Removed the entire `flops.py` CLI (including bandwidth, gpu-info, flops, cp-comm). All MFU functionality is now accessed via `log_mfu=true` on the existing training scripts — no separate CLI to run. This avoids the hang you hit and also addresses the "training scripts should stay simple" concern raised elsewhere.

gagank1:
@trvachov @pstjohn — ready for re-review. I took your feedback and substantially reworked the PR.

## TL;DR

Every concern you raised is addressed, and the PR is now considerably smaller and simpler:

- **~3000 lines removed** (`flops.py` × 4, `test_flops.py` × 4, `compare_mfu*.py` × 4, CLI, comm-overhead code)
- **~300 lines added** (FLOPs helper inlined in each of 4 existing `perf_logger.py` files)
- Training scripts are **back to their previous simplicity** — just a single extra kwarg on the `PerfLogger` constructor

## What changed vs. the previous revision

### Deleted
- `flops.py` in all 4 recipes — gone entirely
- `tests/test_flops.py` in all 4 recipes — gone
- The entire CLI (`gpu-info`, `flops`, `cp-comm`, `bandwidth`) — no more hanging `torchrun flops.py bandwidth`
- All comm-overhead code (`_estimate_comm`, `estimate_cp_comm_bytes`, `_estimate_model_params`, PCIe default) — split out for a future PR
- The `MFUTracker` class, `ModelFLOPsConfig` dataclass, and all its helpers
- Per-script `mfu_tracker = MFUTracker(...)` blocks and `step_start_time = time.perf_counter()` scaffolding in all 11 training scripts
- `last_step_time` attribute on esm2's `PerfLogger`
- Entry for `flops.py` / `test_flops.py` in `ci/scripts/check_copied_files.py`

### Added / consolidated
- Each recipe's existing `perf_logger.py` gets ~30 lines: a `_GPU_PEAK_TFLOPS_BF16` table, `_detect_peak_tflops_bf16()`, and `_compute_per_token_flops(config_dict, seq_len)`
- `PerfLogger.__init__` gets one new optional kwarg: `model_config_dict: dict | None = None`. When present (and `args.log_mfu=True`), the logger adds `train/tflops_per_gpu` and `train/mfu_pct` to its metrics dict
- MFU is computed **inside** the `step % logging_frequency == 0` branch, so it flows through the same `torchmetrics` → WANDB pipeline as loss/grad_norm/step_time (addresses `pstjohn`'s "WANDB like other metrics" ask and `trvachov`'s "logging_frequency spam" concern)

### The MFU formula

Simple, and the same in all 4 perf_loggers:

```python
flops_per_step = _per_token_flops * num_unpadded_tokens_on_this_rank
tflops_per_gpu = flops_per_step / step_time / 1e12
mfu_pct       = tflops_per_gpu / peak_tflops * 100
```

Why this form works everywhere without DP/CP/grad-acc factoring:

| Scenario              | Why unpadded tokens on rank are correct                                       |
| --------------------- | ----------------------------------------------------------------------------- |
| **DP (DDP/FSDP2)**    | Each rank sees its own micro-batch; `attention_mask.sum()` = rank's tokens    |
| **CP (TE/PyTorch)**   | Sequence is sharded by CP; each rank's batch is already `[B, S/cp]`           |
| **Gradient accum.**   | `num_unpadded_tokens` accumulates across grad-acc micro-batches via `log_micro_step` (same counter `tokens_per_second_per_gpu` already uses) |
| **BSHD (no packing)** | `attention_mask` marks real tokens vs. pad                                    |
| **THD (packing)**     | `attention_mask.sum()` or `input_ids.numel()` counts packed tokens            |

So: no `num_gpus` divide, no `batch_size` ambiguity (there's no `batch_size` arg at all anymore), no explicit `cp_size` / `dp_size` / `grad_acc_steps` factor.

### Training script change per recipe (this is the whole diff)

```diff
-perf_logger = PerfLogger(dist_config, args)
+perf_logger = PerfLogger(
+    dist_config,
+    args,
+    model_config_dict=config.to_dict() if args.get("log_mfu", False) else None,
+)
```

Plus deletion of the 10-line `mfu_tracker = MFUTracker(...)` block and the 6-line `if mfu_tracker is not None` block inside the loop, for each of the 11 scripts.

## Point-by-point

Individual replies are on each inline thread; consolidated here:

- **`batch_size` naming (trvachov, main review)** — resolved by removing the `batch_size` argument entirely; the formula uses per-rank unpadded tokens from the actual batch.
- **Comm overhead should be a separate PR (trvachov)** — done; removed entirely. Will follow up in a separate PR that can address the precision/FSDP2-pattern assumptions properly.
- **Bandwidth CLI hangs on 8× A100 PCIe (trvachov)** — done; CLI removed entirely.
- **`num_gpus` division (trvachov, esm2/flops.py:326)** — done; new formula doesn't need it. See math above.
- **Gradient accumulation (trvachov, flops.py:203)** — done; `num_unpadded_tokens` is the per-rank accumulated token count across grad-acc, so FLOPs/step naturally scales with grad-acc.
- **`logging_frequency` spam (trvachov, esm2/train_ddp.py:206)** — done; MFU fires only on logging steps.
- **FSDP2 comm-overhead concerns (trvachov, flops.py:390)** — moot; comm code removed.
- **PCIe Gen3 default (trvachov, flops.py:366)** — moot; comm code removed.
- **Training scripts should stay simple (pstjohn)** — done; all MFU scaffolding removed from training scripts. The only per-script change is adding one kwarg.
- **FLOPs recomputation footgun (pstjohn)** — per-token FLOPs is computed **once** in `PerfLogger.__init__` and stored; at each logging step we just do `per_token_flops * num_unpadded_tokens / step_time`. No recomputation, no model introspection at runtime.
- **"Do we need to copy `flops.py` to every recipe?" (pstjohn, check_copied_files.py:203)** — moot; no more copied file. Each recipe's existing `perf_logger.py` gets the FLOPs helper inlined.

## Verified

- Pre-commit passes (`ruff`, `ruff format`, `mdformat`, license check, copied-files, secrets).
- Unit tests for `_compute_per_token_flops` (GQA+SwiGLU, MHA+standard-FFN, seq-len scaling, linearity, vocab=0 no-LM-head) + existing `TestPerfLoggerLoss` in llama3 → 11/11 pass. codonfm existing tests → 5/5 pass.
- Smoke-tested on 2× A6000 for 5 steps per recipe/strategy: llama3 DDP + FSDP2 + FSDP2+CP, esm2 DDP + FSDP2, codonfm FSDP2, opengenome2 FSDP2 + FSDP2+CP — all emit `train/mfu_pct` and `train/tflops_per_gpu` via WANDB. Default path (no `log_mfu`) still works unchanged.
- The esm2 FSDP2+CP (`train_fsdp2_cp.py`) hits an `IndexError: list index out of range` in PyTorch's DTensor redistribute during `init_empty_weights` — reproduces without `log_mfu=true` too, so it's a pre-existing issue unrelated to this PR.

## Usage (unchanged from the outside)

```bash
torchrun --nproc_per_node=2 train_fsdp2.py log_mfu=true
```

Appears in WANDB as `train/mfu_pct` and `train/tflops_per_gpu`, alongside loss/grad_norm/step_time/etc.

Thanks for the thorough review — the PR is significantly better for it.

bionemo-recipes/recipes/codonfm_native_te/perf_logger.py
```diff
@@ -36,6 +36,69 @@
...
+    return None, name
+
+
+def _compute_per_token_flops(model_config_dict: dict, seq_len: int) -> int:
+    """Training FLOPs per token for a transformer (forward + backward = 3x forward).
+
+    First-principles matmul count: Q/K/V/O projections (GQA-aware), attention
+    logits/values (the S^2 cost expressed per-token as 4*S*H), 2-or-3-projection
+    MLP (SwiGLU detected via model_type), and LM head. The returned value is
+    multiplied by the actual unpadded token count at log time, so it naturally
```
trvachov: This behavior is very strange for two reasons:
1. This docstring is documenting behavior outside of this function, which is just a weird thing to do code wise. Why should this function document code that is outside of this function?
2. I don't think what is being said here is correct -- "returned value is multiplied by the actual unpadded token count"... I don't think this is actually happening, and even if it was, I'm not sure that's correct.

bionemo-recipes/recipes/llama3_native_te/perf_logger.py
```diff
@@ -54,6 +127,24 @@ def __init__(self, dist_config: DistributedConfig, args: DictConfig, start_step:
...
         self.logging_frequency = args.logger.frequency
 
+        # MFU setup: compute per-token FLOPs and peak TFLOPS once at init. Actual FLOPs per
+        # step are derived at log time from the tracked unpadded token count, which already
+        # reflects each rank's share under DP/CP and sequence packing.
+        self._log_mfu = bool(args.get("log_mfu", False)) and model_config_dict is not None
+        self._per_token_flops = 0
+        self._peak_tflops: float | None = None
+        if self._log_mfu:
+            self._per_token_flops = _compute_per_token_flops(model_config_dict, args.dataset.max_seq_length)
```
trvachov: A few issues here:
1. self._per_token_flops  isn't being multiplies by any sort of padded vs unpadded sequence factor, as is documented in _compute_per_token_flops docstrong.
2. Even if it was doing the above, that would simply make a correction for BSHD padding. I think this would be totally incorrect for THD sequence packing, where the attention matrix size is not O(max_seq_length^2) but rather a sum of the individual cu_seqlens fragments squared.

Here's some docs that explain this https://docs.nvidia.com/nemo-framework/user-guide/25.09/sft_peft/packed_sequence.html <--- these docs are a little old but probably right -- worth checking with TE team. @jomitchellnv @pstjohn is my understanding correct that TE does a sum of little squares for THD sequence packing?

bionemo-recipes/recipes/esm2_native_te/perf_logger.py
```diff
@@ -32,18 +32,84 @@
...
+    handles BSHD, THD (sequence packing), DP, and CP: unpadded tokens on each
+    rank already reflect that rank's share of work.
+    """
+    h = model_config_dict["hidden_size"]
+    n_heads = model_config_dict["num_attention_heads"]
+    n_kv = model_config_dict.get("num_key_value_heads", n_heads)
+    head_dim = h // n_heads
+    kv_dim = n_kv * head_dim
+    ffn = model_config_dict["intermediate_size"]
+    vocab = model_config_dict.get("vocab_size", 0)
```
trvachov, line 82: should this be vocab_size, or padded_vocab_size?

I dont think this is a big correction because usually the num_layers term is much bigger than this one, but nonetheless maybe worth getting right.

trvachov (changes requested):
I had a prior comment to replace `batch_size` with `micro_batch_size` or `global_batch_size` to make it explicit, but it seems like this code doesn't use the concept of batch_size at all and somehow does FLOPs calculations using effective sequence length as the core input. Why do this? It would make sense to me if this simplified things, but I think it obfuscates it more. e.g. I think it still doesn't get the right FLOPs number for THD sequence packing and makes the FLOPs counting code rather tricky to understand.

For non sequence packed attention, the inputs are max seq length per sequence and micro batch size * DDP ranks.
For sequence packaged attention, the inputs are max seq length for *all* the sequences and the boundaries of the sequences * DDP ranks.

Then for both of the above, there's true tokens and padded tokens. I think having this be explicit (e.g. MFU_padded and MFU_unpadded) in the token counting would make this very clear for the user and be a good recipe starting point. 

As it stand, had batch_size as a concept in the first round of this PR, which was a little confusing because it didn't distinguish between micro vs global batch size, and now we have total sequence length as a concept, but a) That is miscalculating the THD FLOPS and b) it's somewhat inconsistent with our prior variable definition e.g. in this README here https://github.com/NVIDIA/bionemo-framework/tree/trvachov/mixtral-recipes-repush/bionemo-recipes/recipes/llama3_native_te 

What do you think?

bionemo-recipes/recipes/llama3_native_te/perf_logger.py
gagank1: re: 2. Yes this is true it doesn't use the sum of the squares of each packed sequence but that was an intentional choice since the formula in the Llama3 native te recipe that was used for benchmarking doesn't either (uses B*S^2). I considered doing that but it would require some tracking inserted into the training scripts to pass around cu_seq_lens (or its sum of squares) to keep track of this. I assumed we'd prefer simplicity over modifying the training scripts more (from Peter's previous comments), should I add that in?

bionemo-recipes/recipes/esm2_native_te/perf_logger.py
gagank1: Good catch — fixed in b979eed8.

The LM-head matmul does run at `padded_vocab_size` (ESM-2: 33 → 64 for FP8/tensor-core-friendly shapes); logits get sliced back to `vocab_size` only after the compute happens (`modeling_esm_te.py:585`), so the FLOPs are real.

Scoped the correction to the hardware-view metric only:
- `mfu_pct` / `tflops_per_gpu` (useful work) — keeps raw `vocab_size`.
- `mfu_padded_pct` / `tflops_per_gpu_padded` (hardware view) — uses `padded_vocab_size` when present.

Applied uniformly across all four recipes' `perf_logger.py` via a `use_padded_vocab=True` kwarg on `_compute_non_attn_per_token_flops`. For llama3 / og2 / codonfm the configs don't set `padded_vocab_size`, so the fallback (`or vocab_size`) makes the two values identical and nothing changes in those recipes — only ESM-2 sees any delta, and it's small (≲0.01 pp of total FLOPs on 3B/15B, up to ~0.3 pp on the tiny sanity-check model) as you predicted.

bionemo-recipes/recipes/esm2_native_te/README.md
```diff
@@ -374,6 +374,28 @@ output = model(**inputs)
...
+
+Enable per-step MFU logging by adding `log_mfu=true`:
+
+```bash
+torchrun --nproc_per_node=2 train_fsdp2.py --config-name L1_3B log_mfu=true
+```
+
+Two pairs of metrics are emitted per logging interval:
+
+- `train/mfu_pct` / `train/tflops_per_gpu` — useful-work rate. Excludes padding of all kinds.
```
trvachov, line 387: Live review -- is this true?

trvachov: Can you check this against the "remainder" bit in "./bionemo-recipes/models/esm2/collator.py:871:def _pt_pad_to_multiple_of(batch: dict[str, Any], pad_to_multiple_of: int, token_pad: int, label_pad: int):"

The way you handle padded vs unpadded sequences via `cu_seq_lens_q` and `cu_seq_lens_q_padded` is reasonable, I am just concerned that something in the collator function isn't honoring that distinction. I remember running into this confusion with the Mixtral recipe in the standup, when I was surprised to see per-sequence padding (before concatenation of the sequences for THD).

@pstjohn @jomitchellnv  Do you know if `cu_seq_lens_q_padded` accounts for the `+remainder` that is added in the collator?

pstjohn: So we support two ways of padding sequences in collator.py. I'm not sure we've ever used them together, in fact we have a [post-init guard ensuring they're not](https://github.com/NVIDIA/bionemo-framework/blob/6f259a83fe4b65b44c0bc6abcc898b02e89b22a3/bionemo-recipes/models/esm2/collator.py#L95).

```
pad_to_multiple_of (int, optional): If set, pads the total sequence length to be 
    divisible by this number.
pad_sequences_to_be_divisible_by (int, optional): If set, each individual sequence
    is padded to this value.
```

**pad_to_multiple_of**
The first is how we did FP8 padding in something like ESM-2. We add a new mock sequence at the end of the THD input (adding a new entry to cu_seq_lens, labels that are all mask tokens, etc.) in order for the total sequence to be a given size. That's what's happening in in `_pt_pad_to_multiple_of`.


**pad_sequences_to_be_divisible_by**
This one is mainly there for context parallelism, where the dual-chunk algorithm assumes you can divide each sequence into 2*CP_size chunks. This one calls into TE's `pad_thd_sequences_for_cp` and generates the `cu_seq_lens_*_padded` arguments. But you can also use it for FP8 size consistency, although it may be less efficient than the previous one

gagank1: Thanks @pstjohn — traced the paths and here's what I found:

**`pad_sequences_to_be_divisible_by` (CP path): correctly handled.**
`cu_seq_lens_q` stays untouched with real per-doc lengths; `cu_seq_lens_q_padded` is added as a new key by `pad_thd_sequences_for_cp`. The perf_logger's unpadded metric reads the real lengths, padded reads the padded ones — the distinction is clean.

**`pad_to_multiple_of` (FP8 path): the distinction collapses.**
`_pt_pad_to_multiple_of` (collator.py:901-909) mutates `cu_seq_lens_q` in place to describe the appended mock sequence (it has to — FA uses `cu_seq_lens_q` to partition the `input_ids` buffer, so the array must describe the padded layout). No `cu_seq_lens_q_padded` is written. In the perf_logger, padded then falls back to `cu_seq_lens_q`, so `mfu_pct` and `mfu_padded_pct` collapse to the same value, and both include the mock sequence's `remainder²` as "real work".

**Quantitative impact:** `remainder < pad_to_multiple_of` (typically ≤15 for FP8). Extra Σ(Lᵢ²) ≤ 225 against a typical real total on ESM-2 of ~26×1022² ≈ 2.7e7 → ~10⁻⁵ inflation. The mock sequence's labels are `-100` so it doesn't contribute to the loss, and the compute FA actually executes on that block is in the measurement noise.

**Pre-concat per-sequence padding (the Mixtral worry):** I don't see that happening here. The inner HF `DataCollatorForLanguageModeling` may pad per-row in BSHD, but those pad positions are stripped via `attention_mask.bool()` indexing at collator.py:169 before THD packing, and `cu_seq_lens_q` is built from raw feature lengths in `_pt_flatten_collate`, not from the BSHD-padded shapes. So nothing sneaks in pre-concat.

**TL;DR:** not a measurable MFU bug today, but the `pad_to_multiple_of` path does lose the unpadded/padded distinction. If we want to make it rigorous, the cleanest fix is to have `_pt_pad_to_multiple_of` write the padded layout to `cu_seq_lens_q_padded` and preserve the original in `cu_seq_lens_q` (requires FA to consume the padded variant when present — worth checking compatibility). I'll also sweep the other recipes' collators (llama3 / og2 / codonfm) to confirm the same analysis holds or flag any that do per-seq padding before concat.

jomitchellnv: For "in sequence padding" we would use `cu_seqlens_padded` but I don't think we want that just for padding the remainder of a sequence vector (T in the THD).


For your question "does cu_seqlens" account for the remainer. so basically if I have a series of sequences  A and B like this
```
[A A A B B B <pad>]
```
Do I want `cu_seqlens=0,3,6` or `cu_seqlens=0,3,7`. I'm pretty sure its the later. `0,3,7`

gagank1: Thanks @jomitchellnv, that matches the constraint from offline chat too — keep `cu_seqlens_padded` reserved for the in-sequence CP case.

Consensus is to leave the current behavior and document it: the inflation is bounded by `pad_to_multiple_of²` / Σ(Lᵢ²), ≤10⁻⁵ in practice, well below measurement noise. Filed #1561 with the full analysis and two proposed fixes (both avoiding `cu_seqlens_padded`) for when/if we revisit. Added inline NOTE docstrings to `_attn_work_from_batch` in all four MFU-tracking recipes' `perf_logger.py` pointing at the issue (29121fe3).

trvachov (approved):

trvachov:
@pstjohn please review

bionemo-recipes/recipes/esm2_native_te/perf_logger.py
```diff
@@ -79,60 +265,135 @@ def __init__(self, dist_config: DistributedConfig, args: DictConfig):
         # Whether to step debug_api.step() after each step
         self.quant_stats_config = args.quant_stats_config.enabled
 
+        # Gradient accumulation tracking (accumulated over the grad-acc micro-batches of
+        # the last optimizer step in the logging window, then drained in log_step).
```
pstjohn: do we have gradient accumulation in ESM-2? I don't think we do. It's odd to force this into the perf logger, and then just call `log_step` and `log_micro_step` every step

bionemo-recipes/recipes/esm2_native_te/train_mfsdp.py
```diff
@@ -178,6 +182,9 @@ def main(args: DictConfig) -> float | None:
             loss = outputs.loss
             loss.backward()
 
+            # Record per-micro-batch metrics (loss, num_tokens, Σ(Lᵢ²), perplexity).
+            perf_logger.log_micro_step(step=step, batch=batch, outputs=outputs)
+
```
pstjohn: why are we adding this for ESM-2?

bionemo-recipes/recipes/esm2_native_te/train_ddp.py
```diff
@@ -156,7 +156,11 @@ def main(args: DictConfig) -> float | None:
         start_step = 0
         epoch = 0
 
-    perf_logger = PerfLogger(dist_config, args)
+    perf_logger = PerfLogger(
+        dist_config,
+        args,
+        model_config_dict=config.to_dict() if args.get("log_mfu", False) else None,
```
pstjohn: throughout, this could fit on the same line if you did this check inside the perf logger. In general, we want to keep these training scripts as clean as possible.

```python
    perf_logger = PerfLogger(dist_config, args, model_config=config)
```


inside perf_logger.py:

```python
if args.log_mfu:
    ...
```

Do we need args.get()? we should just make this default to false in the hydra default.yaml

bionemo-recipes/recipes/esm2_native_te/train_ddp.py
```diff
@@ -171,6 +175,9 @@ def main(args: DictConfig) -> float | None:
             loss = outputs.loss
             loss.backward()
 
+            # Record per-micro-batch metrics (loss, num_tokens, Σ(Lᵢ²), perplexity).
+            perf_logger.log_micro_step(step=step, batch=batch, outputs=outputs)
+
```
pstjohn: don't add this to esm-2 since we're not doing gradient accumulation

bionemo-recipes/recipes/esm2_native_te/perf_logger.py
gagank1: You're right — ESM-2 doesn't grad-acc. Reverted in ff0410d0: collapsed `log_micro_step` + `log_step` back into a single `log_step(step, batch, outputs, grad_norm, lr)` and dropped the dead cross-step accumulators (`running_loss`, `grad_acc_step_count`, `num_tokens`, `_attn_work_*_accum`). The Σ(Lᵢ²) attention math and the dual `mfu_pct` / `mfu_padded_pct` flavors stay (orthogonal to grad-acc). The other three recipes (llama3, og2, codonfm) keep the split since they do grad-accumulate.

bionemo-recipes/recipes/esm2_native_te/train_mfsdp.py
gagank1: Same fix in ff0410d0 — `log_micro_step` call removed from `train_mfsdp.py` (and the other four ESM-2 train scripts).

bionemo-recipes/recipes/esm2_native_te/train_ddp.py
gagank1: Done in b9f31aec. Train scripts now just call `PerfLogger(dist_config, args, model_config_dict=config.to_dict())` — the `args.log_mfu` gate moved inside `PerfLogger.__init__` across all 4 recipes (esm2, llama3, og2, codonfm). Also dropped `args.get("log_mfu", False)` for `args.log_mfu` since the default already lives in each recipe's `hydra_config/defaults.yaml`. 11 train scripts touched in total.

bionemo-recipes/recipes/esm2_native_te/train_ddp.py
gagank1: Removed in ff0410d0 — `log_micro_step` call gone from all 5 ESM-2 train scripts (`train_ddp`, `train_ddp_cp`, `train_fsdp2`, `train_fsdp2_cp`, `train_mfsdp`). The unified `log_step` reads batch + outputs directly.

PR #1553, opened 2026-04-17
Fix HFInferenceParams missing is_compileable for transformers 5.x compatibility

svc-bionemo:
## Problem

Nightly CI has been failing for 7 consecutive days (4/11–4/17) across `models/mixtral`, `models/llama3`, `models/qwen`, and `verify-recipe-tests`.

All failures share the same root cause:
```
AttributeError: 'HFInferenceParams' object has no attribute 'is_compileable'
```

The 26.03 PyTorch container update (#1540) brought `transformers==5.5.4`, which added an `is_compileable` property check on cache objects during `generate()`. The custom `HFInferenceParams` class does not implement this property.

## Fix

Added `is_compileable` property returning `False` to `HFInferenceParams` in all 6 affected files (4 model sources + 2 recipe copies). The property returns `False` since this custom TE-based cache wrapper is not compatible with `torch.compile`.

## Failing CI run
https://github.com/NVIDIA/bionemo-framework/actions/runs/24558420210

## Testing
- Pre-commit checks pass
- `check_copied_files.py` confirms all copies are in sync
- Fix is minimal and targeted (5 lines per file)

### DIFF
--- a/bionemo-recipes/models/llama3/modeling_llama_te.py
+++ b/bionemo-recipes/models/llama3/modeling_llama_te.py
@@ -591,3 +591,8 @@ class HFInferenceParams(InferenceParams):
             updated_key_cache = key_cache.index_select(0, beam_idx)
             updated_value_cache = value_cache.index_select(0, beam_idx)
             self.cache_manager.cache[layer_number] = (updated_key_cache, updated_value_cache)
+
+    @property
+    def is_compileable(self) -> bool:
+        """Return False as this cache is not compatible with torch.compile."""
+        return False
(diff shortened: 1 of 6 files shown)


pstjohn (approved):

pstjohn (approved):

pstjohn:
> Nightly CI has been failing for 7 consecutive days (4/11–4/17) across models/mixtral, models/llama3, models/qwen, and verify-recipe-tests.

that's your only job! 😠
bad 🦞

PR #1576, opened 2026-05-18
fix: guard division-by-zero and uninitialized state in metrics/masking code

mooreneural:
## Fix
- **`mlm_memmap.py` - codon weight normalization**: `pos_weight[1:-1] / pos_weight[1:-1].mean()` produced `NaN` when all token weights are zero, which propagated into `np.random.binomial` and either crashed or silently corrupted masking probabilities. Fixed by skipping normalization when the mean is zero.
- **`mlm_memmap.py` - random-replacement probability**: `random_replace_prob / (1 - mask_replace_prob)` caused `ZeroDivisionError` when `mask_replace_prob == 1.0`, and produced an out-of-range probability (> 1.0) that crashed `np.random.binomial` for other extreme combos. Fixed by guarding against zero denominator and clamping to [0, 1].
- **`dead_latents.py` - uninitialized `_last_avg_nonzero`**: `DeadLatentTracker.get_stats()` referenced `self._last_avg_nonzero`, which was only assigned inside `update()`. Calling `get_stats()` before the first `update()` raised `AttributeError`. Fixed by initializing to `0.0` in `__init__`.
- **`evo2_dataset.py` - duplicate set entry**: ASCII value `45` (`-`, gap character) appeared twice in `VALID_DNA_AND_DEGENERATE`. Python sets silently deduplicate so no runtime impact, but it is a copy-paste error. Removed the duplicate.

## Test plan

- [ ] `process_item` with `codon_weights` all zeros no longer raises or produces NaN
- [ ] `process_item` with `mask_replace_prob=1.0` and `random_replace_prob > 0` no longer raises `ZeroDivisionError`
- [ ] `DeadLatentTracker().get_stats()` can be called immediately after construction without `AttributeError`
- [ ] `Evo2Dataset.VALID_DNA_AND_DEGENERATE` still contains `45` exactly once


trvachov:
Thanks for the contribution -- running CI and reviewing.

mooreneural:
The single failing test (test_infer_phylogenetic_prompt) is a pre-existing CI flake, unrelated to any of the four changes in this PR. The error is:

torch.distributed.DistNetworkError: EADDRINUSE - port 40935 already in use
The test hardcodes --master_port 40935 in its torchrun invocation. Another process on the CI runner was holding that port during this run. Our only change to evo2_megatron was removing a duplicate entry from a Python set literal, which cannot cause a port conflict.

Could someone re-run the mbridge-unit-tests (recipes/evo2_megatron) job? A fresh run should pass.

mooreneural:
Hi team, friendly ping on this one! It's been a few weeks since my last comment, and the PR is ready to merge. Four defensive bug fixes, DCO signed, and the only CI failure is a pre-existing port flake on test_infer_phylogenetic_prompt (unrelated to these changes). Would appreciate a review when anyone gets a chance. Happy to answer any questions!

@trvachov @jstjohn

PR #1565, opened 2026-04-30
Support subquadratic-ops kernels in evo2 autoregressive inference

farhadrgh:
### Description

Closes the gap noted in `hyena_mixer.py` (`# todo: support inference_context for b2b_kernel`) and the README caveat that `--use-subquadratic-ops` "does not apply to autoregressive inference (`infer_evo2`)". After this PR, the same fused kernels that accelerate training and batch prediction also accelerate the prefill phase of autoregressive inference. 

Summary of change:

1. **`engine.parallel_fir`** now accepts `use_subquadratic_ops` and routes to `fft_causal_conv1d` (filters ≥ 128) or `causal_conv1d` (short filters), wired through both call sites in `hyena_utils.py`.
2. **`HyenaMixer.forward`** detects prefill (no FIR cache yet) and runs `b2b_causal_conv1d` for the fused proj+mixer convolution. The kernel doesn't expose its intermediate, so we run a tiny windowed proj-conv on the last `K_proj + K_mixer − 2` input positions to materialize the `(x2*v)` tail and seed the mixer's FIR cache. Works for both `hyena_short_conv` and `hyena_medium_conv`.
3. Removed the `del self._parameters["short_conv_weight"]` micro-optimization in `ParallelCausalDepthwiseConv1dWithState._get_weight()` — `B2BCausalConv1dModule` reads that raw param on every prefill, so deleting it after first decode broke multi-prompt inference. Memory cost is ~4 MB for a 1B model.

`infer_evo2` gets a `--use-subquadratic-ops` flag.

## Testing

- New parametrization `test_forward_manual[1b-8k-bf16-subquadratic-ops-flash]` covers the `(flash_decode=True, subquadratic_ops=True)` combination that was previously skipped.
- New `test_subquadratic_ops_matches_baseline` runs greedy autoregressive generation with and without `--use-subquadratic-ops` and asserts identical output — this is the strict check that Phase 2 state population is correct (a wrong cache would diverge during decode).
- Existing kernel comparison tests (`test_hyena_mixer_kernel.py`) and inference-context unit tests pass unchanged.

## Performance

`infer_evo2`, evo2/1b-8k-bf16, single A6000, multiple identical prompts in one process to amortize the one-time JIT compile cost (~15 s the first time each subq-ops kernel sees a new shape). Steady-state numbers from batches 3+:

| Prompt | Generation | Baseline | Subq-ops | Speedup |
|---|---|---|---|---|
[...]

PR #1580, opened 2026-05-22
Short sequence prefix-invariant evo2 implementation

jstjohn:
### Description
Changes:
* codex added to top level devcontainer
* bump causal-conv1d, megatron-bridge, and associated dependencies
* add test coverage for prefix invariance when running evo2 on very short sequences through inference and training

## Summary by CodeRabbit

* **Chores**
  * Updated development container configuration for improved build environment setup.
  * Added build dependencies to support enhanced model training infrastructure.

* **Bug Fixes**
  * Improved robustness of GPU kernel operations with enhanced validation checks.
  * Enhanced model compatibility across different system configurations.

* **Tests**
  * Expanded test coverage for model inference, training, and GPU kernel correctness.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -18,18 +18,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/
 
-# Install Node.js 22 LTS and the OpenAI Codex CLI.
-RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
-    && apt-get install -y --no-install-recommends nodejs \
-    && rm -rf /var/lib/apt/lists/* \
-    && npm install -g --no-fund --no-audit @openai/codex \
-    && npm cache clean --force
-
-# Default Codex to Landlock where nested namespaces are restricted.
-RUN mkdir -p /home/ubuntu/.codex \
-    && printf '[features]\nuse_legacy_landlock = true\n' > /home/ubuntu/.codex/config.toml \
-    && chown -R ubuntu:ubuntu /home/ubuntu/.codex
-
 USER ubuntu
 RUN curl https://cursor.com/install -fsS | bash  # Install cursor-agent CLI tool
 RUN curl -fsSL https://claude.ai/install.sh | bash  # Install Claude CLI tool
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/
 
 USER ubuntu
-RUN curl https://cursor.com/install -fsS | bash  # Install cursor-agent CLI tool
-RUN curl -fsSL https://claude.ai/install.sh | bash  # Install Claude CLI tool
+RUN curl https://cursor.com/install -fsS | bash || true  # Install cursor-agent CLI tool
+RUN curl -fsSL https://claude.ai/install.sh | bash || true # Install Claude CLI tool
 RUN uv tool install pre-commit --with pre-commit-uv --force-reinstall
 ENV PATH="/home/ubuntu/.local/bin:${PATH}"
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,7 +10,26 @@ RUN --mount=type=cache,target=/root/.cache/pip \
     --mount=type=bind,source=requirements.txt,target=/workspace/requirements.txt \
     PIP_CONSTRAINT= pip install -r /workspace/requirements.txt
 
+# Sandboxed agent CLIs use these helpers on Linux.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    bubblewrap \
+    uidmap \
+    && rm -rf /var/lib/apt/lists/*
+
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/
+
+# Install Node.js 22 LTS and the OpenAI Codex CLI.
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/* \
+    && npm install -g --no-fund --no-audit @openai/codex \
+    && npm cache clean --force
+
+# Default Codex to Landlock where nested namespaces are restricted.
+RUN mkdir -p /home/ubuntu/.codex \
+    && printf '[features]\nuse_legacy_landlock = true\n' > /home/ubuntu/.codex/config.toml \
+    && chown -R ubuntu:ubuntu /home/ubuntu/.codex
+
 USER ubuntu
 RUN curl https://cursor.com/install -fsS | bash  # Install cursor-agent CLI tool
 RUN curl -fsSL https://claude.ai/install.sh | bash  # Install Claude CLI tool
--- a/bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
+++ b/bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
@@ -83,6 +83,7 @@ def parallel_fir(
     fir_length,
     compute_state,
     use_subquadratic_ops=False,
+    check_subquadratic_ops=True,
 ):
     """Compute parallel finite impulse response filtering with optional state computation."""
     L = u.shape[1]  # noqa: N806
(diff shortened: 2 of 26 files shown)


bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
```diff
@@ -98,21 +101,14 @@ def parallel_fir(
...
-        else:
-            z = F.conv1d(
-                u.to(torch.float32),
-                weight.to(torch.float32),
-                bias=None,
-                stride=1,
-                padding=fir_length - 1,
-                groups=u.shape[1],  # always set to D, regardless of filter grouping
-            )[..., :L]
+        z = F.conv1d(
```
jstjohn: I'm not sure why this was the chosen path. Maybe the regular causal_conv1d path would work here...

farhadrgh (changes requested):
Two bugs, two questions, and the last section flags that this PR regresses subq-ops inference support that already landed in #1565.

## Bugs

**1. `hyena_utils.fftconv_func` fix is incomplete, bidirectional path still broken.**

The fix lands only inside the `else: # causal` branch:

```python
if use_subquadratic_ops:
    y = fft_causal_conv1d(u, k.squeeze(0))
else:
    fft_size = max(fft_size, 2 * k.shape[-1])   # <-- only here
    k_f = torch.fft.rfft(k, n=fft_size) / fft_size
```

The `if bidirectional:` branch immediately above still does `torch.fft.rfft(k, n=fft_size)` with the original `fft_size = 2 * seqlen`. Same truncation bug if anyone runs the bidirectional path with `seqlen < K`. Suggest hoisting the `max(...)` line to right after `fft_size = 2 * seqlen` so both branches benefit.

**2. The short-filter `causal_conv1d` subq path was reverted, but the xfail only covers the fused B2B path.**

Two separate code paths got removed in this PR, but the xfail (`test_b2b_causal_conv1d_module_matches_sequential_reference`) only documents one:

- `engine.parallel_fir` lost its `if use_subquadratic_ops: _subq_causal_conv1d(...)` arm in the `< 128` branch.
- `ParallelCausalDepthwiseConv1d.forward` now always uses `causal_conv1d_fn` instead of dispatching to subq when `use_subquadratic_ops=True`.

Neither of those is the fused B2B kernel, they're plain depthwise short-filter convolutions. Is the issue actually with subq's `causal_conv1d` under causal-conv1d 1.6+, or did these get caught in the same revert? If it's the latter, worth keeping them — they're the easier speedup with no fusion semantics to verify.

## Questions

**3. `@torch.compile` removed from `ImplicitModalFilter.filter`**  does the comment refer to a specific reproducer? A pointer in the comment would help future readers, and if the bad-interaction scope is narrow we may be able to keep `@torch.compile` with `dynamic=False` or wrap the offending call site in `torch.compiler.disable` instead of dropping it altogether.

**4. `hyena_block.py`** variable-arity `get_cpu_offload_context` call, clean fix for the 6-vs-7-arg drift, but `len(inspect.signature(...).parameters)` is a brittle proxy (it counts a `*args` parameter as 1, which would silently break the slice). Worth a `# tied to MCore <= 0.x` note so future readers know to revisit if MCore changes the signature again.

## Regression of #1565 (already on `main`)

This PR removes the two inference subq-ops code paths that landed in #1565 (merged 2026-04-30):

- **`engine.parallel_fir` short branch**: the `if use_subquadratic_ops: _subq_causal_conv1d(...)` arm from #1565 is removed (item 2 above).
- **`HyenaMixer.forward` prefill**: #1565 added `_populate_b2b_inference_state` and gated the fused b2b kernel on `use_subquadratic_ops`. This PR forces the gate off via `self.use_fused_b2b_causal_conv1d = False` (hardcoded), so the fused path can never fire even when the user passes `--use-subquadratic-ops`. This also disables the original training and `predict_evo2` b2b path that predates #1565.

Net effect for `infer_evo2 --use-subquadratic-ops` after this PR lands:

- The flag still routes long-filter FFT convs through subq-ops (`_subq_fft_causal_conv1d`), so the existing `test_subquadratic_ops_matches_baseline` correctness test will still pass.
- But the short-filter and fused-B2B prefill paths are gone, so the measured ~15% prefill speedup at 8K prompt on the 1B model (single A6000) goes back to zero. Users get the CLI flag without the performance it was added for.

I get why this is happening, the xfail in `test_hyena_utils.py` shows the fused B2B kernel doesn't match the reference under causal-conv1d 1.6+. That's a real kernel-side bug. But two things:

**(a)** The fix for the fused-B2B mismatch shouldn't take out the short-filter `causal_conv1d` path too. They're independent (see item 2 above). If the subq short-filter kernel is also broken under 1.6+, a passing/failing test would clarify; if it isn't broken, please keep that path.

**(b)** Disabling the fused B2B path is reasonable as a temporary measure, but hardcoding the flag to `False` makes the regression permanent until someone re-edits the file. Please make it a real config attribute so it can be flipped back on once subquadratic-ops ships the 1.6+ fix, without another PR. Suggested:

```python
self.use_fused_b2b_causal_conv1d = getattr(
    transformer_config, "use_fused_b2b_causal_conv1d", False
)
```

That way #1565's runtime behavior is recoverable via config, and we don't lose the speedup permanently. (And anyone hitting a `predict_evo2` perf regression after this lands can re-enable it for the training/predict path independently.)

farhadrgh (approved):

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
```diff
@@ -29,6 +29,11 @@
     _subq_error_msg = f"subquadratic_ops_torch not available: {_subq_import_error}"
 
 
+def _linear_causal_fft_size(input_len: int, filter_len: int) -> int:
+    """Return an FFT size that cannot alias a causal convolution prefix."""
+    return max(2 * input_len, 2 * filter_len)
```
moradza: Here you can do this,

```
if filter_len <= 2* input_len:
   return min(input_len + filter_len, 2 * filter_len)
return 2 * max(input_len, filter_len)
```

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
moradza: we can use subquadratic_ops rearrange here. [rearrange](https://nvidia-digital-bio.github.io/subquadraticOps-docs/generated/subquadratic_ops_torch.rearrange.rearrange.html#rearrange)

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/hyena_utils.py
```diff
@@ -451,10 +455,19 @@ def hyena_no_weight_decay_cond_with_embeddings(name, param):
...
+    dropout_mask,
+    gelu=True,
+    k_rev=None,
+    bidirectional=False,
+    use_subquadratic_ops=False,
+):
     """Apply a 1D convolution to the input sequence u using the filter k and the shortcut D."""
     seqlen = u.shape[-1]
-    fft_size = 2 * seqlen
+    fft_size = max(2 * seqlen, 2 * k.shape[-1])
```
moradza: same as first fft_size selection algorithm.

moradza (approved):

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/hyena_utils.py
jstjohn: fixed

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
jstjohn: Done

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/models/megatron/hyena/engine.py
jstjohn: fixed

savitha-eng (approved):

.devcontainer/Dockerfile
```diff
@@ -10,7 +10,26 @@ RUN --mount=type=cache,target=/root/.cache/pip \
...
+    && rm -rf /var/lib/apt/lists/*
+
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/
+
+# Install Node.js 22 LTS and the OpenAI Codex CLI.
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/* \
+    && npm install -g --no-fund --no-audit @openai/codex \
+    && npm cache clean --force
```
pstjohn: You can actually do `curl -fsSL https://chatgpt.com/codex/install.sh | sh` now!

pstjohn (approved):

PR #1613, opened 2026-06-06
feat(sae): Triton sparse decoder, tensor-parallel training, streaming

jwilber:
Update SAE to support larger models:
- triton kernels
- producer/consumer streaming for training
- tp


jstjohn (approved):
🛳️

trvachov (approved):

PR #1618, opened 2026-06-09
evo2_megatron: SAE-recipe base Predict empty-batch fix, shared layer-truncation helper

polinabinder1:
Two small, self-contained additions to `evo2_megatron`.

### What's here
1. **predict empty-batch fix** — skip empty batches on a DP shard boundary.
2. **`load_model_to_layer` helper** (`run/predict.py`) — a reusable model loader:
   - `full=False`: truncate to `layer` with `post_process=False` → hidden states at that layer (for activation extraction / SAE encoding).
   - `full=True`: keep all layers + the LM head (logits) for generation.

### Why
`predict.py` already reads hidden states at a chosen layer via the `--embedding-layer` CLI path; this exposes that same truncated-load as a small reusable function, so tooling can load Evo2 to a layer programmatically instead of re-deriving the model-provider / precision / distributed setup each time.

### Tests
- CPU: `layer` is required when `full=False` (fails fast, no checkpoint I/O).
- GPU (`@pytest.mark.slow`): truncated load returns a usable `(model, tokenizer)`.

### DIFF
--- a/bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/run/predict.py
+++ b/bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/run/predict.py
@@ -314,6 +314,105 @@ def initialize_inference_distributed(
         )
 
 
+def load_model_to_layer(
+    checkpoint_dir,
+    layer: Optional[int] = None,
+    *,
+    full: bool = False,
+):
+    """Load an Evo2 model from a checkpoint directory for activation extraction or generation.
+
+    The single shared loader used by the SAE recipes (streaming extraction, the inference
+    engine, and the dashboard backend) so the layer-truncation logic lives in exactly one
+    place instead of being copy-pasted into each.
+
+    Args:
+        checkpoint_dir: Evo2 checkpoint directory (resolved via ``resolve_checkpoint_path``).
+        layer: layer whose hidden states to expose (negative counts from the end). Required
+            when ``full=False``; ignored when ``full=True``.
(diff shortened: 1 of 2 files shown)


polinabinder1:
@coderabbitai full review

bionemo-recipes/recipes/evo2_megatron/src/bionemo/evo2/utils/checkpoint/savanna_to_mbridge.py
```diff
@@ -135,7 +135,7 @@ def load_savanna_state_dict(path: Path) -> dict[str, torch.Tensor]:
     Returns:
         Flat state dict with keys like 'sequential.{i}.xxx'.
     """
-    raw = torch.load(str(path), map_location="cpu", weights_only=True, mmap=True)
+    raw = torch.load(str(path), map_location="cpu", weights_only=False, mmap=True)
```
jstjohn: Is there any way around this? These kinds of things trigger security issues :/

jstjohn: Maybe we can register some layer that's missing registration or something and keep weights only set to True.

polinabinder1: Removing this - we will do the nemo to mbridge path for model loading. Will document the issue I saw in model conversion and raise it.

jstjohn (approved):

PR #1619, opened 2026-06-09
sae: config-gated dead-latent/FVU training fixes + inter-shard shuffle & pre-bias init shard sampling

polinabinder1:
## Why

Training TopK SAEs on Evo2 activations hit a severe **dead-latent** problem (a large fraction of features never fired, wasting capacity). `normalize_input` (already merged) fixed most of it; this PR adds the remaining **training-dynamics fixes we found necessary for Evo2 SAE training**.

**Every change defaults to the previous behavior and is opt-in** — so you can reproduce or continue prior training runs **exactly as before**, and enable each fix only when you want it. The training recipe opts in; both `topk` options serialize in the checkpoint config so a reloaded SAE keeps its behavior.

## Changes — default = previous behavior, opt in per flag

**1. Dead-latent inactivity counted in *total* tokens** — `dead_count_global` (default `False` = previous per-rank count)
The auxk revival fires once a latent has been inactive for `dead_tokens_threshold` (10M) tokens, but the counter advanced by *this rank's* micro-batch — so under DDP it ran `world_size`× too slow and revival kicked in `world_size`× too late (≈80M effective tokens on 8 GPUs). Opt in with `dead_count_global=True` to count total tokens (× world_size); the `all_reduce(MIN)` still means "fired on any rank ⇒ reset."

**2. Aggregate FVU + auxk loss** — `aggregate_loss` (bool, default `False` = previous per-token)
The per-token loss ratio `mean_t(mse_t / var_t)` down-weights rare high-variance tokens, starving the latents that specialize on them (notably Evo2's heavy-tailed **sink tokens**) → they die. Opt in with `aggregate_loss=True` for a batch-level ratio (which also matches the reported `var_exp` metric). This single bool also fixes the **auxk residual** end-to-end: `False` keeps the previous `x - recon + pre_bias`; `True` uses the corrected `x - recon` (the true error, not `pre_bias`-dominated).

**3. Shuffle + blend shards** — `mix_shards` (int, default `1` = previous)
Shards are written in corpus order (all prokaryota, then all eukaryota). A contiguous per-rank slice trains a rank on one kingdom then switches mid-epoch → a visible **FVU cliff**. `mix_shards=1` (default) = previous behavior (one shard at a time, contiguous slice). Set `mix_shards=N>1` to **globally shuffle the shard list** before the per-rank split (so each rank gets a cross-section) **and** buffer/blend N shards per batch (≈N shards of peak RAM).

**4. Spread the pre-bias-init sample** — `sample(num_shards=…)` (default `1` = previous single shard)
`pre_bias` is initialized to the geometric median of a sample of activations (so the SAE starts centered). A single-shard sample biases it toward whatever is first in corpus order (one kingdom) → mis-centered init → more dead latents. Set `num_shards>1` to draw the sample across that many random shards spanning the store (≈one shard of peak RAM — each sub-sampled then freed).

## How to opt in (what the Evo2 recipe sets)
```python
TopKSAE(..., aggregate_loss=True, dead_count_global=True)
store.get_streaming_dataloader(..., mix_shards=8)             # shuffle + blend 8 shards
pre_bias0 = geometric_median(store.sample(n, num_shards=8))   # sample across 8 shards
```
The training recipe (separate PR) exposes these as CLI flags (`--aggregate-loss`, `--dead-count-global`, `--mix-shards`, `--presample-shards`).

## Opt-out summary

| behavior | knob | default | opt in |
|---|---|---|---|
| global dead-token count | `dead_count_global` (bool) | `False` | `True` |
| aggregate FVU + auxk loss | `aggregate_loss` (bool) | `False` | `True` |
| shard shuffle + blending | `mix_shards` (int) | `1` | `>1` |
| spread pre-init sample | `sample(num_shards=)` | `1` | `>1` |

## Tests — `sae/tests/test_topk.py` (CPU, no GPU)
global-vs-local dead-token counting, the aggregate-FVU formula (`mse.mean()/var.mean()`), and that the opted-in flags round-trip through `_get_config()`.

### DIFF
--- a/bionemo-recipes/interpretability/sparse_autoencoders/sae/src/sae/activation_store.py
+++ b/bionemo-recipes/interpretability/sparse_autoencoders/sae/src/sae/activation_store.py
@@ -332,18 +332,25 @@ class ActivationStore:
         rank: int = 0,
         world_size: int = 1,
         max_shards: Optional[int] = None,
+        mix_shards: int = 1,
     ) -> DataLoader:
         """Get a streaming DataLoader that reads one shard at a time from disk.
 
-        Each rank gets a disjoint slice of shards. Peak RAM per rank is ~1 shard.
+        Each rank gets a disjoint slice of shards. Peak RAM per rank is ~mix_shards shards.
 
         Args:
             batch_size: Batch size for training
-            shuffle: Whether to shuffle shard order and within-shard data
+            shuffle: Whether to shuffle within-shard (and within-buffer) data
             seed: Random seed for reproducibility
             rank: This rank's index (0-indexed)
             world_size: Total number of ranks
             max_shards: Limit total shards used (for subsampling). None = all.
+            mix_shards: How many shards to blend together. 1 (default) = previous behavior
+                (one shard at a time, contiguous per-rank slice, no global shuffle). >1
+                globally shuffles the shard list before the per-rank split (so each rank gets
+                a cross-section — found needed for Evo2 training, where shards are kingdom-
+                ordered, to avoid an fvu cliff) AND buffers/mixes that many shards per batch
+                (at ~mix_shards shards of peak RAM).
 
         Returns:
             DataLoader yielding [batch_size, hidden_dim] tensors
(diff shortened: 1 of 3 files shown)


polinabinder1:
@coderabbitai review

jwilber:
Can you remove Evo2 comments from areas that aren't relevant (i.e. outside of the sharding shuffle context)

jwilber:
I'd split 2 into separate bullet points in the future, because there are a couple things going on. 

I agree with the FVU change.

Sounding out the logic here for the first bit, on the FVU loss update, for posterior sake and ensuring i understand.

So, the current loss is per-token, normalized by the token's own variance: `mean_t( mse_t / x_var_t )`. The logic here is to make every token equally good in relative-error terms.

So, say A `var_t = 100` and B `var_t = 1`, then, for our `loss = mean_t(mse_t / x_var_t)`, if `mse_t` is 1:

- A (high-variance): `1/100 = 0.01`
- B (low-variance): `1/1 = 1`

so low-variance token contributes 100x more loss.
For the same recon error, the high-variance token has 100x less pressure to improve from gradient. So far that does seem not preferred... :D 

so the update here is to use `loss = mean_all(error^2) / mean_all(x_centered^2)`, which would give a shared denominator (`100 + 1 / 2`), so then 

- A (high-variance): `1/50.5 = 0.01`
- B (low-variance): `1/50.5 = 1`.

So I guess the main benefit is that the high-variance token is no longer locally divided down by its own large denominator. So, with the new change, we'll do a better job of minimizing FVU across all tokens.

But then, I suppose this raises the question of: are there cases where we want to pay attention to high-variance behavior (or the opposite, assigning high-variance tokens high loss, so maybe use `1/var_t` as the denominator...).

jwilber (approved):
Approving. We may need to revisit the recon pre_bias stuff after, since that came from the openai paper.

PR #1583, opened 2026-05-29
evo2 SAE recipe: streaming extract.py + env-overridable sweep runner

polinabinder1:
## Read this before reviewing

This PR is **stacked on top of #1579** ([`evo2-sae-recipe`](https://github.com/NVIDIA-BioNeMo/bionemo-framework/pull/1579)). Both branches live on the same fork, so GitHub can't show a base-on-fork comparison — the diff below includes #1579's commits at the bottom. **The new work here is the top three commits only:**

| Commit | Scope |
|---|---|
| `a5b1db7` evo2 sae recipe: streaming-extract pipeline, drop pt_to_parquet shim | 1b.sh rewrite, pt_to_parquet.py deletion, README update |
| `9b7856a` evo2 sae: streaming extractor + prok+euk FASTA composer | `extract.py` (+206) and `compose_prokeuk_fasta.py` (+142), both new |
| `89ff40c` evo2_megatron predict: skip empty batches on DP shard boundary | 8-line bug fix in `predict.py` |

Net diff for the new work: **6 files, +439 / −107**. Once #1579 merges into `main`, this PR's diff auto-cleans to just these three commits.

## What changes

The recipe's extraction pipeline collapses from two steps to one. Before:

```
predict_evo2  --embedding-layer N  (FASTA in, .pt out)
        |
pt_to_parquet shim  (.pt -> ActivationStore parquet shards)
        |
train.py
```

After:

```
extract.py  (FASTA in, ActivationStore parquet shards out)
        |
train.py
```

`extract.py` monkey-patches `predict_evo2`'s writer so activations stream straight into the SAE `ActivationStore` parquet format during inference. No `.pt` intermediate, no shim. This is the pipeline used for the current 100M / 500M layer-22 prok+euk training runs — the old `.pt` path was never used by any real run, so `pt_to_parquet.py` is deleted.

## Files in the new work

- **`scripts/extract.py`** (new, +206) — streaming activation extractor. Re-uses `bionemo.evo2.run.predict` for all the heavy machinery (Megatron model load, DP/CP/TP/PP, FASTA dataloader, inference loop) but swaps the per-batch `.pt` writer for an in-process `ActivationStore`. Handles multi-rank merge with a file-based wait (`dist.barrier()` no-ops after `predict.main()` tears down the process group).
[...]

PR #1579, opened 2026-05-21
Add Evo2 1B SAE recipe + fix Savanna weights_only loading

polinabinder1:
**Evo2 1B SAE — working on Lepton**

**TL;DR:** New SAE recipe at `bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/`. ~200 lines original code + one-line bug fix in the `evo2_megatron` converter.

**Pipeline:** `chunk → convert (Savanna→MBridge) → predict_evo2 → pt_to_parquet shim → train`. Longer than esm2's `extract → train` because Evo2 is a Hyena model in Megatron-Core (no HF AutoModel path), ships as Savanna checkpoints needing conversion, and takes unbounded-length genomes that have to be chunked. The shim and chunker are the only model-specific code; `train.py` is reused verbatim from codonfm.

**Results on 100 organelle sequences (558 chunks at 8192 bp, Evo2 1B layer 12):** SAE trained in 2m 14s on one H100. Loss **0.67 → 0.045** (15× reduction). FVU **0.90 → 0.10**, variance explained **0.10 → 0.90**, monotonic. Dead latents **5.4%** at end (normal range; auxk revival is working). Encoder/decoder shapes [1920 ↔ 15360].

**Three gotchas worth flagging:**
1. **`weights_only=True` (torch ≥ 2.6) silently kills Savanna HF checkpoint loads** — exit 0, empty dir. One-line patch at `savanna_to_mbridge.py:138`.
2. **Hyena fftconv OOMs on long sequences even at micro-batch=1.** Chunk inputs to the model's trained context (8192 bp for 1B). Bonus: ~17× per-batch speedup.
3. **"0% dead latents" is a stuck reading until you cross `dead_tokens_threshold` (10M tokens)** — smoke tests always show 0%. Trust it only after the window fills.

**Reusable for the next model:** three-stage pattern (extractor → ActivationStore parquet → universal `train.py`) holds for esm2, codonfm, and evo2. Only the extractor changes per model.

## Test plan
- [ ] Smoke test: `FASTA=<small.fasta> bash scripts/1b.sh` should run end-to-end and produce `checkpoint_final.pt` with the expected encoder/decoder shapes
- [ ] Loss curve sanity: training log should show loss decreasing and var_exp increasing monotonically (up to the auxk revival event)
- [ ] Confirm the `weights_only=False` patch resolves the converter failure on torch ≥ 2.6

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1621, opened 2026-06-10
evo2 SAE recipe: streaming extract + train (7B layer-26)

polinabinder1:
## What

Adds the **Evo2 SAE recipe** under `recipes/evo2/` — the runnable pipeline that turns Evo2 activations into a trained SAE:

```
chunk FASTA  →  stream-extract (extract.py)  →  train (train.py)
```

`scripts/7b.sh` runs all three and reproduces our **layer-26 7B (`normalize_input`)** run.

## Contents
- **`extract.py`** — streaming activation extractor: reuses `predict_evo2` for the Megatron forward but writes layer-L activations **directly to a parquet `ActivationStore`** (no intermediate `.pt`). Model-agnostic (1B/7B/40B).
- **`train.py`** — trains a TopK/ReLU SAE from the activation cache. Never loads the model (reads only the cache; `--model-path`/`--layer` validate metadata only).
- **`chunk_fasta.py`**, **`7b.sh`** (orchestrator), **`pyproject.toml`**.

_(README/usage docs deferred to a follow-up, once the pipeline is merged and exercised end-to-end.)_

## Opt-in training fixes (from the merged `sae` PR #1619)
All default to the **previous behavior** — omit them to reproduce a baseline run exactly:

| flag | effect |
|---|---|
| `--aggregate-loss` | batch-level FVU + AuxK loss vs the per-token ratio |
| `--dead-count-global` | count dead-latent inactivity in total tokens (× world_size) under DDP |
| `--mix-shards N` | shuffle + blend N shards/batch (replaces the old `--shards-per-buffer`) |
| `--presample-shards N` | spread the pre-bias-init sample across N shards |

The DDP per-epoch batch cap is computed from each rank's assigned shards + `all_reduce(MIN)`, so it stays correct when `mix_shards>1` shuffles the shard list.

## Model format
Assumes a local **Evo2 MBridge checkpoint** (`--ckpt-dir`, loaded by `predict_evo2`). Getting it — **NGC pull** or nemo2→MBridge convert — is a documented **prerequisite**, not recipe code (no Savanna conversion baked in). (Contrast CodonFM, whose Encodon model is an HF/TransformerEngine `.safetensors` — different model family + runtime, hence the different extractor.)

## Duplicate code & planned consolidation
This recipe deliberately **mirrors the existing CodonFM recipe** rather than deduplicating now:

- **`train.py` is essentially the same file as `codonfm/scripts/train.py`** — byte-identical before the Evo2 skin + the four opt-in flags. Both are thin wrappers over the shared `sae.training.Trainer`.
- **`extract.py` shares its skeleton with `codonfm/scripts/extract.py`** — both run a model forward and stream into a `sae.ActivationStore`, including a near-identical **per-rank shard merge** (`_merge_temp_stores` ≈ CodonFM's `_merge_rank_stores`). Only the model-loading differs (Evo2 via Megatron `predict_evo2` vs CodonFM's HF/TE loader).

Consolidating either means factoring the shared **train-CLI** and the **parquet-shard merge** into the `sae` package **and migrating CodonFM onto them** — a cross-recipe refactor touching another recipe. To keep this PR single-concern (and avoid re-tangling, which this PR stack exists to undo), that dedup is a **planned follow-up** (alongside the duplicated-dashboard-component dedup), not part of this PR.

## Tests
Following the existing recipes' convention (CodonFM, ESM2 ship **no recipe-level tests**), the tested logic lives in the **`sae` package** — e.g. the opt-in flags' config round-trip is covered by `sae/tests/test_topk.py` (#1619). This recipe is a thin driver over that tested package. The streaming `extract.py` and the DDP shard-count path need a multi-GPU run to verify.

## Supersedes
Replaces #1579 (recipe) and #1583 (tangled extract+recipe), which also smuggled the now-merged `evo2_megatron` (#1618) and `sae` (#1619) changes. Closed both in favor of this clean, single-concern recipe carved off `main`.

### DIFF
--- a/bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/pyproject.toml
+++ b/bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/pyproject.toml
@@ -1,0 +1,25 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "evo2-sae"
+version = "0.1.0"
+description = "Sparse Autoencoders for the Evo2 DNA language model"
+requires-python = ">=3.10"
+
+dependencies = [
+    "sae",
+    "torch>=2.0",
+    "numpy>=1.20",
+    "pyarrow>=23.0.0",
+]
+
+# No package code lives here yet — the recipe is just an entry-point for
+# scripts/ that depends on the shared `sae` workspace package. Declare no
+# packages so setuptools doesn't try to discover anything.
+[tool.setuptools]
+packages = []
+
+[tool.uv.sources]
+sae = { workspace = true }
--- a/bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/scripts/7b.sh
+++ b/bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/scripts/7b.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+# Evo2 7B layer-26 SAE recipe: chunk FASTA -> stream-extract activations -> train SAE.
+# This reproduces the layer26_7B (normalize_input) run.
+#
+# Prerequisites (this recipe does NOT download or convert the model):
+#   - An Evo2 7B *MBridge* checkpoint directory (CKPT_DIR). Obtain it from NGC, e.g.:
+#         ngc registry model download-version "nvidia/clara/evo2:7b_<ver>" --dest "${WORK_ROOT}/checkpoints"
+#     (or convert a nemo2 checkpoint to MBridge with the evo2_megatron converter).
+#   - bionemo-recipes/recipes/evo2_megatron built (.ci_build.sh) with its .venv active,
+#     providing `predict_evo2`.
+#   - The `sae` workspace package importable in that same venv.
+#
+# Override any of these by exporting before invocation.
+
+set -euo pipefail
+
+EVO2_MEGATRON_DIR="${EVO2_MEGATRON_DIR:-/workspace/bionemo-framework/bionemo-recipes/recipes/evo2_megatron}"
+RECIPE_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+LAYER="${LAYER:-26}"
+# Context length the activations were extracted at (the model is context-extended; we
+# trained the SAE on 8192-bp chunks).
+CHUNK_BP="${CHUNK_BP:-8192}"
+
+# An Evo2 7B MBridge checkpoint directory (see prerequisites above).
+CKPT_DIR="${CKPT_DIR:?Set CKPT_DIR to an Evo2 7B MBridge checkpoint directory (see header)}"
+FASTA="${FASTA:?Set FASTA to the (prok+euk) input sequences}"
+WORK_ROOT="${WORK_ROOT:-/data/interp/evo2}"
+
+NPROC="${NPROC:-8}"            # GPUs / DP ranks
+MAX_TOKENS="${MAX_TOKENS:-1000000000}"
+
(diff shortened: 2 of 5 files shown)


polinabinder1:
@coderabbitai review

jwilber (approved):

polinabinder1:
Thanks — verified each finding against the current code; fixed the still-valid ones minimally, skipped the rest with reasons:

**Fixed**
- **`chunk_fasta.py` `--window`** — now validates `> 0` after parse (was a cryptic `range()` error / silent no-op).
- **`extract.py` token budget** — `max_tokens // world_size` floored to `0` for `max_tokens < world_size` and was then read as *no cap*; clamped to `≥ 1` so a positive cap is always enforced.
- **`extract.py` `--embedding-layer=26`** — the arg loop only matched the space-separated form, so the `=` form silently left `layer` at the default (wrong store metadata); now handles both.
- **`extract.py` merge overwrite** — `_merge_temp_stores` now starts `idx` past any shards already in `store_root` (metadata.json is written last, so a crash-then-`--recover` no longer overwrites moved shards).
- Also deduped a doubled license header in `extract.py`.

**Skipped (with reason)**
- **Exact per-batch token truncation** — the cap is approximate by design (per-rank, checked before each batch); a ≤1-batch overshoot is immaterial for an activation cache. The floored-to-0 bug (the real issue) is fixed above.
- **Per-rank finalize persisting model/layer** — unnecessary: the *merged* store's `metadata.json` already stamps `model_name`+`layer` from `main`'s args; the per-rank tmp metadata only needs `n_sequences` (seq count) + existence (the rank-ready poll).
- **`hidden_dim` consistency check across shards** — a single model+layer produces a uniform width, so a mismatch can't occur.
- **`train.py` positive_int/non_negative_float validators on all numeric flags** — defensive polish; these values fail fast downstream, it'd touch ~10 flags (not minimal), and the sibling CodonFM recipe (which this mirrors) has none.
- **`train.py` presample shard clamp** — `store.sample()` already clamps internally (`k = min(n_shards, max(1, num_shards))`), so it can't read out of range; and the `_load_shard(0)` default path is kept deliberately to reproduce the previous pre-bias init exactly.

polinabinder1:
Round 2 — the new items verified against current code:

**Fixed**
- **`pyproject.toml` pyarrow** — bumped `pyarrow>=10.0` → `pyarrow>=23.0.0` to match the `sae` package's contract.
- **`pyproject.toml` tqdm** — dropped: no recipe script imports it (scripts use only torch/numpy/pyarrow), so it was a dead dependency.
- **`chunk_fasta.py` `--input`/`--output` same file** — added a guard: the output is opened `"w"` (truncates) before the input is read, so identical paths would destroy the input. Rejected up front.

**Skipped (with reason)**
- **Both README findings** (` ```text ` fence, `pytest tests/` line) — the README was removed from this PR (docs deferred until the pipeline is merged + exercised), so neither applies.
- **Remove `torch`/`numpy` from deps ("rely on sae")** — kept: both are *directly imported* by the recipe scripts, so declaring them is the correct practice; their versions already match `sae`. (Only the genuinely-unused `tqdm` was removed.)
- **D103 docstrings on `parse_args`/helpers** — kept the `# noqa: D103` to match the sibling CodonFM recipe (identical convention); the module already has a docstring. Style polish, not minimal.

**Already addressed in the previous push** (re-flagged here): `chunk_fasta --window>0`, the `extract.py` budget floor-to-0 clamp, the `--embedding-layer=26` parse, and the `_merge_temp_stores` idx-past-existing guard. Still intentionally skipping: exact per-batch token truncation (approximate cap by design), per-rank finalize model/layer (the merged store stamps both from args), hidden_dim cross-shard consistency (one model+layer ⇒ uniform width), and the train.py mass-validators / presample-clamp (`sample()` already clamps; `_load_shard(0)` preserves the prior init) — details in the comment above.

PR #1622, opened 2026-06-10
evo2 SAE: inference engine + steering + server/CLI, tests, Dockerfile

polinabinder1:
## Summary
The importable **Evo2SAE inference engine** + feature steering — the base of the serve stack — with tests and a runnable (layer-cached) Docker image. A **single** Evo2 inference engine is loaded once and serves both paths: `encode` reads the residual stream off a layer-`L` forward hook; `generate` drives the same model's decode with decode-only feature steering. This PR now also includes the **FastAPI server + CLI** (folded in from #1637, which merged into this branch); the dashboard (#1623) and steering eval (#1635) build on it.

Rebased onto the post-#1633 top-level layout (`interpretability/sparse_autoencoders/`).

> **Stack update (2026-06-29):** #1637 (FastAPI server + CLI) was merged into this branch — its code (`server.py`, `cli.py`, their tests, `launch_inference.sh`) is now part of this PR. This PR's scope is therefore **engine + steering + server + CLI**.

## Architecture: one model, both paths
Earlier iterations loaded **two** copies of Evo2 — a truncated `post_process=False` model for encode/highlight and the full inference engine for generate (~1.8× the weights). This collapses to a **single** engine (`infer.setup_inference_engine`, run eager with `cuda_graph_impl="none"` so the steering hook applies):

- `load()` builds the one engine and takes `self.model = unwrap_model(comp.model)` + `comp.tokenizer` from it.
- **encode/highlight** (`_forward_hidden`) runs a normal full-sequence forward and reads layer `L` off a **forward hook** — the engine model is `post_process=True` (it produces logits for generation), so `output_embeddings` can't be used; the hook captures the same `[S, B, H]` module output the steering `clamp_hook` reads, so **encode and steer see identical activations by construction**.
- **generate** steers on `self.model.decoder.layers[L]` — the same module encode reads.

Validated end-to-end on the **1B-8k-bf16** (21/21 tests, incl. a highlight↔steer interleaving test proving no state bleed between the shared model's encode forward and decode path). 7B fidelity is the remaining gate.

### Note on the encode forward
The single engine model is `post_process=True` with no embeddings-only mode, so `encode` runs the **full** forward (all layers + the LM head) and reads layer `L` off the hook in passing — the trailing layers after `L` and the logits are computed and discarded. Cost is ~the layers past `L` (e.g. 6 of 32 on the 7B, 6 of 25 on the 1B; the vocab-512 head is cheap). Negligible for interactive encode; a ~20% overhead only for *batch* encode — and bulk extraction doesn't use this path anyway (`scripts/extract.py` reuses `predict_evo2 --embedding-layer`, which stops at `L`). An early-exit (raise from the hook to abort the forward) was left out deliberately: aborting mid-forward risks dirtying the engine's inference context (the no-bleed property) and would deadlock under tensor/pipeline parallel.

## GPU & scaling
**The engine is single-GPU.** Everything keys off one `self.device`; `serve` is a single `python -m evo2_sae.cli` process (not `torchrun`), pinned via `DEVICE` / `CUDA_VISIBLE_DEVICES`. The 7B fits on one 80 GB H100, so model-parallel isn't needed for ≤7B.

- **Model-parallel (TP/PP) is *not* supported as written.** The SAE consumes the **full** `[S,B,H]` hidden off the layer-`L` hook (`input_dim=4096`); under tensor-parallel that hidden is sharded across ranks (SAE would see a fragment), and under pipeline-parallel layer `L` lives on a single rank (the hook fires nowhere else). Supporting it needs an `all_gather` of the layer output + rank-aware hook placement — real work, only relevant for a model too big for one GPU (e.g. 40B).
- **Data-parallel *serving* scales trivially, no code change** — run N independent `serve` processes, each pinned to its own GPU (`CUDA_VISIBLE_DEVICES`) + `PORT`, behind a load balancer. Each is a full, independent engine.
- **Offline bulk extraction already runs multi-GPU** — `scripts/extract.py` under `torchrun --nproc_per_node N` is data-parallel (each rank a full model writing its own shards); this is how the 7B activation set was built.

## Contents
**Engine + steering**
- `src/evo2_sae/core.py` — `Evo2SAE`: `load → encode / encode_batch / feature_tracks / generate` (decode-only clamp via `sae.steering`) + input-sanitization guards (`_sanitize_steering`: feature-id range, clamp-magnitude cap, non-finite/`top_k`/temperature coercion). `encode_batch` is **length-bucketed** (work sorted by token length to minimize padding waste on mixed-length inputs; results un-sorted back to input order).
- **Load-time SAE/model fit check** — `load()` verifies the SAE's `input_dim` equals the model's hidden size (`_model_hidden_size` via config, or a 1-token forward) and raises a clear error on a mismatch ("wrong SAE/model pairing"), instead of a cryptic matmul failure on the first encode. *Known gap:* a wrong **layer number** with the same hidden size can't be caught here (the SAE checkpoint records no training layer) — it silently yields out-of-distribution features; `/health` surfaces the configured layer, and stamping the layer into the checkpoint at train time is a follow-up.
- `sae/src/sae/steering.py` — model-agnostic delta-clamp hook + `steer()`.

**Server + CLI** (folded in from #1637)
- `src/evo2_sae/server.py` — FastAPI app (`/health`, `/features`, `/annotate`, `/generate`) under an `/api` prefix, with an optional static-dashboard mount (used by #1623).
- `src/evo2_sae/cli.py` + `scripts/launch_inference.sh` — `serve` / `encode` / `batch` / `generate` entry points (required env vars, no hardcoded paths).

**Build / run / CI**
- `.ci_build.sh` (`env` | `install` | `all`) + `.ci_test_env.sh` — build the env by delegating to `evo2_megatron`'s own build (no fork of the pinned megatron stack), then install `sae` + this recipe into that venv. The phase arg lets the Dockerfile cache the two steps separately.
- `Dockerfile` — thin, non-forking, **layer-cached**: the ~30-min mbridge megatron build is its own layer (depends only on `recipes/evo2_megatron`), and the SAE source + editable installs are a separate trailing layer — so editing engine/SAE code rebuilds only the cheap install layer, not megatron. (+ a per-Dockerfile `.dockerignore`.)
- `tests/conftest.py` — 1B-8k-bf16 fixture (`bionemo_load` → `run_nemo2_to_mbridge`) + a synthesized tiny SAE, GPU-memory-gated; honors `EVO2_CKPT_DIR` / `SAE_CKPT_PATH` for manual / 7B runs. The GPU tests are gated by `@pytest.mark.skipif(not torch.cuda.is_available())`, so they run on a GPU box and skip otherwise.

## Dependency on `bionemo.evo2`
The engine reuses `bionemo.evo2`'s model code (the mbridge `recipes/evo2_megatron` recipe), which isn't pip-installable. `.ci_build.sh` (and the Dockerfile) build it via evo2_megatron's own script; it's intentionally **not** in `pyproject.toml`, matching the codonfm/esm2 recipes (base model is environment-provided).

## Interfaces: CLI vs serve (and why FastAPI)
The same `Evo2SAE` is exposed two ways via `cli.py`:
- **CLI one-shot** (`encode` / `batch` / `generate`) — load → do one thing → exit. Pays the model load per invocation; right for scripted/batch one-offs (`batch`: FASTA → parquet).
- **`serve`** — load **once**, serve many HTTP requests; right for interactive use (the dashboard). Carries the production machinery: single-GPU lock + bounded threadpool, 503-until-ready gate, body-size cap, and the CUDA-wedge supervisor.

The FastAPI app is built by `build_app(engine)` with the engine **injected** — prod passes a real `Evo2SAE`, the tests pass a torch-free `FakeEngine` and drive the *same app* via `TestClient` (no model, CPU-only). That DI is why `test_server.py` locks the full HTTP contract without a GPU.

**Why FastAPI** (over Flask / gRPC / Gradio-Streamlit / Triton-TorchServe): the need is a thin JSON API over one GPU-bound engine, served from the **same origin** as a React SPA, testable without a model, in one container. FastAPI gives async + a bounded threadpool (`anyio`, so concurrency caps cleanly to the single serialized GPU), Pydantic request validation, auto OpenAPI docs, and a `StaticFiles` mount for single-origin serving (`/api` + dashboard at `/`, no CORS) — all out of the box. The alternatives are either too thin (Flask/raw — re-roll all of that), UI-coupled (Gradio/Streamlit — can't cleanly serve a separate React SPA + JSON API), browser-hostile (gRPC), or heavyweight model-serving platforms (Triton/TorchServe/Ray Serve) built for autoscale/multi-model/dynamic-batching that this single-model, single-GPU, single-container tool doesn't need. Those become worth it only if deployment grows to many replicas / multi-model — at which point FastAPI sits behind them.

## Error handling / failure modes
An escalation ladder — nothing leaves the server permanently broken; worst case is a brief 503 during a respawn.
- **Bad input → 4xx, the engine never sees it.** Oversized body (>16 MB, `MAX_BODY_BYTES`) → **413** in middleware; over-context sequence → **413**; non-DNA / unknown organism / out-of-range or negative feature id / bad annotate mode / missing pick ids → **400**; wrong JSON shape → **422** (Pydantic). Some of these are **safety, not UX**: an out-of-range feature id would trip a CUDA device-side assert and a *negative* id would silently index the wrong feature via torch negative-indexing — so both are rejected *before* reaching the engine.
- **Not ready → 503.** `_require_ready()` and `/health` return **503** while the model loads (so k8s/LB readiness probes shed the pod). A startup failure is caught and logged; `/health` simply **stays not-ready** — process up, no crash loop.
- **Unrecoverable CUDA wedge → self-heal.** A device-side assert poisons the CUDA context process-wide (uncleanable in-process). `_is_unrecoverable_cuda` detects it → marks the engine not-ready (→ 503) → and if `EXIT_ON_CUDA_WEDGE=1`, the worker **exits** and `launch_inference.sh` **respawns** it (backoff, ≤10 consecutive fails, SIGTERM/SIGINT forwarded for `docker stop`/k8s). Recovery is a clean restart, host-independent.
- **Other uncaught exception → 500, server continues.** Model calls are serialized behind a lock, so one failed request can't corrupt shared state for the others.

(The dashboard's `/gene_embed` adds a partial-failure path — skip + *report* invalid sequences, all-invalid → 400 — documented in #1623.)

## How to run
```bash
# Build once from the repo root, then run with a GPU:
docker build -f interpretability/sparse_autoencoders/recipes/evo2/Dockerfile -t evo2-sae .
docker run --gpus all -it evo2-sae bash -lc "source .ci_test_env.sh && pytest tests/"
```

> **On build time / making it easier.** The engine needs `bionemo.evo2` (the mbridge
> `evo2_megatron` recipe), which isn't pip-installable — so the *first* `docker build`
> compiles the full megatron stack (megatron-bridge, causal-conv1d, …) and takes ~30 min.
> After that, the build is **layer-cached**: editing engine/SAE code re-runs only the two
> editable `pip install`s (seconds), not the megatron compile. Other shortcuts:
> - **Build once, reuse:** push the built image to a registry; coworkers `docker run` it and never rebuild.
> - **Skip the compile:** the Dockerfile's `ARG BASE_IMAGE` can point at a prebuilt evo2_megatron / bionemo image once one exists — the build then collapses to just the two `pip install`s.
> - **No container at all (dev):** inside an existing megatron env, `pip install -e sae/ && pip install -e recipes/evo2/` (what local validation does).

```python
from evo2_sae import Evo2SAE
eng   = Evo2SAE(evo2_ckpt_dir, sae_ckpt_path, layer=19).load()    # 1B layer 19 (7B: 26)
codes = eng.encode("ATGGCC...")                                    # [S, n_features], sparse (TopK)
out   = eng.generate(prompt="ATGGCC...", features=[{"feature_id": 123, "strength": 200}])
```

## Tests
There's **no dedicated CI lane** right now (deferred — it should later fold into the repo-wide recipe lane, which already runs `.ci_build.sh` + `pytest`). Run them manually:
```bash
bash .ci_build.sh && source .ci_test_env.sh   # build + activate the megatron venv
pytest tests/
```
- **CPU (no model):** `test_core.py` (engine plumbing — `top_features`, `_load_sae`, `generate` guards, the SAE/model dim check, encode_batch length-bucketing order) + `test_server.py` + `test_cli.py` (FastAPI app + CLI via a mocked `FakeEngine`) + `test_steering.py` sanitize guards + `sae/tests/test_steering.py` (exact clamp math). Quick CPU-only run without the venv: `PYTHONPATH=src:../../sae/src pytest tests/test_core.py`.
- **GPU:** `test_steering.py` — bf16 encode, generation in-distribution, steering changes the continuation (+ `compare_baseline`), batched/empty-sequence encode, max-clamp stays finite, and **highlight↔steer interleaving** (encode bit-identical across a steered generate; baseline unaffected by history). Gated by `@pytest.mark.skipif(not torch.cuda.is_available())` — runs on a GPU box (megatron venv); set `EVO2_CKPT_DIR`/`SAE_CKPT_PATH` for a specific model, else the fixtures build the 1B-8k-bf16 + a synthesized SAE.

## Base of
#1623 (dashboard) and #1635 (steering eval). *(#1637 — server + CLI — merged into this branch and is now part of this PR.)*

---
### Note: `recipes/evo2/` is co-owned with the eval stack (#1629)
This PR owns the recipe's Dockerfile / `.ci_build.sh` / `src/evo2_sae` (engine + server + CLI) + `tests/conftest.py`; the eval stack (**#1629**, which absorbed #1636) adds `eval/probing` (labelers, probe harness) and its `biopython`/`pyrodigal` deps to the **same** `recipes/evo2/`. The branches are reconciled — the `pyproject.toml` dep union is already resolved on #1629. Merge order: this PR first, then #1629 / #1635 / #1623 rebase onto `main`.


jwilber:
Have you tried this? any examples you can share/screenshots?

polinabinder1:
@jwilber This only deals with the steering backend. The visualization is in PR 1623.

polinabinder1:
@coderabbitai review

polinabinder1:
@coderabbitai review

polinabinder1:
Re-requested a fresh CodeRabbit review — the prior pass predated several fixes, so most of its findings are already resolved on this branch: CORS is now env-driven (`CORS_ORIGINS`), `/annotate` validates `mode` (400), feature IDs are range-checked before indexing the SAE, the `/generate` 503 path is covered by tests, and the FASTA header edge case was fixed when `_read_fasta` was deduped into `evo2_sae.fasta.read_fasta`. Will triage anything the re-review surfaces against current code.

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/tests/test_steering.py
```diff
@@ -0,0 +1,163 @@
...
+    H, F = 4, 2
+    torch.manual_seed(0)
+    pre_bias = torch.randn(H)
+    enc = torch.randn(F, H)
+    bias = torch.randn(F)
+    dec = torch.randn(H, F)
+    specs = [(enc[f], float(bias[f]), dec[:, f], float(f + 1)) for f in range(F)]
+
+    eng = Evo2SAE.__new__(Evo2SAE)  # bare instance — exercise the hook only
+    hook = eng._clamp_hook(specs, pre_bias)
```
trvachov: I think this attribute doesn't exist, so this test will fail. Moreover, none of these tests are hooked into CI, therefore they are all at risk of not working.

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/pyproject.toml
```diff
@@ -13,13 +13,15 @@ dependencies = [
...
-# packages so setuptools doesn't try to discover anything.
-[tool.setuptools]
-packages = []
+# The `evo2_sae` package (src/) holds the live inference engine + server + CLI;
+# scripts/ (extract, train) are standalone entry points alongside it.
+[tool.setuptools.packages.find]
+where = ["src"]
 
 [tool.uv.sources]
 sae = { workspace = true }
```
trvachov, line 28: somehow my agent is having a hard time installing this, and the culprit is this line. I think what would help here is having a Dockerfile (like the other recipes) which is then tested in CI, so the install path is tested.

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/pyproject.toml
```diff
@@ -13,13 +13,15 @@ dependencies = [
     "torch>=2.0",
     "numpy>=1.20",
     "pyarrow>=23.0.0",
+    "fastapi>=0.110",
```
trvachov: why is this in this PR? is this leaking from the frontend work?

trvachov (changes requested):
With difficulty, I was able to run some of the code in this PR. I think this needs some rework in terms of environment hygiene. The PR mentioned "evo2_megatron" venv is needed to run this but I do not know how to create this virtual environment -- I think it might be referring to the evo2_megatron recipe environment, but that one fails to build because there might be some bionemo-core issue, I think? Also one of the tests in this PR fails.

What would really help here is if all of this was hooked up into CI, so the install path and also the tests were actually running. I think then all of these issues would be quickly exposed and fixed.

jwilber (approved):

PR #1623, opened 2026-06-10
evo2 SAE recipe: feature-explorer dashboard (viz)

polinabinder1:
## Summary
React **feature-explorer dashboard** (4 tabs) over the Evo2SAE backend (#1622) — and the packaging
to ship it: the front-end can be built to static files baked into the recipe image, so **one
container serves the dashboard and the API on a single port**. Runnable end-to-end.

> **Stack update (2026-06-29):** #1637 (server + CLI) merged into the base branch, so the server this
> dashboard talks to now lives in **#1622**. This PR was rebased onto that base — references below that
> say "#1637" mean the server/`/api` mount that is now part of #1622.

Tabs: **Feature atlas** · **Sequence inspector** (`/api/annotate`) · **Generative steering**
(`/api/generate`) · **Sequence UMAP** (`/api/gene_embed`).

## How to run
Full instructions (three modes) live in [`feature_explorer/README.md`](interpretability/sparse_autoencoders/recipes/evo2/feature_explorer/README.md). In short:

```bash
# Deploy / share — one container, UI + API on :8001, no Node at runtime (dashboard is opt-in):
docker build --build-arg WITH_DASHBOARD=1 \
  -f interpretability/sparse_autoencoders/recipes/evo2/Dockerfile -t evo2-sae .
docker run --gpus all -p 8001:8001 \
  -e EVO2_CKPT_DIR=/ckpt/evo2 -e SAE_CKPT_PATH=/ckpt/sae.pt -e EMBEDDING_LAYER=26 \
  -v /ckpt:/ckpt evo2-sae scripts/launch_inference.sh serve     # -> http://localhost:8001

# Local dev — Vite hot reload (Node >=18) + backend (run from recipes/evo2); tunnel :5176:
scripts/launch_inference.sh serve
scripts/launch_dashboard.py                                     # tsh -L 5176:localhost:5176 <box>
```

## What this PR adds
- **Dashboard front-end** (`feature_explorer/`, React + Vite): 4 tabs, dark/light, search, CSV export,
  KaTeX steering equation, crash-safe localStorage, shared widgets in `components.jsx`.
- **Opt-in single-container serving** — `--build-arg WITH_DASHBOARD=1` adds a multi-stage Node build
  that compiles the front-end to static `dist/` (Node only in the build stage) and bakes it in;
  `server.py` mounts it at `/` via `DASHBOARD_DIST`. One `docker run`, no Node at runtime, no second
  process. **The default build is engine + server only** — it never pulls Node or builds the
  front-end, so an SAE-only image isn't coupled to the dashboard toolchain.
- **`/gene_embed` endpoint + `Evo2SAE.embed_bundle`** — pool sequences into per-feature vectors for the
  Sequence-UMAP tab; shared with the offline precompute so the static bundle matches the live response.
- **Per-pane descriptions + Limitations** — each tab documents its caveats in-app (static atlas + 2-D
  UMAP; ±300 steering clamp, continuation-only; single-layer encode; stochastic 2-D UMAP, pooled,
  `min_firing`-filtered, 1000-seq cap).
- **UMAP no longer silently truncates** — `/gene_embed` skips (not truncates) over-length/too-short
  sequences and anything past the 1000 cap, **reports the counts**
  (`n_received`/`n_skipped_short`/`n_skipped_too_long`/`n_dropped_over_cap`/`max_seq_len`/`max_genes`),
  and the tab shows a warning banner; all-invalid → 400 with an accounting message. Matches
  `/annotate`'s reject-don't-truncate.
- **Offline / static mode** — probes `/api/health`; with no backend, the **Feature atlas** (static
  parquets) and **Sequence UMAP** (precomputed `sequmap_embeddings.json`) work, steering/inspector hide.
  Precompute via `scripts/dashboard.py atlas|examples|embeddings`; `npm run build` → a fully static site.

## Architectural decisions
- **Thin viz over the HTTP API** — no model code in JS; every tab calls the `evo2_sae` backend (#1622).
  Keeps one validated inference path; the UI can't drift from the engine.
- **API under `/api`, frontend on `/`** — so a single origin serves both. The front-end always calls
  `/api/*`; in dev Vite proxies it through *without* a path rewrite, so dev and the single-container
  build hit identical paths (no dev/prod drift). (`/api` prefix + static mount are in the base, #1622.)
- **Dashboard build is opt-in; default image is engine-only** — the front-end is a Node *build-time*
  dependency, so gating it behind `WITH_DASHBOARD` keeps the SAE engine image buildable without npm
  network access or a working front-end, and an SAE-only user pays nothing for the viz. Node never
  reaches the runtime image regardless.
- **`embed_bundle` is a method on `Evo2SAE`** (in `core.py`) — it needs `encode_batch` + the SAE, and
  living in one place keeps the live `/gene_embed` response and the offline precompute identical. It's
  dashboard-only (nothing else in #1622 uses it), so it ships with this PR.
- **Report, don't silently drop** — UMAP surfaces what it couldn't embed rather than quietly returning a
  partial layout, consistent with the backend's reject-don't-truncate stance.
- **`/gene_embed` wire format** — one encode per sequence yields **both** mean- and max-pooled vectors
  (client toggles pooling without re-running the model), and the response ships **only** features firing
  in ≥ `min_firing` sequences (65,536 → `n_firing`, with a remap to real feature ids) as a **base64
  float32** `[n_genes, n_firing]` matrix. Shipping only the firing set is sound because TopK/ReLU codes
  are ≥ 0, so the firing set is **pooling-invariant** (mean and max agree on what's nonzero).
- **Atlas needs only the SAE, not the 7B** — `dashboard.py atlas` builds firing-rates from the cached
  activation store and the 2-D layout from the **SAE decoder geometry**, so most dashboard data is
  generatable with no big model loaded; `examples` / `embeddings` load the 7B but only over a *small*
  FASTA. (Splits cheap, model-free precompute from the expensive model-touching kind.)
- **`--layout auto` (UMAP → PCA fallback)** — UMAP gives the best clusters but needs numba (NumPy ≤ 2.3);
  `auto` falls back to PCA/t-SNE (numba-free) so the atlas runs single-env without forcing a fragile
  numba/NumPy pin on the whole recipe.

## Running the tests
No dedicated CI lane (deferred — see #1622). Run them via the recipe's build script:
```bash
bash .ci_build.sh && source .ci_test_env.sh
pytest tests/
```
- **CPU (no model):** `test_server.py` (incl. `/api/gene_embed` decodable matrix + the skip/over-cap accounting + unknown-organism 400) and `test_cli.py` via `FakeEngine`; plus #1622's contract tests. The dashboard front-end is build-validated (`npm run build`), no JS unit suite.
- **GPU:** inherited `test_steering.py`, gated by `@pytest.mark.skipif(not torch.cuda.is_available())` — runs on a GPU box, skips otherwise.

**Stacked on #1622** (which now includes the server + CLI, formerly the separate #1637).


polinabinder1:
@coderabbitai review

polinabinder1:
@coderabbitai review

jwilber:
This used to be static, right? Why is it on a live backend now? Is that optional? 

We mentioned rolling the Atlas out into it's own separate library, would this support that or require rewriting the setup of the other Atlas?

(Also, can you add some screenshots? :) )

polinabinder1:
@jwilber There is a vizualization demo here: https://5176-shzs9ka8n.brevlab.com/
This has a live back end because we have the steering and generation demo, which require that now.

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/feature_explorer/src/InfoButton.jsx
```diff
@@ -0,0 +1,78 @@
+import React, { useState, useEffect, useRef } from 'react'
```
trvachov: I think this whole file is never used??

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/feature_explorer/src/App.jsx
```diff
@@ -0,0 +1,1365 @@
+import React, { useState, useEffect, useRef, useMemo, useCallback } from 'react'
+import * as vg from '@uwdata/vgplot'
+import { wasmConnector, MosaicClient } from '@uwdata/mosaic-core'
+import { Query, sql, literal } from '@uwdata/mosaic-sql'
+import FeatureCard from './FeatureCard'
+import FeatureList from './FeatureList'
+import EmbeddingView from './EmbeddingView'
+import Histogram from './Histogram'
+import InfoButton from './InfoButton'
```
trvachov: this info button isn't used anywhere?

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/feature_explorer/src/SequenceUMAPView.jsx
```diff
@@ -0,0 +1,439 @@
...
+      setBusy(false)
+    }
+  }
+
+  // Biologist-contributed label: persist via the backend (scoped to this SAE), reflect locally.
+  async function saveLabel(fid, text) {
+    const label = (text || '').trim()
+    setEditingFeature(null)
+    try {
+      const resp = await fetch(`${BACKEND}/label`, {
```
trvachov: /label isn't defined anywhere, and I can confirm that I can't label any features in the demo dashboard https://5176-shzs9ka8n.brevlab.com/

bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/feature_explorer/src/SequenceUMAPView.jsx
```diff
@@ -0,0 +1,439 @@
...
+        r = await sb.json() // precomputed full-library bundle (same shape as /gene_embed)
+      } else {
+        throw new Error('Embedding custom sequences needs a live backend; the precomputed bundle covers the preset library only.')
+      }
+      const dec = (b64) => new Float32Array(Uint8Array.from(atob(b64), (c) => c.charCodeAt(0)).buffer)
+      const Gmean = dec(r.G_b64)
+      const Gmax = r.Gmax_b64 ? dec(r.Gmax_b64) : Gmean // back-compat if server only sends mean
+      const nf = r.n_features, ng = r.n_genes
+      const items = buildItems(Gmean, nf, ng, r.genes) // default pooling = mean
+      setBundle({ G: Gmean, Gmean, Gmax, nf, ng, meta: r.genes, items, stats: r.feature_stats, saeId: r.sae_id })
```
trvachov: @coderabbitai sae_id wont be found here right? see me other comment on this.

interpretability/sparse_autoencoders/recipes/evo2/src/evo2_sae/core.py
```diff
@@ -245,6 +245,56 @@ def encode_batch(self, seqs: list[str], batch_size: int = 8) -> list[torch.Tenso
...
+                }
+            )
+        stats.sort(key=lambda s: -s["n_firing"])
+        return {
+            "G_b64": base64.b64encode(gmean.tobytes()).decode(),
+            "Gmax_b64": base64.b64encode(gmax.tobytes()).decode(),
+            "n_features": int(gmean.shape[1]),
+            "n_genes": int(gmean.shape[0]),
+            "genes": meta_out,
+            "feature_stats": stats,
```
trvachov, line 544: @coderabbitai there is no sae_id here, but I think the frontend is expecting it here?

trvachov (changes requested):
There's a lot of disconnected pieces here (e.g. misconnected endpoints, or unused codepaths) -- see comments.

PR #1624, opened 2026-06-10
sae eval: feature-probing metrics + Evo2 probe CLI

polinabinder1:
> **Draft** — opening early for visibility; depends on nothing (sibling off the merged base). Will un-draft after the layer-26 ablation confirms the metrics + a rebase onto latest `main`.

## What

Adds the **SAE evaluation / feature-probing suite** — how we measure what an SAE's features actually encode.

- **`sae/eval/probing.py`** (shared, model-agnostic): scoring over an `ActivationBuffer` (per-token feature codes + optional dense-residual twin + per-token labels):
  - per-feature **AUROC** vs a label battery
  - **winner's-curse-corrected best-single** feature (train-select / test-eval)
  - **linear / softmax probes** (SAE vs dense twin → disentanglement / distributed concepts)
  - **InterPLM domain-adjusted F1** (per-instance recall for sparse regional features)
  - exported via `sae.eval`, beside the existing `loss_recovered` / `reconstruction` / `sparsity`.
- **Evo2 probe CLI** (`recipes/evo2/scripts/`): `probe.py` (`auroc`/`linear`/`codon-aa`/`context`/`extract`/`loss-recovered`), `evo2_buffer.py` (builds buffers), `labelers.py` (DNA labelers), `euk_windows.py` (exon/intron/CDS windows from FASTA+GFF), `probe_loss_recovered.py`.

## Design
- **Scoring is a pure function of saved buffers** → runs on CPU, no GPU/model. Only `extract`/`loss-recovered` touch a model (via a **lazy** `Evo2SAE` import), so the package + all scoring import and run **without the inference engine** — this PR is independent of the infer PR.
- **`probe_loss_recovered`** is a thin adapter onto Jared's `sae.eval.loss_recovered` (reuse, not a copy).
- **CPU test** in `sae/tests/test_probing.py` (AUROC / split / standardize / buffer round-trip), matching where the package's tests live.

## Known follow-up
`domain_f1` overlaps the ESM2 / CodonFM F1 implementations — unifying into shared `sae.eval` touches those recipes, so it's deferred (same pattern as the dashboard-dedup).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1626, opened 2026-06-10
sae: causal feature-steering primitive (delta-clamp hook)

polinabinder1:
> **Draft.** The shared steering primitive; the Evo2 ΔP harness (measure-what-steers) follows on top.

Adds **`sae/steering.py`** — the causal-steering primitive, the complement to G's (#1624) correlational probing.

- **`clamp_hook(sae, {feature_idx: value})`** + a **`steer(module, sae, clamps)`** context manager. A forward hook on the SAE's layer that re-encodes the activation, overrides features in code-space, and adds **`decode(clamped) − decode(original)`** back — the *delta* approach, so the SAE's reconstruction error cancels and only the clamped feature's decoder contribution moves the activation. `value=0` ablates; negative reverses.
- **Model-agnostic** — needs only `encode_pre_act` / `decode` / `top_k`; measure the effect by running the model with vs. without the hook (ΔP of a target token). Ported from the CodonFM steering notebooks (`make_sae_hook_clamp`).
- **CPU test** (`sae/tests/test_steering.py`): empty clamp is a no-op; a clamp shifts the activation by exactly `decode(clamped) − decode(orig)`; tuple-output modules keep their non-hidden elements.

Next (on top): an Evo2 harness that uses this to measure **ΔP(next base) under clamp** + dose-response + selectivity vs control features → the "what can we actually steer" report.

### DIFF
--- a/bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/scripts/steer.py
+++ b/bionemo-recipes/interpretability/sparse_autoencoders/recipes/evo2/scripts/steer.py
@@ -1,0 +1,121 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-Apache2
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


polinabinder1:
Superseded by the split stack: shared sae.steering primitive in #1629 + the evo2 steering harness in #1631 (stacked on #1629). Closing in favor of that structure.

PR #1635, opened 2026-06-12
evo2 SAE steering analysis: dose-response / selectivity harness + tested metrics

polinabinder1:
## Summary
Evo2 SAE **steering analysis** — a dose-response + selectivity harness over the decode-only steering hook, with pure, CPU-tested metrics. Now lives in **`evo2_sae.eval.steering`** (alongside `evo2_sae.eval.probing`).

**Stacked on #1622** (uses the `Evo2SAE` engine + its `sae.steering` clamp hook).

## Contents — `evo2_sae.eval.steering`
- **pure metrics:** `common_prefix_len`, `edit_distance`, `divergence`, `dose_response`, `selectivity` — string math, no model.
- **engine-driven:** `pick_target` (top labeled feature on a sequence) and `run_steering` (sweep clamp strengths on a target + control features → dose-response + selectivity).
- `scripts/steer.py` — thin CLI wrapper (builds an `Evo2SAE`, runs the sweep); imports the harness from `evo2_sae.eval.steering`.

## Design decisions (the non-trivial ones)
- **Edit distance, not positional Hamming.** Steering decodes greedily (`temperature=0`), so a single early insertion/shift makes *every* downstream position differ — Hamming would saturate to ~1.0 from one change and report a fixed sweep as maximally divergent. `divergence` uses normalized **edit distance** (doesn't saturate on a shift) plus `first_divergence` (shared-prefix length), so the metric tracks the real magnitude of the effect.
- **The engine is *injected*, not imported.** `pick_target`/`run_steering` take the engine as an argument, so the whole harness is torch-free and runs against a fake in tests — no GPU, no checkpoints, no `evo2_sae.core` import. (The fake is intentionally *local* to the test file, not a conftest fixture, so it doesn't collide with #1637's GPU fixtures.)
- **Drives the production `Evo2SAE.generate` path** (decode-only clamp) — the analysis can't drift from how steering actually behaves in the server. It also forces **greedy/deterministic** decode (`temperature=0`, `top_k=1`), so strength-0 reproduces the baseline exactly and dose-response isn't polluted by sampling noise.
- **`max_clamp` mirrors `core._sanitize_steering`** — requested strengths beyond `MAX_CLAMP_STRENGTH` are capped, and the result surfaces the capped values, so the sweep matches what the engine would actually apply.
- **`pick_target` prefers the top *labeled* feature** (the interpretable target) while still surfacing the top-*active* feature in the returned table — so steering a meaningful concept is the default, not whatever fired hardest.

## How to run
```bash
bash .ci_build.sh && source .ci_test_env.sh        # or: PYTHONPATH=src:../../sae/src
pytest tests/test_steer_analysis.py
# CLI: python scripts/steer.py --evo2-ckpt-dir <mbridge> --sae-checkpoint <sae.pt> --layer 26 --sequence ATGC...
```
No dedicated CI lane (deferred — see #1622; CI should fold into the repo-wide recipe lane later).

## Tests
**9 passed (CPU, no model):** the string metrics (`divergence` / `dose_response` / `selectivity` / `edit_distance`) against hand-computed references, plus `pick_target` / `run_steering` driven against an injected torch-free fake engine. No GPU tests (the real-engine steering path is covered by `test_steering.py` in #1622).

## Scope & limitations (and future work)
This harness is a **steering smoke test**: `dose_response` measures the *magnitude* of a clamp's effect, and `selectivity` measures whether it beats a few *hand-picked* control features. It deliberately does **not** check:
1. **Direction** — that the output moved toward the target feature's *labeled concept* (a strong-but-wrong-direction feature still scores well).
2. **Robust specificity** — `selectivity` is only as good as the chosen controls (no activation-matched null).

**Future work — the real validation:** concept-density scoring (run the feature's own labeler on the steered output; does the labeled concept rise with strength?) vs an activation-matched null (z-score/percentile). ~150 lines, self-contained in `evo2_sae.eval.steering`, stays torch-free. That's the check that answers "steering does what the label says, specifically."


polinabinder1:
@coderabbitai review

PR #1663, opened 2026-06-24
support lepton launching + add train stream script for esm2

jwilber:
Adding lepton support for sae recipes.

Starting with ESM2 SAE, which I added producer/consumer support (`train_streaming` here) so no activation dataset needs to be extracted/saved during training.

### DIFF
--- a/ci/lepton/model_convergence/configs/recipes/esm2_sae_smoke.yaml
+++ b/ci/lepton/model_convergence/configs/recipes/esm2_sae_smoke.yaml
@@ -1,0 +1,152 @@
+# @package _global_
+defaults:
+  - /base
+  - _self_
+
+############################################################
+# lepton job info
+############################################################
+node_group: yo-bom-lepton-001
+mount_from: node-nfs:fs1
+num_nodes: 1
+device_type: gpu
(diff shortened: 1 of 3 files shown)


polinabinder1 (approved):
LGTM once checks pass

trvachov:
Is this the same pathway we do for other recipes integrations to Lepton? or this is a brand new pathway?

jwilber:
@trvachov Same pathway. the extra code you see here is because we don't want to save activations for CI, because those datasets are so large. so instead, we extract + train as we go, never materializing them to disk

ci/lepton/model_convergence/configs/recipes/esm2_sae_smoke.yaml
```diff
@@ -0,0 +1,152 @@
...
+defaults:
+  - /base
+  - _self_
+
+############################################################
+# lepton job info
+############################################################
+node_group: yo-bom-lepton-001
+mount_from: node-nfs:fs1
+num_nodes: 1
```
jwilber, line 11: Let's do 2 gpus

PR #1667, opened 2026-06-30
ci: GPU matrix lane for interpretability SAE recipes (merge-first, presence-guarded)

polinabinder1:
## What
A GPU CI **matrix lane** for the SAE interpretability recipes under `interpretability/sparse_autoencoders/recipes/*`. Modeled on the repo-wide `unit-tests-recipes.yml`: a cheap ubuntu `changed-dirs` job discovers which interp recipes to test and emits a matrix; a per-recipe `unit-tests` job runs each on an L4 (its own `.ci_build.sh` + `pytest tests/`).

## Which recipes are eligible
Any interp recipe that has its own **`.ci_build.sh`**. Today that's **evo2**. `codonfm`/`esm2` have no `.ci_build.sh`/tests yet, so they're **skipped until they add them** — at which point they auto-join this lane, no workflow change needed. (This is also the **presence guard**: before the evo2 SAE recipe lands, #1622, nothing is eligible → the whole lane is a green no-op, so it can merge first.)

## What runs, when
| What the PR changes | Which recipes run |
|---|---|
| a recipe's own dir — `…/recipes/<X>/**` | just **`<X>`** |
| the shared `sae` lib — `…/sparse_autoencoders/sae/**` | **all** eligible recipes (they all depend on sae) |
| this workflow file | **all** eligible recipes |
| **nightly** `schedule` @ 09:00 UTC | **all** eligible recipes |
| anything else (other recipes' dirs, docs, …) | **none** — empty matrix, green no-op |

- **Megatron *build* is per-recipe — only evo2 builds it.** Each recipe's own `.ci_build.sh` owns its build: **evo2** compiles/installs the mbridge `bionemo.evo2` (megatron) stack; a future HF-native **esm2** or custom **codonfm** builds its own thing, *no megatron*. So a codonfm/esm2 change never pays for the Evo2 megatron build — it runs only on an evo2 change (or an `sae`/nightly run, which tests every consumer, evo2 included).
- **Container *image* is currently shared.** All matrix entries use one base image (`svcbionemo023/…pytorch26.04-py3-squashed`, dockerhub-cached → cheap after first pull) — the megatron-capable base evo2 needs. A future HF-native recipe *runs in* that base but doesn't build megatron on top. If a recipe later wants a lighter image, the matrix can carry a **per-recipe `image`** (like the repo-wide lane's `matrix.recipe.image`) — not needed while evo2 is the only active recipe. Net: **shared base image, per-recipe (megatron-or-not) build.**
[...]

PR #1668, opened 2026-06-30
feat(evo2-sae): streaming smoke test (producer/consumer) + Lepton convergence config

polinabinder1:
Evo2 analogue of the ESM2 SAE smoke (#1663). Adds a **producer/consumer** streaming path — Megatron `predict` forward passes feed activations into a bounded queue (`sae.streaming`); the SAE `Trainer` consumes them in-process, **no activation store on disk**. Self-contained like the ESM2 smoke: the 1B checkpoint is fetched by identifier and converted in-job (nothing pre-staged).

**2 GPUs (producer/consumer, not data-parallel):** when a second GPU is visible, the Evo2 producer runs its forwards on `cuda:0` and the SAE consumer trains on `cuda:1`, so the two stages run concurrently instead of contending for one device (`num_devices: 2`, `resource_shape → gpu.2xh100-sxm`). It stays one `predict` rank — `--nproc_per_node 1`, `--dp-size 1`; activations cross the boundary as CPU tensors via the queue, so no NCCL/DDP.

### Files
- `scripts/train_streaming.py` — `Evo2ActivationProducer` bridges `predict`'s per-batch callback to a pull-generator (daemon thread + queue); SAE trains off the stream. Auto-places the SAE consumer on a 2nd GPU when available; wraps the `predict` forward in bf16 autocast (see Robustness).
- `scripts/prepare_1b_checkpoint.py` — `bionemo_load("evo2/1b-8k-bf16")` → `run_nemo2_to_mbridge` (idempotent). Honors `EVO2_CKPT_DIR`.
- `ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml` + `convergence-tests.yml` — 2×H100 convergence config, plus the `evo2_sae_smoke` dispatch option.

### Review feedback addressed
- **Deduped against `/base`** — dropped `container` (identical 26.02 image), `node_group`, `mount_from`.
- **Dropped TE-recipe markers** (`te_enabled`/`fp8_enabled`/`extras`) — the run_script never references them (verified the launcher doesn't either).
- **Bumped to 2 GPUs** — via the producer/consumer split above (`gpu.2xh100-sxm`).
- **No Docker** — `checkout_script` runs `evo2_megatron/.ci_build.sh` and loads an MBridge checkpoint; nothing uses Docker.

### Robustness
- **`--input-dim` is validated** against the layer's true residual width on the first streamed chunk — a mismatch fails with a clear message instead of an opaque matmul error inside the SAE encoder.
- **bf16 autocast around the `predict` forward** — Evo2's params are bf16 and `predict`'s minimal-arg path sets no autocast region, so a `fp32_residual_connection=true` checkpoint (e.g. a 7B `EVO2_CKPT_DIR` override) would otherwise trip a TransformerEngine param/input dtype assertion. Harmless no-op on the default `1b-8k-bf16` (bf16-residual) path.
- **`--init-pre-bias` is scoped out of the streaming path** and fails fast pointing at `--no-init-pre-bias` (it would require a second Evo2 `predict` pass, which doesn't compose with Megatron's process-global init). A single-pass implementation is a natural follow-up.

### Validation
- **Streaming train (Evo2-1B / layer 12):** NGC pull + MBridge convert, then 20 steps → loss ↓ to ~0.94, `dead_latents 0%`, `checkpoint_final.pt` written.
- **2-GPU split validated:** on 2×H100 the run places producer→`cuda:0` / consumer→`cuda:1` and trains to a **loss identical to the single-GPU run to 4 decimals** — the split changes device placement, not the math.
- **CPU unit tests** (`tests/test_streaming.py`, fake `predict` module — no GPU/Megatron): queue semantics (backpressure, sentinel, exception propagation), the `--input-dim` mismatch, and the `--init-pre-bias` rejection.
- **Lepton (end-to-end, 2×H100):** dispatched to `vfco61g2` / `yo-bom-lepton-001`, `gpu.2xh100-sxm`; built via `.ci_build.sh`, ran Evo2 `predict` on GPU, streamed-trained 20 steps → **`SMOKE OK`** (`checkpoint_final.pt` written).

### Launching on Lepton
Runs on the existing convergence system (`convergence-tests.yml` → `launch_job.py`), **not** the #1667 L4 unit-test lane. This PR adds `evo2_sae_smoke` to the `workflow_dispatch` `model_config` options.

**Dispatch:** Actions → *BioNeMo Model Convergence Tests* → *Run workflow* → `model_config = evo2_sae_smoke`, `gpu_type = h100-sxm`. Success = `checkpoint_final.pt` written. **Dispatch-only** (a 20-step liveness smoke, telemetry off) — deliberately not on the biweekly cron.

### Ordering
Stacked on **#1622** (`pbinder/evo2-sae-serve`), which adds the recipe's `.ci_build.sh`. Also uses `chunk_fasta.py`, already on `main`. Review/merge after #1622. Do **not** fold into #1667 (the presence-guarded unit-test lane, which must merge before the recipe).


ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
```diff
@@ -0,0 +1,208 @@
+# @package _global_
+defaults:
```
polinabinder1, line 2: Make sure not redundant to base

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
```diff
@@ -0,0 +1,208 @@
...
+
+############################################################
+# Container
+# Base NGC PyTorch image (same as the other convergence recipes). Its CUDA
+# forward-compat libs run on the workspace's current node driver (CUDA 12.4);
+# the evo2 GPU-lane image (pytorch26.04-squashed) needs driver >= 12.8 and dies
+# on those nodes with "Inference requires CUDA". Megatron/TransformerEngine are
+# added in-job by .ci_build.sh (checkout_script), so this base image is enough.
+############################################################
+container:
```
jwilber: these are in the base config so can remove

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
```diff
@@ -0,0 +1,208 @@
...
+############################################################
+# lepton job info
+############################################################
+node_group: yo-bom-lepton-001
+mount_from: node-nfs:fs1
+num_nodes: 1
+device_type: gpu
+num_devices: 1
+gpu_type: h100-sxm
+resource_shape: gpu.h100-sxm
```
jwilber: bump to 2 gpus: (resource shape needs to change for the)

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
```diff
@@ -0,0 +1,208 @@
...
+############################################################
+# recipe identifiers
+############################################################
+recipe_subdir: interpretability/sparse_autoencoders/recipes/evo2
+model_type: evo2_sae
+variant: smoke
+framework: sae
+precision: bf16
+te_enabled: true
+fp8_enabled: false
```
jwilber: we don't need these markers for these tests (these are for TE recupes)

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
```diff
@@ -0,0 +1,208 @@
...
+  if [ -n "${commit_sha}" ]; then
+    echo "Checking out commit: ${commit_sha}"
+    git checkout "${commit_sha}"
+  elif [ "${branch}" != "main" ]; then
+    echo "Checking out branch: ${branch}"
+    git checkout "${branch}"
+  fi
+  # Builds evo2_megatron's venv (bionemo.evo2 / predict) + installs sae + evo2_sae on top.
+  cd interpretability/sparse_autoencoders/recipes/evo2
+  bash .ci_build.sh
```
jwilber, line 105: Docker in Lepton must run on a privileged node (make sure we have that set)

jwilber: It looks like instead we actually are loading a checkpoint, so we don't need docker:

```
Successfully installed evo2-sae-0.1.0 fastapi-0.139.0 starlette-1.3.1 uvicorn-0.49.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.
Reusing existing MBridge checkpoint: /data/interpretability/sae/evo2_smoke/outputs/evo2_1b_mbridge/iter_0000001
Chunked 4 sequences -> 4 chunks (16,384 bp) at window=8192
Using device: cuda
SAE: topk, input_dim=1920, hidden_dim=15360
```

jwilber: (from lepton logs: https://dashboard.dgxc-lepton.nvidia.com/workspace/vfco61g2/compute/jobs/detail/evo2-sae-1b-smoke-59w4/replicas/list#/job/evo2-sae-1b-smoke-59w4/logs
)

jwilber: It looks like the ci build actually comes from this: https://github.com/NVIDIA-BioNeMo/bionemo-recipes/blob/main/recipes/evo2_megatron/.ci_build.sh

Thus, if you're able to load a mbridge checkpoint and run evo2 inference (like what seems to be done here), maybe your other scripts don't need Docker to run? We don't use Docker here, can you double-check.

jwilber:
Run logs:
This pr edits this gh action script: https://github.com/NVIDIA-BioNeMo/bionemo-recipes/blob/aabc352953793ea7470ae385cb04e083f6f1ea54/.github/workflows/convergence-tests.yml

Which triggers runs on  Lepton. See example run: https://dashboard.dgxc-lepton.nvidia.com/workspace/vfco61g2/compute/jobs/detail/evo2-sae-1b-smoke-59w4/replicas/list#/job/evo2-sae-1b-smoke-59w4/logs

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
polinabinder1: Done — removed `container` (identical 26.02 image), `node_group`, and `mount_from`; all inherited from `/base`.

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
polinabinder1: Done — removed; inherited from `/base`.

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
polinabinder1: Done — `num_devices: 2` + templated `resource_shape` (→ `gpu.2xh100-sxm`). Implemented as a producer/consumer split rather than data-parallel: the Evo2 producer runs on `cuda:0` and the SAE consumer trains on `cuda:1` (activations cross as CPU tensors via the queue), so it stays one predict rank / `--dp-size 1`. Validated on Lepton — the job ran on `gpu.2xh100-sxm` and hit `SMOKE OK`, loss identical to the 1-GPU run.

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
polinabinder1: Done — removed `te_enabled`/`fp8_enabled`/`extras`; the run_script never references them and the launcher does not either, so it is launcher-safe.

ci/lepton/model_convergence/configs/recipes/evo2_sae_smoke.yaml
polinabinder1: Confirmed — no Docker. `checkout_script` runs `recipes/evo2_megatron/.ci_build.sh` and the run loads an MBridge checkpoint (the log you pasted is that path). Separately hardened the forward with a bf16 autocast wrap so an `fp32_residual_connection=true` ckpt override does not trip the TE dtype assertion.

Issue #1681, opened 2026-07-11
[Feature] Inference-side PTQ / weight-compression recipe for bio models (ESM-2 / Geneformer / Evo2) — wanted? where should it live?


huthvincent:
### Problem & Motivation

bionemo-recipes already has solid **training-time** low-precision support (TransformerEngine FP8 / MXFP8 / NVFP4 — e.g. #1484, #1500, `recipes/fp8_analysis`, `recipes/geneformer_native_te_mfsdp_fp8`). What I couldn't find anywhere in the repo is an **inference-side post-training quantization (PTQ) / weight-compression** example:

- nothing wiring the bio models to NVIDIA **ModelOpt** (`mtq.quantize`) for PTQ,
- no real low-precision weight *storage* (actual memory reduction) or KV-cache compression for inference,
- `recipes/vllm_inference/esm2` runs inference but does no quantization.

I've been prototyping exactly this on a fork and would like to contribute it back. Before I invest in making it fully recipe-compliant, I'd rather ask whether it's wanted and where you'd want it to live than open a large PR blind.

### BioNeMo Recipes Version

Forked off `main` around commit `4e54aaa`; I'll rebase onto current `main` before opening any PR.

### Category

Inference

### Proposed Solution

A self-contained recipe that provides:

1. **ModelOpt PTQ bridge (simulated Q/DQ)** — an adapter connecting BioNeMo model loading to `mtq.quantize()`, plus an automated quality harness comparing BF16 baseline vs. quantized logits (cosine similarity / top-k agreement / MSE). Covers ESM-2, Geneformer, and Evo2 (Evo2 needs subprocess isolation because of its Hyena/SSM non-deterministic init).
2. **Real low-precision weight storage** — drop-in `nn.Linear` replacements that store weights at INT8/FP8/INT4 for genuine memory savings (dequant-on-the-fly). INT8 per-channel is the part that currently holds up; FP8/INT4 are experimental (honest results below).

Guidance I'm looking for:

1. **Do you want an inference PTQ / weight-compression recipe in this repo at all**, or is low-precision *inference* intended to go through the NeMo + TensorRT-LLM export path outside bionemo-recipes?
2. **Where should it live?** A new `recipes/{model}_{framework}_quant/` (e.g. `recipes/evo2_megatron_quant/`), or folded into an existing recipe such as `recipes/evo2_megatron/` / `recipes/esm2_native_te/`?
[...]

PR #1682, opened 2026-07-12
feat(recipes): add evo2_megatron_quant — INT8 weight compression + ModelOpt PTQ harness

huthvincent:
### Description

Adds a new inference-side recipe, `recipes/evo2_megatron_quant/`, for BioNeMo biological foundation models (ESM-2, Geneformer, Evo2). Upstream already covers low-precision **training** (TE FP8/MXFP8/NVFP4); this adds an **inference** path:

- **Real INT8 weight compression** — a drop-in `nn.Linear` replacement (`INT8Linear`, per-channel symmetric) that stores weights in int8 and dequantizes on the fly. Measured on Evo2 7B: **~15% GPU-memory reduction, cosine sim ≈ 0.998, top-1 100%** vs BF16.
- **ModelOpt PTQ harness** — bridges BioNeMo model loading to `mtq.quantize()` with an automated quality report (cosine sim / top-k agreement / MSE).
- **CPU-only L0 sanity test** (`tests/test_compressed_linear_roundtrip.py`) — no GPU or checkpoint required, so it runs in fast CI.

This is a **Draft** opened to get maintainer direction on the questions in #1681:
1. Do you want an inference PTQ / weight-compression recipe in this repo, or should low-precision inference go through NeMo + TensorRT-LLM?
2. Is `recipes/evo2_megatron_quant/` the right home, or should this fold into `recipes/evo2_megatron/`?
3. I intentionally scoped this first PR to the parts that hold up (INT8 + PTQ harness). FP8/INT4 real weight compression and KV-cache experiments are held back for a follow-up once their GPU quality is re-measured.

> Note: the INT8 quantization math is verified numerically and the CPU L0 test is ready, but I have not yet executed the pytest in a torch environment — it should run under CI once authorized. Happy to align file layout (e.g. flatten `src/`), pin `requirements.txt` to the target container, and add a Dockerfile/hydra config to match your recipe conventions.

Follows up on #1681.

#### Usage

```python
from src.compress_model import compress_model

stats = compress_model(evo2_model, precision="int8")  # swaps Linear -> INT8Linear
# -> ~15% GPU memory reduction on Evo2 7B, cosine sim ~0.998 vs BF16
```

### Type of changes

### CI Pipeline Configuration

Only the CPU-only L0 sanity test is intended for the default PR pipeline. The GPU benchmark and `tests/test_single_model.py` need NGC checkpoints and are meant to be run manually.

### Pre-submit Checklist

- [ ] I have tested these changes locally (INT8 math verified numerically; pytest pending a torch/CI run)
- [x] I have updated the documentation accordingly (recipe README)
- [x] I have added/updated tests as needed (CPU-only L0 sanity test)
- [ ] All existing tests pass successfully (pending CI `/ok to test`)

## Summary by CodeRabbit

* **New Features**
  * Added an INT8 weight-compression quantization recipe for Evo2, ESM-2, and Geneformer using ModelOpt methods.
  * Introduced a benchmarking workflow that measures latency, GPU memory savings, and logit quality, and exports results to CSV.
  * Added container startup and pretrained model download scripts.
  * Added quantization configuration definitions covering multiple precision families (FP8/INT8/INT4/NVFP4/microscaling/mixed).

* **Documentation**
  * Added recipe documentation with setup, supported checkpoints, and quick-start/benchmark guidance.

* **Tests**
  * Added CPU-only INT8 linear roundtrip tests and a single-model quantization validation script.

### DIFF
--- a/recipes/evo2_megatron_quant/.ruff.toml
+++ b/recipes/evo2_megatron_quant/.ruff.toml
@@ -1,0 +1,1 @@
+extend = "../.ruff.toml"
--- a/recipes/evo2_megatron_quant/README.md
+++ b/recipes/evo2_megatron_quant/README.md
@@ -41,7 +41,7 @@ pass does not touch, so the saving reflects compressing the Linear layers only.
 
 ## Layout
 
-```
+```text
 evo2_megatron_quant/
 ├── src/
 │   ├── adapters.py            # per-model load / tokenize / forward adapters
--- a/recipes/evo2_megatron_quant/README.md
+++ b/recipes/evo2_megatron_quant/README.md
@@ -50,10 +50,10 @@ evo2_megatron_quant/
 │   ├── compressed_linear.py   # INT8Linear (real int8 weight storage)
 │   └── compress_model.py      # walk model, swap Linear -> INT8Linear
 ├── tests/
-│   ├── test_compressed_linear_roundtrip.py   # L0 sanity: CPU-only, no checkpoint
-│   └── test_single_model.py                  # PTQ harness demo (GPU + NGC, manual)
+│   └── test_compressed_linear_roundtrip.py   # L0 sanity: CPU-only, no checkpoint
 ├── scripts/
 │   ├── download_models.sh
+│   ├── run_single_model.py                   # ModelOpt PTQ harness runner (GPU + NGC)
 │   └── benchmark_real_quant.py               # INT8 memory + quality benchmark (GPU)
 ├── configs/quant_methods.yaml
 ├── docker/start_container.sh
@@ -69,6 +69,9 @@ pytest tests/test_compressed_linear_roundtrip.py -v
 
 # Real INT8 weight compression benchmark (inside the BioNeMo container, GPU + NGC):
 python scripts/benchmark_real_quant.py --precisions int8
+
+# ModelOpt PTQ harness on one model (GPU + NGC):
+python scripts/run_single_model.py --model esm2 --quant INT8_DEFAULT_CFG
 ```
 
 Programmatic use:
@@ -91,8 +94,12 @@ stats = compress_model(evo2_model, precision="int8")   # swaps Linear -> INT8Lin
 
 ## Tests
 
-`tests/test_compressed_linear_roundtrip.py` is the CPU-only L0 gate: it
-round-trips INT8 on tiny random layers and asserts near-lossless reconstruction,
-no systematic weight bias, and a real reduction in the stored buffer size.
-`test_single_model.py` and the benchmark require a GPU and NGC checkpoints and
-are meant to be run manually, not in fast CI.
+`tests/` holds only the CPU-only L0 gate,
+`tests/test_compressed_linear_roundtrip.py`: it round-trips INT8 on tiny random
+layers and asserts near-lossless reconstruction, no systematic weight bias, and a
+real reduction in the stored buffer size. It needs neither a GPU nor a
+checkpoint, so it runs in normal CI.
+
+The GPU runners live under `scripts/` (`run_single_model.py`,
+`benchmark_real_quant.py`) — they require a CUDA device and NGC checkpoints and
+are meant to be run manually, so they are deliberately not pytest tests.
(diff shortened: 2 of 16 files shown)


huthvincent:
@coderabbitai review

huthvincent:
Thanks for the thorough review! Addressed the findings in `46eb3bde`:

**Fixed**
- `tests/test_single_model.py` — renamed `test_single_method` → `run_single_method` so pytest won't collect it as a test; `main()` now returns a non-zero exit code when any method FAILs/ERRORs.
- `src/compress_model.py` — switched to a package-relative import (`from .compressed_linear import ...`) with a top-level fallback for the `sys.path`-insert case.
- `src/quantize.py` — calibration forward-pass failures are now logged (with a warning if *all* passes fail) instead of being silently swallowed; dropped the always-zero `num_quantizers` field.
- `scripts/benchmark_real_quant.py` — exception-safe sequence-length sweep (vars initialized before `try`, cleaned up in `finally`), free the warmup output, `--max-seqlen` is now disableable (`BooleanOptionalAction` → `--no-max-seqlen`), stale result file removed before each worker run, `TimeoutExpired` handled, and stopped labeling the linear sweep a "binary search".
- `docker/start_container.sh` — default mount is now the recipe root (was an empty `./quantization` subdir); `--net=host` is opt-in via `USE_HOST_NET=1`.
- `scripts/download_models.sh` — unknown model names now fail fast with a non-zero exit.
- `src/__init__.py` — dropped AMPLIFY from the supported-models docstring (only ESM-2/Geneformer/Evo2 are registered).
- `README.md` — language-tagged the layout fence (MD040); `requirements.txt` — pinned `pyyaml`/`pytest` (torch and modelopt come from the BioNeMo base image, so they're intentionally not re-pinned).

**One I believe is a false positive**
- `tests/test_compressed_linear_roundtrip.py` (cosine test): `return_bias=False` does **not** drop the bias from the computation — `INT8Linear.forward` always applies it via `F.linear(x, w, self.bias)`; `return_bias` only controls whether `forward` returns `(out, None)` vs a bare tensor. So the test compares equivalent paths. Happy to add an explicit bias-enabled case as well if you'd prefer.

huthvincent:
@coderabbitai review

huthvincent:
Second pass addressed in `6f0f5e48`:

**Fixed as suggested**
- `scripts/benchmark_real_quant.py` — `json.load` is now guarded against a truncated worker result: a config whose worker was killed mid-write returns a `CRASH` row instead of raising `JSONDecodeError` out of `run_test()` and aborting the whole sweep before any CSV is written. Good catch.
- `src/quantize.py` — `--quant` tokens are stripped, and the resolved name is whitelisted against `ALL_QUANT_METHODS` before it reaches `getattr(mtq, ...)`.
- `src/quantize.py`, `src/adapters.py`, `scripts/benchmark_real_quant.py` — replaced the deprecated `torch.cuda.amp.autocast` with `torch.autocast(device_type="cuda", ...)`.

**Two I solved differently — reasoning below, happy to revisit**

1. **`quant_cfg` list layout (`quantize.py`).** I agree with the risk you identified — a silent no-op for `--all-mlp` is the real danger here. But I couldn't verify the "ModelOpt 0.45+ stores `quant_cfg` as an ordered list" claim from my environment, and writing speculative code against an unverified internal layout seemed worse than the bug itself. So instead of guessing at the list shape, I made the failure **loud**: if `quant_cfg` isn't a mapping containing a `default` entry, it now raises with the offending type rather than silently doing nothing. For a measurement harness, erroring out beats quietly reporting default-scope numbers as "all-MLP". If you can point me at the ModelOpt version and the concrete entry shape, I'll happily add explicit list handling.

2. **Focused tests for the exit-code contract (`test_single_model.py`).** The root problem was that the file was misnamed: it is a **manual GPU runner** (needs a CUDA device and an NGC checkpoint), not a pytest test — and its `tests/test_*.py` name is exactly what made pytest try to collect it in your first review. Rather than bolt mocks onto a GPU harness to make it look like a unit test, I moved it to **`scripts/run_single_model.py`**. `tests/` now contains only the genuine CPU-only L0 gate (`test_compressed_linear_roundtrip.py`), which needs no GPU or checkpoint and is what CI should actually run. That removes the collection hazard at the root rather than papering over it.

PR #1758, opened 2026-09-21
Prevent FASTA-to-JSONL conversion from erasing its source

VishnuR23:
### Description

Passing the same file as both input and output to `bionemo_fasta_to_jsonl` opens the output in write mode before reading the source, silently erasing the FASTA and reporting zero records. Different path strings can trigger the same loss through symlinks or hard links.

Check file identity before opening the output and raise a clear `ValueError` when input and output refer to the same file. The CLI reports the error and exits with status 1. Conversion to a distinct output retains the current overwrite and uppercase behavior. Regenerate the enforced Eden/phage copies.

### Type of changes

- [x] Bug fix

### Validation

- Four source-protection regressions fail on main: identical path, relative alias, symlink, and hard link.
- All seven CPU tests pass after the fix, including CLI behavior and normal multiline conversion with/without uppercasing.
- All applicable pre-commit hooks pass.
- GPU recipe suites were not run locally; this change operates on local text files.


recipes/evo2_megatron/src/bionemo/common/io/fasta_to_jsonl.py
```diff
@@ -97,7 +104,11 @@ def main() -> None:
     if not args.input.exists():
         print(f"Error: input file not found: {args.input}", file=sys.stderr)
         sys.exit(1)
-    count = fasta_to_jsonl(args.input, args.output, uppercase=args.upper)
+    try:
+        count = fasta_to_jsonl(args.input, args.output, uppercase=args.upper)
+    except ValueError as error:
+        print(f"Error: {error}", file=sys.stderr)
+        sys.exit(1)
```
jstjohn, line 111: I'm pretty sure this doesn't add anything, right? the stack trace will be printed either way and the program will exit on that raise. Otherwise looks good.

jstjohn (approved):
Approved with one cleanup request around exception handling (I think it would be better to not do it in this case), since as far as I can tell all you are doing is prepending the stack trace message with the word "Error: ".

